PR #5, opened 2015-02-24
Install script for Ubuntu

srijanshetty:
I've created an install script which uses fc-cache. Should work atleast on Ubuntu; haven't tested in other distros.


ryanoasis:
@srijanshetty Finally got around to briefly testing this and merging it in! :tada: 

Works well so far thanks!

PR #18, opened 2015-08-04
readme: mention python-fontforge minimum version

blueyed:
@ryanoasis 
Can you provide information about how to install it best?
The Ubuntu repos do not provide it (20120731.b-5+b3 being the latest in Debian/Ubuntu).

Is manually installing the way to go?
There is no package on PyPI, is there?

*Edit*: the Fontforge book has install instructions, and I've added a link to it.


ryanoasis:
@blueyed Thanks!

Issue #19, opened 2015-08-04
Missing file: glyph-source-fonts/Pomicons.otf


blueyed:
The Pomicons.otf file appears to be missing from the repo:

    % /usr/bin/python ./font-patcher ~/.fonts/Input/Input_Fonts/Input/Input-Regular_\(InputMono-Regular\).ttf
    The glyph named Tcommaaccent is mapped to U+021A.
    But its name indicates it should be mapped to U+0162.
    The glyph named Delta is mapped to U+0394.
    But its name indicates it should be mapped to U+2206.
    The glyph named Omega is mapped to U+03A9.
    But its name indicates it should be mapped to U+2126.
    The glyph named tcommaaccent is mapped to U+021B.
    But its name indicates it should be mapped to U+0163.
    The glyph named mu is mapped to U+03BC.
    But its name indicates it should be mapped to U+00B5.
    The requested file, Pomicons.otf, does not exist
    Traceback (most recent call last):
      File "./font-patcher", line 92, in <module>
        pomicons = fontforge.open("glyph-source-fonts/Pomicons.otf")
    EnvironmentError: Open failed

Is this an oversight or due to licensing?

ryanoasis:
@blueyed whoops completely an oversight thanks! I'll also get #20 in soon

PR #20, opened 2015-08-04
font-patcher: only load Pomicons.otf with --pomicons

blueyed:
Ref: https://github.com/ryanoasis/nerd-fonts/issues/19


ryanoasis:
Thanks, this is a good defensive coding update. I might have to resolve some conflicts when I eventually push my latest local changes but this is a good change either way :+1:

PR #21, opened 2015-08-04
readme: Fixes typo on some font names

leandrost:


ryanoasis:
Yes they are odd I did copy them initially from [powerline fonts](https://github.com/powerline/fonts) and I know this would be a good change however this looks like it was done to prevent possible license issues.

Maybe the text reference can be the original name but the actual font name itself might violate license unless changed.

For reference: https://github.com/powerline/fonts/pull/46 https://github.com/powerline/fonts/issues/34 https://github.com/powerline/fonts/issues/63

So, might need to look into this more and like I said maybe your change is fine and it would just be that changing the actual font file is the only issue

leandrost:
:+1:

ryanoasis:
@leandrost just FYI I think your PR still might be perfectly valid or at least partially. I just need to confirm the usage rules on the names :blush:

leandrost:
Yes, but maybe is better to keep consistet with powerline.

PR #27, opened 2015-09-02
Patcher Script clean up

cheba:
* It now runs on OS X
  * find incompatibility is fixed
  * FontForge from Homebrew can be found by Python
* It checks for FontForge availability and directs to installaction
  instructions if it's not found
* It is much shorter


ryanoasis:
Hey this is really awesome! Thanks a lot.

I had in my mind to do refactoring of the combinations but what you have looks very nice.

Thanks for the contributions!

I am on vacation now but I'll get this merged in asap :)

PR #28, opened 2015-09-02
Hack font

cheba:
https://sourcefoundry.org/hack/
https://github.com/chrissimpkins/Hack

A nice looking monospace font. It comes in 2 weights (normal and bold) and two variations each (regular and oblique). It also covers many natural languages.

I'm not sure about licensing, though. It seem to go with an open license but I don't know if that permits keeping this font in another repository.

---

:warning::warning::warning:

Requires #27. This branch must be rebased on master before merging this PR.

:warning::warning::warning:


cheba:
It looks like Hack font already has Powerline glyphs and they're properly sized. Patching them in produces far worse results both in regular and single-width modes. Probably Powerline patching should be made optional.

ryanoasis:
Great! Thanks. Funny I was just looking at this font a few days ago :laughing: 

That's a good idea to make the Powerline patching optional :+1: 

License looks okay but 'hack' looks to be reserved

cheba:
Updated PR with a fix for patcher that makes Powerline an option.

PR #98, opened 2016-08-14
Fix typo

controversial:
#### Description

_Fix a typo of the header in the Fira Code README._

A one-character patch.

### DIFF
--- a/patched-fonts/FiraCode/README.md
+++ b/patched-fonts/FiraCode/README.md
@@ -1,4 +1,4 @@
-Fura Code Powerline
+Fira Code Powerline
 ===================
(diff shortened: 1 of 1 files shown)


controversial:
Wait, maybe this is intentional, it's in a few places.

ryanoasis:
@controversial Hey yes it is due to [licensing (reserved font names)](https://github.com/ryanoasis/nerd-fonts/blob/master/font-patcher#L161-L162)

Perhaps the readme could be clearer about this.

controversial:
Ah, ok. Thanks for clearing that up

PR #101, opened 2016-09-02
Better installation script

zeorin:
It now has more options than copying all patched files or one font’s
patched font files.

In short, you can choose which patchset, which compatibility, which
method (copy or symlink), and which fonts you want to install, and sets some
sensible defaults for these options whilst not breaking past behaviour much.

Features that break with past behaviour:

* Installs fonts to a subdirectory of the user's fonts directory.
* Only installs the 'Complete' version of a font by default.
* Prefers OTF version if there are both OTF and TTF versions.

I did my best to follow the options convention of the font patcher script.

I wrote this on a Debian box, I haven't tested this on OS X.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,45 +1,278 @@
 #!/bin/bash
+# Install Nerd Fonts
+__ScriptVersion="0.1"
 
-# Set source and target directories
-powerline_fonts_dir="${PWD}/patched-fonts"
+# Default values for option variables:
+quiet=false
+mode="copy"
+clean=false
+mono=false
+windows=false
+extension="otf"
+patches=("Complete")
+compat=()
 
-# Accept a font / directory name, to avoid installing all fonts
-if [ ! -z "$1" ];
-then
-  # Ensure thar directory exists, and offer suggestions if not
+# Usage info
+usage() {
+  cat << EOF
+Usage: ./install.sh [-q -v -h] [[--copy | --link] --clean | --list]
+                    [--use-single-width-glyphs] [--windows] [--otf | --ttf]
+                    [--complete | --minimal | <patches>] [FONT...]
 
-  if [[ ! -d "$powerline_fonts_dir/$1" ]];
-  then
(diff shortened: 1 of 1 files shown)


ryanoasis:
I just quickly scanned through the changes and this seems really great. A lot of work you put into this! Some of the things are over my head but seems like great stuff.

I am currently away without laptop but when I am back I'll test it out and try to get this into the next release :smile: 

I aslo don't have OS X so if someone with that would be willing to do some sanity checks it would be greatly appreciated

jrolfs:
This is huge, thank you! Seems to work great on OS X, although `fc-cache` produces the following output:

```
/Users/jrolfs/Library/Fonts/NerdFonts: caching, new cache contents: 12 fonts, 0 dirs
/usr/local/var/cache/fontconfig: cleaning cache directory
/usr/local/var/cache/fontconfig: invalid cache file: b0a71e6bf6a8a1a908413a823d76e21f-le64.cache-7
/Users/jrolfs/.cache/fontconfig: not cleaning non-existent cache directory
/Users/jrolfs/.fontconfig: not cleaning non-existent cache directory
fc-cache: succeeded
```

I haven't really had time to debug but it doesn't seem to cause any problems outside of the cache file sticking around in `/usr/local/var/cache/fontconfig` and I tried installing 3 separate fonts in succession.

zeorin:
The invalid cache file warning isn't related to the font script per se, it can happen after installing any fonts with any method and then running `fc-cache -fv`.

Also, FYI you can install 3 fonts at once if you like, e.g.: `./install.sh AnonymousPro Inconsolata Monoid`

jrolfs:
@zeorin yeah I ran `fc-cache` on it's own and the problem is not at all specific to the script. @ryanoasis shall we merge?

ryanoasis:
Thanks for this and the review, really appreciate it. I'd like some time to review it. Maybe this weekend if I have time. Sorry for the delayed follow-up.

Issue #102, opened 2016-09-02
Suggest shallow cloning of repo


zeorin:
Suggest in the docs that users clone the repo with `git clone --depth 1` if they're not looking to contribute and just want to get the fonts easily. Together with the improved install script from #101 this makes it real easy to get and stay up to date without wasting bandwidth or disk space.

ryanoasis:
Good idea. Yes I looked again and it is still not clear that one does not need to clone the repo and if they contribute shallow should be sufficient. Thanks

nowox:
+1

Issue #58, opened 2016-02-23
Reduce complexity in output?


eugenesvk:
Partially following up on https://github.com/ryanoasis/nerd-fonts/issues/42 and also continuing my thoughts after dealing with https://github.com/ryanoasis/nerd-fonts/issues/56 may I offer a few thoughts on how to reduce the staggering complexity in the number of fonts produced by the patcher?

With short names (e.g. "NC" instead of "Nerd Font Complete") Windows version is never needed.
Also, we could also get rid of most of thee complexity with variations.

Currently, you have **two conflicts** among glyph sets:
1. 'Custom SetiUI' (52 glyphs) and 'Devicons' (198)
2. 'FontAwesome' (609) and 'Octicons' (164)
For the **1st** there is no variation, 'SetUI' takes precedence. For the **2nd** you either shift 'Octicons' if 'FA' is present or just have 'Octicons' in default positions without 'FA'.

So, in your scheme, **only two fonts are needed**:
**2a)** complete with 'Octicons' shifted and 'FA' in default position and
**2b)** complete with 'Octicons' in default position and 'FA' shifted
- either completely or
- have 'FA' glyphs following 'Octicons' range remain in their spot (~400 glyphs) and only shifting first ~200 (bigger than 164 above since 'Octicons' has a lot of holes in its range)

That's it in the current implementation!

Personally, I'd prefer in the **1st** conflict to prioritize 'Devicons' to make it match the intended codepoints and allow using well-established reference tables (https://vorillaz.github.io/devicons/#/dafont) and then just shift 'SetiUI' to right after 'Devicons'

eugenesvk:
I've customized your script for my own purposes to follow this logic (prioritizing 'Devicons' over 'SetiUI', moving 'SetiUI' right after 'Devicons') and only retaining 'FA' glyphs that don't conflict with 'Octicons' (haven't tested yet moving the displaced glyphs elsewhere).

Now that I've thought of additional simplification (with only two fonts required) the patcher could have those two additional aggregate options for producing *SourceFontNFA* prioritizing 'FA' and *SourceFontNO* prioritizing 'Octicons'.

If you're interested in such a simplification, I'd be happy to do a PR with the script and help with fixing the names for couple of fonts that I use myself.

ryanoasis:
I really appreciate your ideas here and they make a lot of sense. :bow: 

[...]

PR #103, opened 2016-09-19
Add Hasklig

jrolfs:
Perhaps there are more variants than we care about?

Chug-a-lug: 
patching


ryanoasis:
Good question, it's both a positive of this whole project but also can be a negative :stuck_out_tongue: 

Sorry I haven't followed up on this yet until now.

jrolfs:
No worries, yeah it's definitely both a plus and minus. Git has gotten better with binary files but dumping a bunch more in can definitely bloat the repo...

The problem here is that the source font provides so many variants which could be useful, however we could limit the variants included to something similar to another font that provides variants (like **Source Code Pro** from which this ligature font is derived).

ryanoasis:
Yeah...

> Over 5,700 unique combinations/variations of patched fonts

Is more of a bragging point than something actually feasible/useful in real life :wink: 

--

My thinking on the whole massive amounts of variations issue is that should be addressed but maybe in a few versions (probably 1.0.0) and for now I'd like to move forward with enhancements and easy bugs (to avoid getting stuck on the hard ones). Anyway I am going to put more time into this project again and start preparing for v0.9.0. I'll probably create the branch and merge this in tomorrow :smile: 

Related:
* #58 , in particular removing the "windows" versions is something I want to do and should be feasible (just name length limitations)
* #102, shallow cloning

--

ryanoasis:
whoops I did something wrong in the feature branch flow. the changes for this and commits are there but it's just not marked as 'merged'

PR #133, opened 2017-02-19
update readme.md

teddy-error:
#### Description
Corrected installation via Homebrew instructions. (In the current instructions, the command given searches for an unrelated cask called Nerdtool)

Corrects the command in the readme instruction

N/A

The command doesn't seem like it would have ever done anything useful, TBH. I'd guess this was mistakenly written at some point. It is possible that this was mistakenly omitted as the first step, in which case it might make sense to add it previous to the search rather than replace it.

N/A

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -238,7 +238,7 @@ Fonts are available for download as packages in the [latest release](https://git
 All fonts are available via [Homebrew Fonts](https://github.com/caskroom/homebrew-fonts) on macOS (OS X)
 
 ```sh
-brew cask search /nerd/
+brew tap caskroom/fonts   
 brew cask install font-hack-nerd-font
 ```
 ### `Option 5: Unofficial Arch User Repository (AUR)`


ryanoasis:
Hey thanks for the fix.

> mistakenly written at some point

Yeah that was me :relaxed: I don't actually use Mac OS so I do a lot of the doc and contribution for it blindly

ryanoasis:
Ah I was trying to make mention of the search to find the correct Nerd Font:
https://github.com/caskroom/homebrew-fonts#search-for-a-font

But that can be mentioned better or elsewhere

teddy-error:
You bet, Ryan. My pleasure!

On February 19, 2017 at 16:14:11, Ryan L McIntyre (notifications@github.com)
wrote:

Hey thanks for the fix.

mistakenly written at some point

Yeah that was me ☺️ I don't actually use Mac OS so I do a lot of the doc
and contribution for it blindly

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/ryanoasis/nerd-fonts/pull/133#issuecomment-280953829>,
or mute the thread

PR #151, opened 2017-04-08
Kill DejaVu/Powerline in favor of original DejaVu fonts

agriffis:
Nerd Fonts patches Powerline symbols into the original DejaVu fonts which are
a more recent release than the "X for Powerline" editions. This PR removes the
confusing duplication in favor of the more recent fonts.


ryanoasis:
@agriffis This makes sense to me. :+1: 

It has the powerline symbols in this `2.37` version vs the `2.33` version (powerline patched).

One thing I noticed is that the readme.rst seems to be from the Powerline version, we probably want to replace it with the readme.md from: https://dejavu-fonts.github.io/

agriffis:
Do you mean the latest README.md from https://sourceforge.net/projects/dejavu/files/dejavu/2.37/ ?

I haven't quite figured out how you're managing README's in the repo. Where would this go?

ryanoasis:
Yeah that is the readme I was referring to.

The readme just has to go in the `src/unpatched-fonts/DejaVuSansMono` directory.

The [standardize-and-complete-readmes.sh](https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/standardize-and-complete-readmes.sh) will take search and build readmes for the patched versions :smile:

agriffis:
@ryanoasis I just added a commit with that change. Feel free to squash if/when you merge.

ryanoasis:
Sure thanks

Issue #49, opened 2015-11-26
Include a python3 version of the font-patcher script


SageEx:
Hi,
First of all thanks a lot for this project.
Recently, I tried to patch the font Fantasque Sans Mono.
The problem is that on ArchLinux the fontforge library has bindings for python3 by default( python3 is the default python). So, running with ` ./font-patcher ... ` didn't work. 
I had to use the fontforge program directly - ` fontforge -script font-patcher --fontawesome --careful FantasqueSansMonoRegular.ttf `. 
But the script is in python2. Using 2to3 on the file gave an attribute error. I found the problem and everything worked.

I suggest including a python3 version of the script in the repo. I am attaching it - it will save a lot of hassle for many people.

Besides, how about including Fantasque Sans Mono in the default patched fonts set ?

[font-patcher.txt](https://github.com/ryanoasis/nerd-fonts/files/44653/font-patcher.txt)

ryanoasis:
@SageEx Thanks for the info and I am glad this project is useful to you :smile: 

> python 3

I will have to look into that, I see the changes you had to make. There has to be a way to make it compatible for py3 and py2 without having 2 separate scripts to maintain?

> how about including Fantasque Sans Mono in the default patched fonts set ?

Looks like the license will allow this (OFL (SIL Open Font License)), the only I was checking for is if there is a Reserved Font Name specified in the license but I don't see one. If the name is reserved we would just have to rename the patched version (e.g. Hack becomes Knack).

SageEx:
Most of the changes are due to the convention changes from python2 to python3 - like 

* brackets in print statements
* dic.iteritems() to dic.items
etc.

I doubt that a single python script would suffice.

Rather, as a solution, you can maintain the python2 script only and use 2to3 to automatically generate the python3 script. The only manual change you will have to make in the generated script is to revert the change at line 306. By default 2to3 will change from Unicode to str - but str is not an attribute - hence the error. Any way to use 2to3 which does not change this will be the answer to this problem.

I really hope to see the Fantasque Sans Mono font in the repo. 

Thanks again for this repo - along with vim-webdevicons.

ryanoasis:
Thanks btw. I will probably get to this but not anytime soon. 

    Fantasque Sans Mono

Probably not too difficult to add just a handful of bugs that I've been focusing on.

Cheers!

ryanoasis:
[...]

PR #152, opened 2017-04-08
Make font-patcher Python 3 compatible.

agriffis:
Fedora's fontforge modules only work on Python 3. The small changes in
this patch make font-patcher work on Python 3 so I can run the script on
Fedora (26 at least, I don't know about older versions)

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -1,6 +1,8 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
 # coding=utf8
 # version: 1.0.0
+
+from __future__ import absolute_import, print_function, unicode_literals
 
 version = "1.0.0"
 projectName = "Nerd Fonts"
@@ -70,8 +72,8 @@ actualVersion = int(fontforge.version())
 # versions tested: 20150612, 20150824
 
 if actualVersion < minimumVersion:
-    print projectName + ": You seem to be using an unsupported (old) version of fontforge: " + str(actualVersion)
-    print projectName + ": Please use at least version: " + str(minimumVersion)
+    print("{}: You seem to be using an unsupported (old) version of fontforge: {}".format(projectName, actualVersion))
+    print("{}: Please use at least version: {}".format(projectName, minimumVersion))
     sys.exit(1)
 
 verboseAdditionalFontNameSuffix = " " + projectNameSingular
@@ -150,7 +152,7 @@ try:
     # now we have the correct item:
     subFamily = sourceFont.sfnt_names[sfntNamesStringIDIndex][subFamilyTupleIndex]
 except IndexError:
-    print projectName + ": Could not find 'SubFamily' for given font, falling back to parsed fontname"
+    print("{}: Could not find 'SubFamily' for given font, falling back to parsed fontname".format(projectName))
     subFamily = fallbackStyle
 
 # some fonts have inaccurate 'SubFamily', if it is Regular let us trust the filename more:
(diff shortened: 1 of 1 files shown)


ryanoasis:
@agriffis Hmmmm I understand, however doesn't the [py3 version](https://github.com/ryanoasis/nerd-fonts/blob/master/font-patcher-py3) not work for you? :thinking:

agriffis:
@ryanoasis hilarious, I hadn't even noticed `font-patcher-py3`

Why have two versions though? It's trivial to make one script compatible with both—that's what this PR does. Can we just kill `font-patcher-py3`?

ryanoasis:
@agriffis Sure no problem

> Why have two versions though?

The reason there are 2 versions is because the script was initially written for `python 2.x` but there was a request to have a script that can run with `python 3.x` (#49)

> It's trivial to make one script compatible with both—that's what this PR does.

Yeah it seems like it is trivial as long as the users have at least `2.7.x` as far as I understand it (then we can use the future pkg).

> Can we just kill font-patcher-py3?

If this script will work in `2.7+` (which seems like it would work in 2.6+) then yes we **_can_**. I still think a better question is: **_should we_?**

1. Do you have any particular experiences on future vs maintaining both (using `2to3`) ?
2. Unmodified code (py2 and py3) using `__future__` (single script) vs maintaining py 2 codebase and running`2to3` pros and cons?

**maintaining both:** cleanly/clearly separates the codebases, py3 doc discourage writing for both versions
**unmodified code:** don't have to worry about running `2to3` each time, `2to3` seems like it is not perfect

I was doing a small amount of googling though as to what the best way to maintain for both versions, just some interesting links:
* https://wiki.python.org/moin/PortingToPy3k/BilingualQuickRef
* https://docs.python.org/3/howto/pyporting.html
* [Supporting Python 2 and 3 without 2to3 conversion](http://python3porting.com/noconv.html)
* https://wiki.python.org/moin/Python2orPython3
* http://python3porting.com/strategies.html

I might be overthinking this :blush:

agriffis:
IMHO two scripts is just unnecessary complexity. Python 2.7 has been around since 2010 and there aren't any developer-facing Linux distributions that are still sporting 2.6.

I maintain a few Python libs that are compatible with both Python 2 and 3. I wouldn't dream of maintaining separate versions—that's just a lot of opportunity for unexpected divergence. I think a lot of the concern about forward/backward compat was resolved with Python 2.7 and 3.3 (when they brought back the `u` syntax for unicode to smooth backward compatibility with 2.x)

But those are just my opinions, and remember I created this PR because I didn't see the py3 version. I still hold these opinions, but don't need to force them on you. :-)

agriffis:
and I had to laugh at "I might be overthinking this" 😆

agriffis:
By the way, the Python 2 version of `font-patcher` depends on `argparse` which was first introduced in 2.7 😉

ryanoasis:
@agriffis You definitely have more knowledge and experience with this. I think I will need to take some more time to digest the options but starting to lean toward your opinion. Thanks again.

Didn't know/realize that about `argparse`, I think targeting >= 2.7 is more than sufficient

ryanoasis:
Maybe tomorrow :sleeping: :zzz: 

Really this is much better though when you made me re-think my workflow. So thanks for that!

ryanoasis:
@agriffis Sorry for the delay on this. I was merging this in today but realized I could no longer test it. I am going to test it in a virtual machine

ryanoasis:
Not sure what I messed up in merging this but got officially in via GitHub button merge :roll_eyes:

PR #178, opened 2017-07-11
Change the font family name for mono fonts

### LINKED ISSUE #176: Mono fonts have same family name as double-width fonts
#### Subject of the issue

When generating single-width fonts, the PostScript Name and Full name of the font both identify it as Mono, but the Family doesn't. This means if I install both mono and double-wide versions of the same font, I end up with one font family that contains two "Regular" fonts. This makes it rather difficult to actually use properly.

Instead I think the patcher should add "Mono" to the family of the font.

#### Your Setup

- macOS 10.12.5

kballard:
#### Description

Add " Mono" to the font family name when generating single-wide fonts.

#### Requirements / Checklist

- [X] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [X] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [X] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [X] Scripts execute without error (if necessary):
- [X] Extended the README and documentation if necessary, e.g. You added a new font please update the table

Run `./font-patcher` with the `-s` flag and check the generated fonts.

Fixes #176.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -172,6 +172,8 @@ if args.windows:
         familyname = familyname[:maxFamilyLength]
 else:
     familyname += " " + projectNameSingular
+    if args.single:
+        familyname += " Mono"
 
 # Don't truncate the subfamily to keep fontname unique.  MacOS treats fonts with
 # the same name as the same font, even if subFamily is different.


kballard:
Assuming there are pre-patched mono fonts, they'll need to be regenerated.

ryanoasis:
@kballard Thanks I'll take a look at and test it this weekend.

> Assuming there are pre-patched mono fonts, they'll need to be regenerated.

Yep, so I will probably tie this to release 1.1 (because rebuilding all the fonts takes a good chunk of time), but at the very least perhaps we can get the logic fix merged into the code

PR #187, opened 2017-07-30
Added support to install fonts to global system path.

brennanfee:
#### Description

Allows execution 'sudo ./install.sh -S <fonts>' which will use the
global system install path rather than the users local path.  May
also use --install-to-system-path as a long version.  Also allows 
-U (or --install-to-user-path) to explicitly set to install to users path, 
but that is the default so just there for completeness.

I made sure to support and test both MacOS and Linux.

#### Requirements / Checklist

Enhances the install.sh script to support installing to the global system font path.

Run the ./install.sh command like so:  sudo ./install.sh -S Hack then check the global system font
path for a sub-folder NerdFonts and the Knack fonts inside.

N/A

N/A

N/A

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # Install Nerd Fonts
-__ScriptVersion="0.1"
+__ScriptVersion="0.2"
 
 # Default values for option variables:
 quiet=false
--- a/install.sh
+++ b/install.sh
@@ -11,12 +11,14 @@ windows=false
 extension="otf"
 patches=("Complete")
 compat=()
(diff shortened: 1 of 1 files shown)


ryanoasis:
Thanks for this! :smile: I'll merge it in for [v1.1.0 Release](https://github.com/ryanoasis/nerd-fonts/milestone/16)

brennanfee:
My pleasure.  I really like what you've done and this was a (small) way of saying thanks.

ryanoasis:
@brennanfee Thanks! Appreciate that :smile:

PR #197, opened 2017-09-19
Fix left arrow powerline glyph for InconsolataGo font

iloginow:
#### Description

I use 'InconsolataGo Nerd Font' as a default font for my terminal. I noticed an ugly 1px gap between powerline left arrow glyph and the rest of powerline. And that's what I fixed.

### Requirements / Checklist

Fixes an ugly 1px gap between powerline left arrow glyph and the rest of powerline for InconsolataGo Nerd Font
Before:
gscreenshot_2017-09-19-145237

After:
gscreenshot_2017-09-19-150151

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -136,7 +136,7 @@ done
 shift $(($OPTIND-1))
 
 # Set source and target directories, default: all fonts
-nerdfonts_root_dir="${PWD}/patched-fonts"
+nerdfonts_root_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/patched-fonts"
 nerdfonts_dirs=("$nerdfonts_root_dir")
 
 # Accept font / directory names, to avoid installing all fonts


ryanoasis:
@iloginow

Thanks for the contribution. Yeah I can see with your changes it looks better :smile: 

**However** the change was applied ad-hoc to only the patched font?

There will be a problem when we re-patch the fonts the change/tweak will get blown away :cry: 

Is it only the `left divider arrow` powerline glyph with the gap not the `right divider arrow` one?

I am _actually okay_ with merging this for now **but** I see it as more of a temporary fix and I don't think this is the root fix. I think we still probably need to do some tweaking of the positioning in the patcher script to solve the minor gap issues that seem to be lingering.

Thoughts?

iloginow:
@ryanoasis 
I am sorry. I just didn't think that these fonts are supposed to be re-patched. It was just a quick fix. And I wanted to share it just in case someone else is struggling to get rid of this little gap. You can just skip it or think of it as an issue, not a pull request. 
The problem appears to be the `left divider arrow` only. And that's why I thought the patcher script is probably doing all right. I'm not sure... but maybe we just have to fix the source glyph itself in src/glyphs/PowerlineSymbols.otf... What do you think?

https://vi.stackexchange.com/questions/4415/spacing-appears-in-vim-airline-on-right-side
https://github.com/powerline/powerline/issues/359

ryanoasis:
@iloginow Hey sorry for the late reply.

> maybe we just have to fix the source glyph itself in src/glyphs/PowerlineSymbols.otf... What do you think?

Yes I am good with that, how much manual tweaking of the powerline glyph did you have to do in the InconsolataGo fonts? If it was very minor could you make the same in the PowerlineSymbols.otf? That way whenever we repatch your tweaks are not lost :smile: After that I can review, test a few other fonts, and if everything seems good I will merge your change in.

----

I see you also added InconsolataLGC. Can you do that in a separate PR?

Is it from: https://github.com/MihailJP/Inconsolata-LGC ?

We typically add the unpatched version the font as well in https://github.com/ryanoasis/nerd-fonts/tree/master/src/unpatched-fonts and update the readme counts. So missing a few steps from the Wiki: https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md#steps-for-adding-a-new-font-or-removing-an-existing-font

viettuan1807:
Same issue with **Sauce Code Powerline**
Please fix.

PR #291, opened 2018-08-30
Move font-patcher log for Blex inside <details>

kevinSuttle:
Just a little UX touch


ryanoasis:
This is really a good idea. I'll merge soon, however we can apply this to all fonts and these readmes get partially generated through the script: https://github.com/ryanoasis/nerd-fonts/blob/d1d9a7eb23fc9f5efc00c89cdd58694547a0df39/bin/scripts/gotta-patch-em-all-font-patcher!.sh#L219-L224

PR #304, opened 2018-11-07
Fix type in readme: hermUt -> hermIt

emil14:
Just a small fix in readme.md


ryanoasis:
So this is not exactly what it should be changed to (because of RFNs) but you are actually correct that it is wrong. It should be `hurmit`

See:
https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/Hermit
https://github.com/ryanoasis/nerd-fonts/blob/master/font-patcher#L242-L243
https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting#why-do-some-of-the-fonts-names-appear-incorrect-or-appear-to-have-typos

PR #310, opened 2018-12-12
Prepend a full-stop to dir. name

Trabing:
This is a minuscule oversight, but .config is a hidden directory

#### Description

#### Requirements / Checklist

- [ ] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)


ryanoasis:
So this looked good to me too on initial scan but I think there is some confusion. That path seems to be referring to the folder on the GitHub repo (https://github.com/source-foundry/Hack/tree/master/config/fontconfig) and the second one is the one referring to the user folder .config.

So I believe it is correct as is (even if it is a bit confusing on initial look).

Trabing:
Okay, I'll take another look at it.

PR #322, opened 2019-02-09
Remove text in english

leoz2007:
#### Description

Remove "There is also a [font patcher](#font-patcher) available if your desired font isn't already patched."
The french traduction is already on the file "Un [généreur de polices](#font-patcher) est aussi disponible si vous souhaitez ajouter une police non présente."


ryanoasis:
Thanks @leoz2007

ryanoasis:
forgot to tag this in the release and subsequently forgot to add you on the changelog (will correct shortly). working on all-contributors update.

PR #328, opened 2019-02-19
Update Iosevka to 2.1.0

MindTooth:
#### Description

Update Iosevka to 2.1.0.  Please look over if I've done it right.  Were a bit scared as the documentation don't specify a approach for just updating a font.

I've copied over the files from the new release, however I'm missing some new styles.  But were a bit unsure wether to include them.  Please hack at my branch if need be.  I've at least done the task of copying each original font into the src folders.

OT: Noticed that the `gotta-patch-em-all` did not complain about missing the `configparser` for Python.  And I did not notice at firs that the fonts did not get created.  Maybe add a check there as well?

#### Requirements / Checklist


ryanoasis:
Thanks for the contribution @MindTooth :smile: . I will look it over.

> Were a bit scared as the documentation don't specify a approach for just updating a font.

Would you mind elaborating? I want to make continuously make the docs better/easier.

If you executed `./gotta-patch-em-all-font-patcher\!.sh XYZ` where XYZ is the name of the font it would have re-patched just that particular font set.

It seems like you followed https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md#steps-for-adding-a-new-font-or-removing-an-existing-font ? If so that should be good enough in most cases

MindTooth:
Thanks. 😄 

Well, when it read only add/remove I felt unsure about right the steps required.  An own guide on how to update would help not just for me, but perhaps others as well.

Is a script suitable to do the downloading and patching?  Like with this font; it's just a version bump.

ryanoasis:
I agree with you on a separate guide for UPDATING as well as script to execute the steps. Thanks

PR #354, opened 2019-06-07
Update readme.md

themediapot:
#### Description

Updated instructions to reflect how to enable fonts in macos's various terminal apps.

#### Requirements / Checklist

Just updated README.MD

Test the instructions if necessary.

I was looking for instructions for enabling font after installing colorls, but could not find any that specifically says that font has to be enabled for terminal apps.
 
Nil.

N/A

### DIFF
--- a/patched-fonts/Hack/readme.md
+++ b/patched-fonts/Hack/readme.md
@@ -73,9 +73,38 @@ Some Linux users may find that font rendering is improved on their distro with [
 
 1. Download the [latest version of Hack][ttf_latest].
 2. Extract the files from the archive (`.zip`) and click to open them.
-3. Follow the instructions from your operating system.
+3. Follow the instructions from your operating system and Terminal application.
 4. Enjoy!
 
+#### Terminal.app
+
+1. Open Terminal preferences
+2. Select Profile Tab
+3. Click Change under section for Text
+4. Choose nerd-font
+
+#### iTerm2
+
+1. iTerm2 -> Preferences
+2. Select Profiles
+3.Select Text
+4. Choose Font
+5. Change Font
+6. Select nerd font
+
+#### Hyper
+
+1. Go to Hyper > Preferences
+2. Locate line similar to below:
+```
+fontFamily: 'Menlo, "DejaVu Sans Mono", Consolas, "Lucida Console", monospace',
+```
+3. Add "Hack Nerd Font" to the list of font family available to Hype, for example:
+```  
+fontFamily: 'Menlo, "DejaVu Sans Mono", Consolas, "Lucida Console", monospace, "Hack Nerd Font"',
+```
+4. Save file.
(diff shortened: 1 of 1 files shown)


ryanoasis:
I like this change but there is a problem. So the readmes we output in the patched dir are actually generated by a script for all fonts, with a Nerd Font specific addendum which beings at the "Which font?" portion. I think the change would need to be [here](https://github.com/ryanoasis/nerd-fonts/blob/master/src/readme-per-directory-addendum.md).

ryanoasis:
Now that I look at this again I think this would work better as a PR on the [official Hack repo](https://github.com/source-foundry/Hack) instead. Just with the Nerd Font bits removed.

I didn't see a great way as structured to add it to the addendum that we tack on to all the patched fonts readmes.

Do you have any opinions @chrissimpkins ?

patched-fonts/Hack/readme.md
```diff
@@ -73,9 +73,38 @@ Some Linux users may find that font rendering is improved on their distro with [
...
+6. Select nerd font
+
+#### Hyper
+
+1. Go to Hyper > Preferences
+2. Locate line similar to below:
+```
+fontFamily: 'Menlo, "DejaVu Sans Mono", Consolas, "Lucida Console", monospace',
+```
+3. Add "Hack Nerd Font" to the list of font family available to Hype, for example:
```
chrissimpkins: I reference the family name in this line in my comments in this thread.

chrissimpkins:
The important bit of information that I see in the changes here for NF's installs (and other project-specific installs) is the family name that they must use in the font stack definitions.  Full font family names aren't always clear to end users when they download fonts.  IMO, users should consult the current application documentation for font installation instructions in platform/application specific circumstances. This isn't a criticism of the work here. It is helpful for some users.  It is just difficult to draw the line on what platform x application should be supported in application-specific install docs and maintain the information in a current state.

ryanoasis:
Thanks @chrissimpkins

ryanoasis:
> This isn't a criticism of the work here. It is helpful for some users. It is just difficult to draw the line on what platform x application should be supported in application-specific install docs and maintain the information in a current state.

I agree, this is good work and helpful and yet there is a balance of helpful and information overload

@themediapot I think this should be re-worked just a bit to be more general and not specific to Hack. I think the instructions could be changed just a little in this regard and then could be added to the [readme addendum](https://github.com/ryanoasis/nerd-fonts/blob/master/src/readme-per-directory-addendum.md) which will get put into all fonts readmes once the readme update script runs

ryanoasis:
@themediapot I am actually going to close this for now, mostly for house keeping purposes.

You are entirely free to chime in again and we can revisit this PR if you would like to make some additional changes as per my comment https://github.com/ryanoasis/nerd-fonts/pull/354#issuecomment-515652049

Thanks!

Issue #352, opened 2019-05-23
Fira Code


ShalokShalom:
Hi there :D

The developer of the awesome font 'Fira Code' allows us the usage of the correct name:

https://github.com/tonsky/FiraCode/issues/767#issuecomment-495163622

ryanoasis:
Great. Either I didn't realize the license had no RFN specified or it was updated some time back.. I think I just assumed it was reserved and error-ed on the safe side

ShalokShalom:
Sure sure. So I can close this issue? Is there already a PR?

ShalokShalom:
I did it for the Readme, already done on the homepage. Closing :hugs:

PR #356, opened 2019-06-22
Change Fura to Fira

ShalokShalom:
See the [according](https://github.com/ryanoasis/nerd-fonts/issues/352) issue, developer allows us to use the name. :hugs:


ryanoasis:
Thanks @ShalokShalom . There is actually more that needs to be updated (for the patched fonts) but this is a good start.

Wello6143:
Haha, I have to rename the fonts because of this typo.

ryanoasis:
@Wello6143 Yes it _was_ named Fura to comply with the license, we can now use the actual canonical name Fira :smile:

PR #249, opened 2018-03-25
Remake the original-source font file

danbee:
#### Description

I sourced a bunch of replacement SVG's for this font because many of the originals were a little rough looking. There were also many glyphs that were made up of multiple overlapping/intersecting shapes so I spent some time reducing these down to a single path each. I also made a few changes to some of the icons to improve clarity.

I used IcoMoon to put together the font, and then converted it to OTF using FontForge. I used Glyphs to clean up the paths as FontForge seemed to be having trouble doing that.

I have been unable to rebuild the fonts on my Mac to check. I managed to run the script (`./gotta-patch-em-all-font-patcher\!.sh`) but I don't get any font files in the `patched-fonts` directory. The script did not output any errors 😕.

#### Any background context you can provide?

The last three icons in the font didn't seem to be rendering. The icons are blank on the web interface when you search for `custom`. My original intention was to add the Elixir icon, but I found it was already there, just not working correctly. I believe it was broken due to the problems with the glyphs.

#### Screenshots (if appropriate or helpful)

screen shot 2018-03-25 at 18 40 58

PR #358, opened 2019-06-28
Repatch all the fonts to incorporate the new icons

danbee:
#### Description

This re patches all the fonts with the improved icons I added in #249.

In #249 I replaced many of the icons in the font with better SVG's and cleaned up a number of icons. This patches all the fonts with these new icons.


ryanoasis:
I believe those were already repatched since then.. but I could merge this in anyway to get all the latest changes

danbee:
@ryanoasis The website is definitely still showing the old Elixir icon and I haven't seen any font updates come through Homebrew so I assumed this hadn't been done. I guess those are separate steps though!

PR #361, opened 2019-07-25
Fix the negative error number

evaryont:
#### Description

This fixes the shellcheck error. Running on my laptop, `exit -1` results in an actual return code of 255, so make that explicit.

#### Requirements / Checklist

This should help with the Travis CI failures.


ryanoasis:
Thanks good catch. Interesting, so it looks like it is actually 255 because it is 'out of range' of the error codes: http://tldp.org/LDP/abs/html/exitcodes.html

Perhaps we should just make it `1` ?

I can merge this tomorrow

ryanoasis:
marking on 2.1.0 release, need to update changelog and all-contributors

Issue #254, opened 2018-04-01
fl/fi Ligatures always appearing in patched Monaco font


timoxley:
I've even tried editing the patch script to remove *all* lookup tables, yet the damn ligatures continue to persist. 

i.e.

```
  for table in sourceFont.gsub_lookups:
      symfont.removeLookup(table)
```

Any ideas on what's going on? I'm guessing there's some default ligature setting being inherited? Is there a way to turn that off in the generated font?

Turning off `set macligatures` in vim is my temporary workaround.

ryanoasis:
@timoxley Hey sorry not that I know of.

The ligatures are present in the source font pre-patching right?

What is the name of the tables that get removed in the iteration?

By the way the font-patcher is setup with the ability to remove subtables (but not tables) via `--removeligatures` in conjunction with `--configfile [json like config file]`

config example: https://github.com/ryanoasis/nerd-fonts/blob/master/src/unpatched-fonts/Meslo/config.json

ryanoasis:
@timoxley Thanks for the details :+1: , that is strange. I would like to take a look but I cannot tell you when that'll be :sweat:

timoxley:
@ryanoasis 

No action required on your part, was hoping it would be something you knew off the top of your head.

Oh, also note that when I first patched the font using --complete, the `fi` and `fl` ligatures were remapped – one became an icon for a telephone. This was an entertaining surprise when I discovered it in my editor.

## Source Font Monaco

![image](https://user-images.githubusercontent.com/43438/38241973-2aafe9f0-3766-11e8-8bea-9b3d35132487.png)

## After `font-patcher --mono --powerline --fontawesome --powerlineextra`

![image](https://user-images.githubusercontent.com/43438/38241990-388ec438-3766-11e8-9d02-7dae4ecbd091.png)

## After `font-patcher --mono --complete`

![image](https://user-images.githubusercontent.com/43438/38242003-3f733fc2-3766-11e8-8a1b-e5d2e90fe95e.png)

🥇

sirplatypos:
I have been having the problem with the phone symbol and such as well using the AUR nerd-fonts-complete package.  Is there a way to get around this problem without patching all of the fonts by hand?

timoxley:
@sirplatypos I just live with the phones

sirplatypos:
[...]

Issue #57, opened 2016-02-19
Ligatures


ryanoasis:
Any interest in looking into and attempting to add ligatures to fonts without?

rohithmeethal:
Please add ligature, I cant even look into my IDE with fonts without ligature, once I started using fonts with ligature.

kballard:
I would also be interested in having an option that adds programming ligatures, like those talked about in https://www.hanselman.com/blog/MonospacedProgrammingFontsWithLigatures.aspx.

michaeltlombardi:
Might be able to leverage [ligaturizer](https://github.com/rojiani/Ligaturizer/) for this.

ryanoasis:
@michaeltlombardi Thanks. That looks cool! Hmm poked around not sure about the future of project though..

michaeltlombardi:
I was thinking of using it to have a personal hacked version of the knack font for my powerline setup. :)

ryanoasis:
@michaeltlombardi Let us know if you do and how it turns out. I would be willing to use something like this in Nerd Fonts

michaeltlombardi:
I'm actually currently using furacode, which already included the ligatures. Will set aside time to run it against knack and report back. :)

foisonocean:
+1

RavenHursT:
Yes please :-)

seefood:
I think the Ligaturizer tool covers it, all nerd-fonts needs to do is not break it (see #248) or maybe run it after the patching to fix ligatures on the new hybrid font?

Xinkle:
+1

ryanoasis:
This _might_ have to a be `3.0` release. Some of the glyphs are clobbering the alphabetic presentation forms (ligatures).

https://jrgraphix.net/r/Unicode/FB00-FB4F

Umbrien:
How about adding <3 ligature? There should be more love in our code! :heart:

Jab2870:
If you decide to go down the ligature route, would it be possible to add bigger icons, perhaps by prepending them with a special character.

As an example, I use the patched Iosevka font, which is a thin font, and this results in a lot of the icons being very small, as shown in this screenshot. The icons are (obviously) constrained by the width of the font.

![image](https://user-images.githubusercontent.com/1909202/78544934-d0608980-77f2-11ea-938a-d41088a41389.png)

If we could have the icons span 2 characters, by using a ligature, this would in my case, make them a lot easier to distinguish.

[...]

PR #364, opened 2019-07-27
Add VictorMono

kule:
#### Description

Added the VictorMono font

#### Requirements / Checklist
 
- [ x ] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)

- [ x ] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [ x ] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [ x ] Scripts execute without error (if necessary):
- [ x ] Extended the README and documentation if necessary, e.g. You added a new font please update the table


kule:
Not sure how to interpret that ci error? If there’s something I need to tweak let me know...

ryanoasis:
Thanks @kule 

You can ignore the build error. That's from some shellcheck violations. To be fixed...

mhartington:
Note that this has the same issue as listed here

https://github.com/ryanoasis/nerd-fonts/issues/254

Because Victor ships with ligatures already, the font needs to either have a config to disable the `fi`/`fl` ligatures. I've attempted this myself, but to no success.

ryanoasis:
Thanks @mhartington 

There is another option that I had forgot about, that is using the `careful` flag. You would lose any glyphs that would stomp over but that would ensure ligatures don't get replaced.

Cross posting reply from #254:

> Sorry for the late reply, yeah I think the real solution will involve not clobbering the ligatures but that will involve moving some of the codepoints of some glyphs. That would probably result in a `3.0` release (breaking change).
> 
> There is another workaround worth mentioning (if one patches themselves). that is using the `--careful` flag. That will ensure existing glyphs don't get clobbered, for example:
> 
> left: original font
> center: straight up patch
> right: patch with the `--careful` flag
> VictorMono-Regular-patching-comparisons
> 
> Ref: [#57 (comment)](https://github.com/ryanoasis/nerd-fonts/issues/57#issuecomment-515654441)

ryanoasis:
I think this looks good, at least good enough to be in `master`. There are issues such as the one mentioned that probably need to be reconciled before a release. That seems out of scope of this PR since other fonts with ligatures currently have the same issue :smile: 

Also looks like you @kule instructions perfectly so appreciate that

ryanoasis:
fixing milestone, need to update changelog and release and all-contributors

PR #366, opened 2019-08-11
Fix typo

ofrank123:
Literation > Liberation in the readme


ryanoasis:
Thanks but the name was renamed on purpose because of the license (particularly the reserved font name of `Liberation`). See https://github.com/ryanoasis/nerd-fonts/blob/master/src/unpatched-fonts/LiberationMono/LICENSE#L4

I know this can be confusing and actually now looking it looks like we are respected the Reserved Font Name of Liberation at all. The patcher will be updated to be 'Li***t***eration' going forward.

PR #380, opened 2019-10-09
Entity error

whizzrahul:


LuRsT:
I'm not sure what this PR is about, could you add some more information to it?

PR #381, opened 2019-10-11
Fix formatting

this-is-you:
#### Description

#### Requirements / Checklist

- [ ] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)


ryanoasis:
Nice fixes, thank you. Looks much better

There is a small problem, the way we build these compiled readmes if from the src folder, in this case it was a txt file and gets compiled into a markdown and that's when the formatting seems to get messed up.

In this case the original coming from https://github.com/ryanoasis/nerd-fonts/tree/master/src/unpatched-fonts/BigBlueTerminal

I think this is a good fix for now, pending possibly a better way to handle combining these patched readmes by the script: https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/standardize-and-complete-readmes.sh

PR #447, opened 2020-02-26
Restore new line in README.ja

delphinus:
This restores a new line that was removed in 8c54efe.


ryanoasis:
I think other readmes are affected as well. Thanks

Issue #323, opened 2019-02-09
Mono fonts lack spacing information in the metadata


resolritter:
#### 🎯 Subject of the issue

As reported in https://github.com/kovidgoyal/kitty/issues/1372, I'm unable to detect mono Nerd Fonts due to the missing 'spacing' attribute, which comes by default on system mono fonts. 

Example is the built-in Nimbus Mono Font Family:

```
        'nimbus mono l': [{
            'path': '/usr/share/fonts/type1/gsfonts/n022004l.pfb',
            'family': 'Nimbus Mono L',
            'style': 'Bold',
            'postscript_name': 'NimbusMonL-Bold',
            'weight': 200,
            'slant': 0,
            'index': 0,
            'scalable': True,
            'outline': True,
            'color': False,
            'spacing': 'MONO'
        }, {
            'path': '/usr/share/fonts/X11/Type1/n022023l.pfb',
            'family': 'Nimbus Mono L',
            'style': 'Regular Oblique',
            'postscript_name': 'NimbusMonL-ReguObli',
            'weight': 80,
            'slant': 110,
            'index': 0,
            'scalable': True,
            'outline': True,
            'color': False,
            'spacing': 'MONO'
        }, {
            'path': '/usr/share/fonts/type1/gsfonts/n022003l.pfb',
            'family': 'Nimbus Mono L',
            'style': 'Regular',
            'postscript_name': 'NimbusMonL-Regu',
            'weight': 80,
            'slant': 0,
            'index': 0,
            'scalable': True,
            'outline': True,
            'color': False,
            'spacing': 'MONO'
        }, {
            'path': '/usr/share/fonts/type1/gsfonts/n022024l.pfb',
            'family': 'Nimbus Mono L',
            'style': 'Bold Oblique',
            'postscript_name': 'NimbusMonL-BoldObli',
            'weight': 200,
            'slant': 110,
            'index': 0,
            'scalable': True,
            'outline': True,
            'color': False
        }, {
            'path': '/usr/share/fonts/X11/Type1/n022024l.pfb',
            'family': 'Nimbus Mono L',
            'style': 'Bold Oblique',
            'postscript_name': 'NimbusMonL-BoldObli',
            'weight': 200,
            'slant': 110,
            'index': 0,
            'scalable': True,
            'outline': True,
            'color': False
        }, {
            'path': '/usr/share/fonts/X11/Type1/n022003l.pfb',
            'family': 'Nimbus Mono L',
[...]

PR #526, opened 2020-09-16
Take the style from the filename if it's longer than the subfamily (fixes #257)

### LINKED ISSUE #257: BlexMono: All italic variants shown as duplicates (on macOS, at least)
#### 🎯 Subject of the issue

macOS states that all italic variants of BlexMono (merged in #253 – tanks a lot!) are duplicates, and regardless of which one you select in the font selector you chose, the "text" variant is shown (besides the bold variant, which you can use, too).

This is not the case with the original IBM Plex font, so I think this is a problem with the patching?

#### 🔧 Your Setup

- Blex Mono Bold Italic Nerd Font Complete.ttf, Blex Mono ExtraLight Italic Nerd Font Complete.ttf, Blex Mono Light Nerd Font Complete.ttf, …
- macOS 10.13.4

#### ★ Optional

duplicate font

duganchen:
The font patcher now takes the style ("Italic", "Regular", etc) from the filename if it's longer (contains more information) than the "subfamily".

#### Description

Updates the font patcher's algorithm for detecting the style (Italic, Regular, etc) in the font's Postscript name, so that it's more accurate.

When I first tested this with IBM Plex Mono 5.3.1, I could confirm that it fixes fixes #257 by successfully giving each generated font a unique Postscript name.

I have since tested it with VictorMono and JetBrainsMono. I have also imported all three into Font Book on OS X, and confirmed that this pull request does, indeed, resolve the issue.

#### Requirements / Checklist

Fixes #257.

The literal command I used to test this is in my comment here:

https://github.com/ryanoasis/nerd-fonts/issues/323#issuecomment-693152613

After that, I did "fc-cache -f -v" and "kitty list-fonts --psnames",  checking that each Blex Mono font file's Postscript name is unique.

I tested the font patcher on Fedora 32. I tested the results on Fedora 32 and on OS X.

https://github.com/ryanoasis/nerd-fonts/issues/257

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -293,6 +293,11 @@ class font_patcher:
         # some fonts have inaccurate 'SubFamily', if it is Regular let us trust the filename more:
         if subFamily == "Regular":
             subFamily = fallbackStyle
+
+        # This is meant to cover the case where the SubFmaily is "Italic" and the filename is *-BoldItalic.
+        if  len(subFamily) < len(fallbackStyle):
+            subFamily = fallbackStyle
+
         if self.args.windows:
             maxFamilyLength = 31
             maxFontLength = maxFamilyLength - len('-' + subFamily)
--- a/font-patcher
+++ b/font-patcher
@@ -294,7 +294,7 @@ class font_patcher:
         if subFamily == "Regular":
             subFamily = fallbackStyle
 
-        # This is meant to cover the case where the SubFmaily is "Italic" and the filename is *-BoldItalic.
+        # This is meant to cover the case where the SubFamily is "Italic" and the filename is *-BoldItalic.
         if  len(subFamily) < len(fallbackStyle):
             subFamily = fallbackStyle
 


font-patcher
```diff
@@ -293,6 +293,11 @@ class font_patcher:
         # some fonts have inaccurate 'SubFamily', if it is Regular let us trust the filename more:
         if subFamily == "Regular":
             subFamily = fallbackStyle
+
+        # This is meant to cover the case where the SubFmaily is "Italic" and the filename is *-BoldItalic.
```
ryanoasis, line 297: can you fix the 'SubFmaily' typo?

font-patcher
```diff
@@ -293,6 +293,11 @@ class font_patcher:
...
+        if  len(subFamily) < len(fallbackStyle):
```
ryanoasis, line 298: Is it a simple length comparison a sufficient check here? Instead of checking sub family is italic _and_ fallbackStyle contains italic _and_ longer in length?

Honestly asking as it's been a while since I worked on the patcher :smile:

ryanoasis (changes requested):
typo change, other comment is open ended question. Thanks for taking on this fix...

font-patcher
duganchen: This should be fixed now.

font-patcher
duganchen: You know, I did fix the typo...

duganchen:
You know, I did fix the typo...

ryanoasis:
Thanks!

PR #533, opened 2020-10-11
Fix font preview link

adrian5:
This appears to be the canonical URL these days. The `app.` subdomain has become a redirect page (with a message explaining that).


ryanoasis:
Thanks! :smile:

Issue #346, opened 2019-04-08
To use a new PUA?


fabrizioschiavi:
Hello Ryan,
could I suggest you to move all the glyphs of the Nerd Fonts into the “new” U+F0000..U+FFFFD [PUA](http://www.unicode.org/faq/private_use.html)?

This request is just to avoid conflicts like this:

Fura:
![Screen Shot 2019-04-08 at 13 24 54](https://user-images.githubusercontent.com/1576663/55720726-cfd7bc00-5a01-11e9-851e-939835d7a525.png)

PragmataPro:
![Screen Shot 2019-04-08 at 13 25 19](https://user-images.githubusercontent.com/1576663/55720732-d2d2ac80-5a01-11e9-9507-c85a7bb11951.png)

In this area there are some important glyphs necessary for Hebrew and Arabic texts, not only in [PragmataPro](https://www.fsd.it/shop/fonts/pragmatapro/) but in every other big font family like Arial, Lucida, DejaVu etc.

I love to use your idea in my fonts but it's necessary to find a new empty area to avoid conflicts. 
All the best!

ryanoasis:
Thanks @fabrizioschiavi I think this is something we will eventually have to do unfortunately. Perhaps next major version (3+)

ryanoasis:
There are some good suggestions and discussion around the conflicts done here: #365 just to link the issues

Finii:
With `v2.3.0` we started the transition to codepoints `F0000` according to plan 5+ of #365.

With `v3.0.0` the offending codepoints will be dropped.

Closing this already now, as everything is prepared and just waits for rollout.

Issue #365, opened 2019-08-02
[Suggestion] Fix invalid code points for some glyphs


delphinus:
## Summary

The current builds overwrites some code points that Unicode Consorthium prohibits to use for custom glyphs. I want to change this by fixing font_patcher.

## Problem Detail

Unicode defines [Private Use Areas][pua] and the consorthium itself decide not to add characters on these areas. So we can add glyphs as we like.

[pua]: https://en.wikipedia.org/wiki/Private_Use_Areas

The areas has these code points: U+E000..U+F8FF, U+F0000..U+FFFFD, U+100000..U+10FFFD.

But the Nerd Fonts is overwriting more code points than them. The font_patcher writes glyphs as below.

* source: the ones from source font files.
* current: the ones font_patcher writes into.

| Font Name               |   source  |    current    |   plan 1  |          plan 2         |
|-------------------------|:---------:|:-------------:|:---------:|:-----------------------:|
| Seti-UI + Custom        | E4FA-E52E |   E5FA-E62E   |     ←     |            ←            |
| Devicons                | E600-E6C5 |   E700-E7C5   | E630-E6F6 | E700-E7C5 (not changed) |
| Powerline Symbols       | E0A0-E0B3 |       ←       |     ←     |            ←            |
| Powerline Extra Symbols | E0A3-E0D4 |       ←       |     ←     |            ←            |
| Pomicons                | E000-E00A |       ←       |     ←     |            ←            |
| Font Awesome            | F000-F2E0 |       ←       |     ←     |            ←            |
| Font Awesome Extension  | E000-E0A9 |   E200-E2A9   |     ←     |            ←            |
| Power Symbols           | 23FB-2B58 |       ←       |     ←     |            ←            |
| Material                | F001-F847 | **F500-FD46** | E700-EF47 |   F500-F8FF,E800-EC47   |
| Weather Icons           | F000-F0EB |   E300-E3EB   |     ←     |            ←            |
| Font Logos (Font Linux) | F100-F11C |   F300-F31C   |     ←     |            ←            |
| Octicons                | F000-F105 |   F400-F505   |     ←     |            ←            |
| Octicons                | 2665-2665 |       ←       |     ←     |            ←            |
| Octicons                | 26A1-26A1 |       ←       |     ←     |            ←            |
| Octicons                | F27C-F27C |   F4A9-F4A9   |     ←     |            ←            |

[...]

PR #609, opened 2021-04-14
Update Symbols-2048-em Nerd Font Complete.ttf

fabrizioschiavi:
#### Description

I just moved all the icons from Unicode Ranges used for Arabic, standard ligatures and other important sets to empty and clean F0000-FFFFD private set where is possible add more than 65,000 glyphs!
The positioning consider the actual codepoint, (e.g. E001 was changed to F0001) to limited the renaming process with the changes of some prefixes

#### Requirements / Checklist

- [v] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)

Avoid conflicts with the necessary glyphs of complete typefaces

Opening the changed ttf file

This is an important and necessary change because complete fonts like PragmataPro can be include all the glyphs possible without conflicts with very useful Nerd Font icons. Please don't ignore this request, it's important for all.

https://github.com/ryanoasis/nerd-fonts/issues/346

Actual:
before

After pull request:
after


polyzen:
I believe you're meant to edit the [1000-em](https://github.com/ryanoasis/nerd-fonts/blob/master/src/glyphs/Symbols%20Template%201000%20em.ttf) and [2048-em](https://github.com/ryanoasis/nerd-fonts/blob/master/src/glyphs/Symbols%20Template%202048%20em.ttf) template files that are used in generating the complete symbols font files.

fabrizioschiavi:
I changed just one file. I hope in the help of the community

andrew-t-james-hm:
+1

TheSecEng:
Any updates on this PR?

andrew-hm:
I see no objections to this PR, hoping it can be merged by end of May 2021?? Just in time for summer!!

Finii:
Well, I believe this MR has two points to be solved before it can be merged.

1. The technical ascpect is far from complete here. As @polyzen already pointed out it does not really help to patch some font file. What about all the patched fonts? Shall they not follow the new codepoints? This also means changes in the `font-patcher`.

2. With this all powerline setups will break. People need to update their prompts; programs that show icons need to be recompiled, powerline templates have to be updated, ... etc pp. The list is endless.

Because of 2 I believe the MR in its deceivingly simple logic (move `uE000` to `uF0000`) will not come. I personally would not want that. Albeit there is a problem with some (only a few) symbols we add that are outside the _private_ areas. Probably they and only they will be shifted ... somewhen.

See Issue #365

wismill:
As a first step, could these symbols be duplicated in Arabic range _and_ PUA? Then mark the Arabic range as deprecated in order to give time to the tools to migrate (one year?). Once done, we could just free the legacy Arabic range.

Finii:
Note possible interference with
* #801

Finii:
This is superseeded by
* #773

Which will solve the PUA problems that we discussed in a lot of threads all over, in a different way than the one suggested with the PR (i.e. `Plan 5`).

Finii:
Thank you @fabrizioschiavi for the PR, as it was part of the solution finding process! :+1:

Issue #608, opened 2021-04-11
Font patcher fails when --outputdir not set


adrian5:
Assuming this is a hard requirement, it would be good if the script checked this before starting any processing. I noticed that the script fails at the _end_ of font building if no `-out / --outputdir` is supplied.

Personally, I wouldn't mind if the script fell back to writing to the working directory if no path was provided.

adrian5:
Looks like it's [intended](https://github.com/ryanoasis/nerd-fonts/blob/master/font-patcher#L145) to fall back the way I imagined. No idea why it errors out instead.

ryanoasis:
Perhaps a permissions issue?

adrian5:
Since explicitly passing the output directory resolves this, I can't think of the way permissions might be at fault here. I can write anywhere (and from anywhere) as long as I give it a path.

What I did notice though is that a `-out .` (working dir) doesn't work, and that might be what the script defaults to internally. Made sure to use Bash, in case this was some shell idiosyncrasy.

My directory structure for the test is:

```
├── CascadiaMonoPL-Regular.ttf
├── CascadiaMonoPL-SemiBold.ttf
├── FontForge.AppImage                (github.com/fontforge/fontforge/releases)
├── font-patcher                      (github.com/ryanoasis/nerd-fonts/blob/master/font-patcher)
└── src
    └── gylphs
        ├── devicons.ttf
        ├── font-awesome-extension.ttf
        └── etc.
```

And then I invoke a variation of the following:

```
./FontForge.AppImage -script $PWD/font-patcher $PWD/CascadiaMonoPL-Regular.ttf --powersymbols
```

adrian5:
@ryanoasis I looked into this some more and I think this could be due to launching the process via the FontForge AppImage. If the relative path `.` places the current working directory _inside_ the runtime AppImage hierarchy, then it's plausible that writing would fail.

I'm not sure there's a straight forward (but flexible) workaround. Since this is a fairly niche case, I doubt you'd want to add an explicit AppImage check to your script.

Nonetheless, with distro packages are often severely outdated, I suspect a lot of manual patching on Linux will be done using the AppImage. Maybe adding a short note to [nerd-fonts#font-patcher](https://github.com/ryanoasis/nerd-fonts#font-patcher) would be helpful. I can send a PR if you're OK with that.

ryanoasis:
Considering this closed with the additional readme addition

PR #624, opened 2021-05-11
Add AppImage patching instructions to Readme

adrian5:
To prevent surprises as described in [#608](https://github.com/ryanoasis/nerd-fonts/issues/608).

I've taken the liberty to adjust the formatting a bit beneath bullet points like so:

---

* Before

  Some text here and...

```
./also.sh --some "code"
```

* After

  Some text here and...

  ```
  ./also.sh --some "code"
  ```

---

Feel free to revert this if you prefer the prior look.


adrian5:
@ryanoasis Can you just merge this, unless something's preventing it?

ryanoasis (approved):
Thanks for this!

PR #644, opened 2021-07-25
font-patcher: remove_ligatures() never displays warnings

fracsinus:
https://github.com/ryanoasis/nerd-fonts/blob/bc4416e176d4ac2092345efd7bcb4abef9d6411e/font-patcher#L416-L419

It seems that the `elif` block is meant to check the case where:
  - `--removeligatures` is set
  - AND no config file is provided.

I guess those blocks are over-indented? For now, no warning or info message is printed even in the very case.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -413,10 +413,10 @@ class font_patcher:
                     print("Successfully removed subtable:", subtable)
                 except Exception:
                     print("Failed to remove subtable:", subtable)
-            elif self.args.removeligatures:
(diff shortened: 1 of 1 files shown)


Finii:
@fracsinus I would change the title to state what is the effect of the issue; at first I thought that the glyphs are indentat in the font file or something ;-D

Something along the lines 'Remove ligature warnings never displayed' or so.

ryanoasis:
yep that's how it was/is supposed to be

PR #654, opened 2021-08-18
Update docker-entrypoint.sh

diddledani:
* Add quotes around the variable containing the detected file name in `/in` to allow for spaces in the file names.


ryanoasis (approved):
good find

Issue #175, opened 2017-07-07
.ttc files don't work


lilyball:
#### Subject of the issue

The font patcher doesn't handle .ttc files correctly. It produces a weirdly-named file that seems to contain the data for the first font in the .ttc file, though Font Book.app says it doesn't pass validation.

I tried this using /System/Library/Fonts/Menlo.ttc.

FWIW here's an [old unmerged PR](https://github.com/powerline/fontpatcher/pull/6) that adds support for .ttc files to the Powerline font patcher.

#### Your Setup

- macOS 10.12.5

ryanoasis:
@kballard Okay thanks :smile: I wasn't aware of this.

It would be nice to support this eventually provided we get some help :blush:, however I would say this is not a high priority

taohex:
I encountered this problem, too. Menlo is the default font in macOS, I think it's important.

RuRo:
Same problem on Manjaro Linux with the default `Noto Sans` font.  
My workaround is to use `fontforge` to convert `.ttc` to `.ttf` before patching the font:  
`fontforge -lang=pe -c "Open(\$1);Generate(\$1:r+'.ttf')" SomeFont-Regular.ttc`

ryanoasis:
I have some renewed interest in supporting this :smile:

LuRsT:
We could add that bit that @RuRo mentioned to the docs or make this a future feature, since you have renewed interest in supporting this, is it going to be in the roadmap @ryanoasis ?

ryanoasis:
>  since you have renewed interest in supporting this

Oh I did say that didn't I? :laughing: 

I think we can try to target v2.1.0 (i.e. next release) unless I am missing it looked somewhat straight forward to implement.

benfrain:
Could this be the reason I can't patch `Iosevka Slab` following the readme? https://github.com/ryanoasis/nerd-fonts/discussions/735

PR #664, opened 2021-09-13
Consider .ttc in docker entrypoint

kraftwerk28:
#### Description

Adds one more glob to docker-entrypoint.sh

#### Requirements / Checklist

By running `docker run ...` (https://github.com/ryanoasis/nerd-fonts#option-8-patch-your-own-font)

Iosevka font has .ttc file extension to the script won't work


kraftwerk28:
I wonder if it's reasonable to just write `for f in /in/* ...`, since there are a lot of font file extensions

ryanoasis:
> I wonder if it's reasonable to just write `for f in /in/* ...`, since there are a lot of font file extensions

That's a good point. I say this fix is a good improvement and later we can make it less selective if it gets too unwieldy

ryanoasis:
I think this is fine, just note that `ttc` not supported: #175

PR #671, opened 2021-09-26
misleading _Note_: Only for Linux & macOS (OS X) | Option 3: Install Script

GiorgosXou:
There's the `./install.ps1` for windows. *(or am I missing something?)*


ryanoasis:
Good catch. Thanks!

Issue #663, opened 2021-09-10
Victor Mono Complete - Italic/oblique variants seen as "duplicate" in MacOS


McAllaster:
#### 🎯 Subject of the issue

> MacOS seems to think all italic variants are the same. It doesn't differentiate between the various weights and instead suggests to delete all but one. Same for all the Oblique variants. No problem with the non-italic variants, though.

_Originally posted by @mbirth in https://github.com/ryanoasis/nerd-fonts/issues/494#issuecomment-877359708_

Looks like a similar issue to #257 to me. Possibly related to #494, but I assume they are different enough to warrant their own issues.

#### 🔧 Your Setup

- VictorMono Nerd Font Complete
- Alacritty, Kitty, iTerm, etc. all present the same issues
- MacOS 11.5.2

#### ★ Optional

<img width="812" alt="Screen Shot 2021-09-10 at 12 03 16 PM" src="https://user-images.githubusercontent.com/5580190/132883495-273e870d-29b7-4434-9732-b73931d65c0d.png">

pr0PM:
Can confirm have seen this for NerdFont Firacode and all it's variants on MacOS!

Finii:
Lets see which fonts we have here and how they are named.
Note that programs look on Family Name and Style, Weight is 'informal'.

* Family Name
    * Style / Weight (Human Name)
    
* **Victor Mono Nerd Font**
    * **Italic** / Book (Victor Mono Italic Nerd Font Complete)
    * **Italic** / Light (Victor Mono Light Italic Nerd Font Complete)
    * **Bold Italic** / Bold (Victor Mono Bold Italic Nerd Font Complete) (! Postscript name invalid)
    * **Italic** / Thin (Victor Mono Thin Italic Nerd Font Complete)
    * **Italic** / Demi (Victor Mono SemiBold Italic Nerd Font Complete)

:unamused: 

There are two solutions:
1. Have one family name per weight
1. Combine weight and style into subfamily (in the Style field)

Here is how it is handled by the original font (they choose option 1)

* **Victor Mono Thin**
    * **Italic** / Thin (Victor Mono Thin Italic)
* **Victor Mono Light**
    * **Italic** / Light (Victor Mono Light Italic)
* **Victor Mono**
    * **Italic** / Book (Victor Mono Italic)
* **Victor Mono SemiBold**
    * **Italic** / Demi (Victor Mono SemiBold Italic)
* **Victor Mono**
[...]

PR #690, opened 2021-11-25
font-patcher: Really set SubFamily to what we want

Finii:
#### Description

**[why]**  
The most convenient way to set something in the SFNT table is by using
Fontforge's `appendSFNTName()`. When we try to set the _SubFamily_, this
will succeed in many instances, but not in all. Sometimes the function
seems to be ignored.

This is for example the case for `VictorMono-Medium.ttf` where we want to
set the _SubFamily_ `"Regular"`, but the final patched file has the
_SubFamily_ `"Medium"`.

**[how]**  
The reason for the unexpected behavior of Fontforge is as follows. It
almost seems like a bug, but then it is coded explicitely in this way.
Maybe noone ever stumbled about it, or it has only been designed only
with creating new fonts in mind.

Fontforge lets you set any value, unless it is the default value. If it
is the default value it does not set anything. It also does not remove
a previously existing non-default value. Why it is done this way is
unclear:
```
  fontforge/python.c SetSFNTName() line 11431
    return( 1 ); /* If they set it to the default, there's nothing to do */
```

Then is the question: What is the default? It is taken from the
currently set `fontname` (??!). The `fontname` is parsed and everything
behind the dash is the default _SubFamily_:
```
  fontforge/tottf.c DefaultTTFEnglishNames()
  fontforge/splinefont.c _GetModifiers()
```

To stick to the example above, the `fontname` is there
`"VictorMonoNerdFontM-Medium"`. Afterwards we want to set the _SubFamily_ to
`"Medium"` - which is the 'default' and Fontforge decides that nothing has
to be done. Albeigth the current SubFamily is `"Regular"` (from the source
font load).

Well.

To fix this without touching Fontforge we need to set the _SubFamily_
directly in the SFNT table, which is a little bit unhandy but possible.

Now in the generated files the _SubFamily_ is again as expected the same
es the `fontname`'s last part.

#### Requirements / Checklist

In progress of fixing issue #663 
`font-patcher` run with debug stuff enabled, see https://gist.github.com/Finii/a496cff7eb38933b169110be3e4e8af2

image

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -394,7 +394,12 @@ class font_patcher:
         self.sourceFont.appendSFNTName(str('English (US)'), str('Preferred Family'), self.sourceFont.familyname)
         self.sourceFont.appendSFNTName(str('English (US)'), str('Family'), self.sourceFont.familyname)
         self.sourceFont.appendSFNTName(str('English (US)'), str('Compatible Full'), self.sourceFont.fullname)
-        self.sourceFont.appendSFNTName(str('English (US)'), str('SubFamily'), subFamily)
+        # Fontforge does not allow to set SubFamily to any value (see commit message), so
+        # instead of self.sourceFont.appendSFNTName(str('English (US)'), str('SubFamily'), subFamily) we do:
+        sfnt_list = list(self.sourceFont.sfnt_names)
+        sfnt_list[subFamilyTupleIndex] = ( sfnt_list[subFamilyTupleIndex][0], sfnt_list[subFamilyTupleIndex][1], subFamily )
+        self.sourceFont.sfnt_names = tuple(sfnt_list)
+
         self.sourceFont.comment = projectInfo
         self.sourceFont.fontlog = projectInfo
 


Finii:
I guess in the end this fix will not be needed.
Leave it in for now, to document that the code seems to be doing something that it is in fact not.

I still fight with tables with all naming details of all 741 font, while in parallel setting up Christmas stuff for the children ;-)

Do not really like the `fontforge` behavior, maybe should MR there. Anyhow, the unexpected-behavior safes us as we can not overwrite the SubFamily with the ridiculous stuff we want to write ;)

Finii:
Here some examples:

image

`Overpass Extrabold` for example, we rip out the weight from the family name and try to stuff it into the SubFamily, that fails, and so the information is lost ;-)

Finii:
Just that I do not forget..

Other ppl have the same problem with `fontforge` and write their own SFNT accessor:
image

https://github.com/ToxicFrog/Ligaturizer/blob/master/ligaturize.py

Issue #691, opened 2021-11-27
Trying to convert latest Cascadia Code PL font, Nerd Fonts cannot open font


ThaDaVos:
#### 🗹 Requirements

- [X] A brief but descriptive _title_ of your issue
- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [X] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [X] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

#### 🎯 Subject of the issue

Trying to convert eh Cascadia Code font myself with the following command;
`fontforge -script "/home/dylan/projects/font-patcher/nerd-fonts/font-patcher" "/home/dylan/projects/font-patcher/CascadiaCode/otf/static/CascadiaCodePL-ExtraLight.otf" -c --careful -out ./out`

But it throws an error and I cannot figure out what's going wrong

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_ CasCadia Code PL
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_ Tmux inside WSL 2
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ Linux through Windows -> WSL 2

#### ★ Optional

![image](https://user-images.githubusercontent.com/5251028/143691094-4b6fa3bc-3532-46c2-b43e-18e051d93e05.png)

ryanoasis:
This should be fixed now. Please re-try.

see my comment for more info: https://github.com/ryanoasis/nerd-fonts/pull/687#issuecomment-980831193

Finii:
Ah, `fontforge` changed:

![image](https://user-images.githubusercontent.com/16012374/143763144-3e4f01cd-6a55-4056-a903-4355d2f495dd.png)

Lets see how we keep the `fstype` fix for older `fontforge`s...

Finii:
They added the plain name options only in 2020: https://github.com/fontforge/fontforge/commit/4a76712f0c74090f8aaab7ba67612d1941f05779

![image](https://user-images.githubusercontent.com/16012374/143763395-afc17ed6-a06d-44dc-bf15-162bcd35c7cb.png)

[...]

PR #687, opened 2021-11-24
Bugfix/fstype font does not open

Finii:
#### Description

_Please explain the changes you made here._

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [x] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [x] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### What does this Pull Request (PR) do?

2 commits:

##### font-patcher: Allow processing of fonts with fsType set
    
[why]
Through fsType certain restrictions can be set on a font. When fontforge
is used in interactive mode the user can override the restrictions with
a popup dialogue. The font-patcher script dies instead, without any
meaningful message.
    
[how]
Allow the script to ignore fsType settings when opening.
The restrictions will still persist into the generated patched font.
    
[note]
This came up with Bicubik by Anton Kudin, that has fsType = 2
(modification restriction) set.
    
Fixes: #686

##### font-patcher: Fail with meaningful message if font can not be opened
    
[why]
When fontforge is not able to open the font we fail with a meaninless
exception. Users might think that the font-patcher script itself is
broken.
    
[how]
Exit the script with a hint how to get more information if fontforge was
not able to open the font.

Issue #686, opened 2021-11-22
Font-patcher fails to patch font: OSError


ExperiBass:
#### 🗹 Requirements

- [x] A brief but descriptive _title_ of your issue
- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

#### 🎯 Subject of the issue

I'm attempting to patch a font I have, but the script fails with this error:
```
fontforge -script ./font-patcher ./Bicubik-capital.otf -out ./ -l --fontawesomeextension --fontawesome --fontlinux --octicons --powersymbols --pomicons --powerline --powerlineextra --material --weather
Copyright (c) 2000-2021. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20201107
 Based on sources from 2021-11-16 19:50 UTC-ML-D-GDK3.
 Based on source from git with hash: 58238cc173c80379d0d4616971fc793da26c6aa4
This font contains both a 'kern' table and a 'GPOS' table.
  The 'kern' table will only be read if there is no 'kern' feature in 'GPOS'.
Traceback (most recent call last):
  File "/home/gingka/Builds/nerd-fonts/./font-patcher", line 936, in <module>
    main()
  File "/home/gingka/Builds/nerd-fonts/./font-patcher", line 930, in main
    patcher = font_patcher()
  File "/home/gingka/Builds/nerd-fonts/./font-patcher", line 54, in __init__
    self.sourceFont = fontforge.open(self.args.font)
OSError: Open failed
```

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_ Bicubik-Capital, modified from Bicubik (cant find the source)
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_ kitty
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ Linux, EndeavourOS

Finii:
```
[...]

PR #692, opened 2021-11-28
font-patcher: Reintroduce fsType fix

Finii:
#### Description

**[why]**
With commit
  f7d6fcb5 `font-patcher: Allow processing of fonts with fsType set`

we added support for fonts with the fsType set. This came up in
issue #686 with font 'Bicubik'.

The solution in that commit uses (modern) textual flags in the
`fontforge` open() method. But they have been only introduced in 2020,
so people using older `fontforge` could not patch anything anymore.
This has been reported in issue #691.

As a quick fix the fsType support has been removed with commit
  ab6fa3c5 `Reverts part of #687 * the patcher refuses to patch all/most fonts with this flag in the open options`

**[how]**
Revert ab6fa3c5 but use the old fashioned numerical open flags
interface instead.

**[note]**
The textual open() flags have been introduced into `fontforge`s python
interface with their commit

https://github.com/fontforge/fontforge/commit/4a76712f0c74090f8aaab7ba67612d1941f05779
```
  Font Open flag improvements
  * Document more Open flags
  * Add string tuple interface to python FontOpen API
```
#### Requirements / Checklist

Reintroduce `fsType` fix as per f7d6fcb5 but avoiding problem with older `fontforge`

Try to patch some font from `src/`, try to patch `Bicubik` (or other `fsTyoe` set font).
Use `fontforge` revision 20190801 or older, **and** revision 20200314 or newer.
Just the ability to open the font is enough, no need to inspect the patched fonts.

_Edit: Just formatting changes_
_Edit: Add version requirements for tests_

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -52,7 +52,7 @@ class font_patcher:
         self.setup_arguments()
         self.config = configparser.ConfigParser(empty_lines_in_values=False, allow_no_value=True)
         try:
-            self.sourceFont = fontforge.open(self.args.font)
+            self.sourceFont = fontforge.open(self.args.font, 1) # 1 = ("fstypepermitted",))
         except Exception:
             sys.exit(projectName + ": Can not open font, try to open with fontforge interactively to get more information")
         self.setup_font_names()


ryanoasis:
I think this could be a variable with the value instead of comment, but works well and tested out. Thanks.

Issue #509, opened 2020-07-30
`fi` shows as a telephone with `Inconsolata NF`


mangelozzi:
#### 🎯 `fi` shows as a telphone with `Inconsolata NF`

Type `file`, and you see `<telephone symbol>li`

#### 🔧 Your Setup

-  sudo scoop install inconsolata-nf
- windows terminal
- Windows 10 .19041

#### ★ Optional

![image](https://user-images.githubusercontent.com/19764314/88912861-38365680-d260-11ea-8722-505d62470866.png)

Tremeschin:
Same here on Arch Linux, but somehow only italics "fi" 's show as a ☎️

gardotd426:
This is actually an issue with other nerd fonts as well. I was on the Arch Linux forums and saw all these telephones and was like "huh?" I had a nerd font set as my monospace font.

unfug:
Is there a workaround or can anyone release new version who know how to deal with fontpatcher?

p-kolacz:
Ubuntu Nerd Font has the same problem.

gardotd426:
Yeah it's definitely not just Inconsolata, I've seen it with numerous
monospace nerd fonts

jffz:
Any temporary workaround to this ? Maybe disable ligature for a special unicode ?

mangelozzi:
> Any temporary workaround to this ? Maybe disable ligature for a special unicode ?

I ended up just patching my own font.

jffz:
Any hint as how to do so ?

mangelozzi:
> Any hint as how to do so ?

https://github.com/ryanoasis/nerd-fonts#option-8-patch-your-own-font

jef:
Duplicate https://github.com/ryanoasis/nerd-fonts/issues/254

ewanm89:
Isn't Nerd Font Mono supposed to have ligatures disabled, this is happening there too? It is annoying cause I like the striped 0 of Inconsolata. And got things to do rather than work out how to patch without ligatures mystelf.

JunjieLeiCoe:
Same problem here...
I am using `monaco nerd font` -- Mac OSX, native terminla, not zsh,  has the same issue

AlbeyAmakiir:
Same for Overpass and MPlus. Also, in MPlus (but not Overpass) I'm getting a different pair of symbols with `ffi`, like you find in "office". 
[...]

PR #711, opened 2021-12-13
font-patcher: Remove obsolete metadata on glyph exchange

Finii:
**[why]**
When we overwrite a glyph that originally had some special handling, be
it a substitution or position table entry (GPOS/GSUB), that special
handling is usually not appropriate anymore and has to be removed.

If we need special lookup table entries for the new glyph we would have
to add them later anyhow, because we can not rely on their existance.

In Issue #509 it was a ligature entry, that replaced 'f' followed by 'i'
with the 'fi' ligature. The ligature glyph is overwritten by us with a
telephone symbol and the substitution table entry makes no sense
anymore.

**[how]**
If we overwrite a preexisting codepoint we remove it from all lookup
tables.

Thanks to all other reporters with details.

Fixes: #509 #254

Reported-by: mangelozzi <mangelozzi@gmail.com>

#### Requirements / Checklist

Remove obsolete ligature (and other) information on any glyph we replace.

Patch for example
* Overpass
* M+ (also `ffi`)
* Inconsolata
* Ubuntu
* Monaco
* Share Tech Mono (?)

and look if the `fi` ligatures were removed: Either in `fontforge` or by installing the font and typing text in a ligature aware environment (i.e. `writer`) and check that there is no ligature for `fi` anymore.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -706,16 +706,20 @@ class font_patcher:
             sym_dim = get_glyph_dimensions(sym_glyph)
 
             # check if a glyph already exists in this location
+            if copiedToSlot.startswith("uni"):
+                copiedToSlot = copiedToSlot[3:]
+            codepoint = int("0x" + copiedToSlot, 16)
             if careful or 'careful' in sym_attr['params']:
-                if copiedToSlot.startswith("uni"):
-                    copiedToSlot = copiedToSlot[3:]
-                codepoint = int("0x" + copiedToSlot, 16)
                 if codepoint in self.sourceFont:
                     if self.args.quiet is False:
                         print("  Found existing Glyph at {}. Skipping...".format(copiedToSlot))
-
                     # We don't want to touch anything so move to next Glyph
                     continue
+            else:
+                # If we overwrite an existing glyph all subtable entries regarding it will be wrong
+                # (Probably; at least if we add a symbol and do not substitude a ligature or such)
+                if codepoint in self.sourceFont:
+                    self.sourceFont[codepoint].removePosSub("*")
 
             # Select and copy symbol from its encoding point
             # We need to do this select after the careful check, this way we don't


ryanoasis:
This looks similar to what I came up with but just cleaner 😊 . I have one question though..

I had this extra check but I think it may be redundant:

```py
isSymbolALigature = fontforge.IsLigature(int(copiedToSlot, 16))
if isSymbolALigature:
  self.sourceFont[currentSourceFontGlyph].removePosSub("*")
```

thoughts on `fontforge.IsLigature` any familiarity?

Finii:
I guess we want to remove any special handling on that glyph, not just ligature related stuff.

For example if a position variation had been imposed on the original glyph, it will be most likely be inappropriate for the new glyph. At least that was my reasoning. Any special instructions regarding that glyph must be obsolete if there is something else. Even if we could reuse that table entries, we would need to provide them along with the glyph a-new, because we can not rely on all fonts havoing them.
So if the new glyph needs some table entries we need to provide them and can not safely reuse old entries. Old entries are just ... old.

gardotd426:
Unfortunately the issue still persists on Inconsolata Nerd Font. 

I literally erased my entire `~/.local/share/fonts/NerdFonts` directory, did a git pull in the repo, then ran `./install.sh` and reinstalled everything. Inconsolata still shows:

Screenshot_20211216_003438

gardotd426:
Oh and I checked via `fc-cache`, I have no other Inconsolata Nerd Fonts located in any other locations. The only ones are the ones from the new `./install.sh`.

ryanoasis:
@gardotd426 Ah that's because the fonts need to be rebuilt.

Only a limited subset of fonts are being rebuilt with changes to the patcher currently and Inconsolata isn't one of them.

See the assets in https://github.com/ryanoasis/nerd-fonts/releases/tag/2.2.0-RC for the current list

gardotd426:
Ah I was gonna say, some of the fonts that I knew used to show the issue
weren't showing it anymore, so I was confused. Alright well that explains a
lot, thanks.

ryanoasis:
Yeah sorry if it's confusing. You'd have to patch it yourself at this moment to get the fix.
Maybe I'll add Inconsolata to the RC so we can prove it out 😄

marcelluseasley:
This repo has it fixed: https://github.com/ToxicFrog/Ligaturizer/releases

Issue #641, opened 2021-07-10
Font patch error for SourceHanCodeJP


MartinP7r:
Hello and thank you for providing this tool!

I'm getting basically the same error as #412 I think.

```bash
// lots of unicode errors above
No glyph with unicode U+05c60 in font
No glyph with unicode U+06753 in font
No glyph with unicode U+06eba in font
No glyph with unicode U+07c3e in font
Traceback (most recent call last):
  File "/root/nerd-fonts/font-patcher", line 908, in <module>
    main()
  File "/root/nerd-fonts/font-patcher", line 902, in main
    patcher = font_patcher()
  File "/root/nerd-fonts/font-patcher", line 55, in __init__
    self.setup_font_names()
  File "/root/nerd-fonts/font-patcher", line 267, in setup_font_names
    fullname = self.sourceFont.fullname + verboseAdditionalFontNameSuffix
TypeError: unsupported operand type(s) for +: 'NoneType' and 'unicode'
```

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_ SourceHanCodeJP
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_ iterm
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ docker

ryanoasis:
Are you still having this problem with the latest version of the patcher or the docker registry?

#439 should've fixed the issue if it was indeed the same as #412

MartinP7r:
Thank you @ryanoasis for having a look at this!  
I honestly haven't tried since then.  
I'm going to check it out towards the end of the month when I'm back from vacation.  
Feel free to close this and I'll re-open if necessary.

ryanoasis:
Thanks. Sounds like a fair plan. Enjoy!

Closing for now

Finii:
Works on my machine when you give the extension.

![image](https://user-images.githubusercontent.com/16012374/146230224-8abe5614-fd30-4dec-b76f-4f5b4b3a36b7.png)

There is clearly something wrong with the extension extraction stuff whatever ;)

Finii:
Edit: Although the line seems to be, as suspected, before #439.

Maybe we should change the issue-template to include information on font-patcher version.

The line number in the error message is clearly Jimmy Haas' commit (i.e. pre #439)

[...]

Issue #412, opened 2019-12-17
Font patcher error with Noto Sans CJK


shiro:
Hello, I'm trying to patch Noto Sans CJK JP, but the script dies every time, it worked fine on Droid Sans for reference.
Here's the dumped stack trace:

```
No glyph with unicode U+051a4 in font
... (lots of same errors) ...
No glyph with unicode U+05be7 in font
No glyph with unicode U+05c60 in font
No glyph with unicode U+06753 in font
No glyph with unicode U+06eba in font
No glyph with unicode U+07c3e in font
Traceback (most recent call last):
  File "font-patcher", line 896, in <module>
    main()
  File "font-patcher", line 890, in main
    patcher = font_patcher()
  File "font-patcher", line 55, in __init__
    self.setup_font_names()
  File "font-patcher", line 267, in setup_font_names
    fullname = self.sourceFont.fullname + verboseAdditionalFontNameSuffix
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
```

NOTE: I'm on manjaro linux and am using the latest patcher from git.

any help is greatly appreciated!

#### 🔧 Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_ Noto Sans CJK JP
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_ st
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ Manjaro linux

onokatio:
I have same problem.

onokatio:
This patch fixes the problem temporary. But it is probaly fontforge's bug.

```
diff --git a/font-patcher b/font-patcher
index ac2d6390..44695119 100755
--- a/font-patcher
+++ b/font-patcher
@@ -52,6 +52,10 @@ class font_patcher:
         self.setup_arguments()
         self.config = configparser.ConfigParser(empty_lines_in_values=False, allow_no_value=True)
         self.sourceFont = fontforge.open(self.args.font)
+        if(self.sourceFont.fullname == None):
+            self.sourceFont.fullname = self.sourceFont.fontname
+        if(self.sourceFont.version == None):
+            self.sourceFont.version = ""
         self.setup_font_names()
         self.remove_ligatures()
         make_sure_path_exists(self.args.outputdir)

```

shiro:
I tried the patch, but I just get a different error now:
```
...
[...]

PR #720, opened 2021-12-15
font-parser: Allow patching without --ext

Finii:
**[why]**
Sometimes fontforge returns `None` for `font.path`. That should not be the
case according to specs:
```
font.path
    (readonly) Returns a string containing the name of the file from
    which the font was originally read (in this session), or if this
    is a new font, returns a made up filename in the current directory
    named something like “Untitled1.sfd”. See also font.sfd_path.
```
This seems to be the case for fonts that do not have a `fullname` set.
I did not search for nor file any issue at Fontforge.

**[how]**
In fact we already have the original font file name, and we want to
retain its extension anyhow (if nothing is specified), so we use the
filename that we opened to determine the extension.

**[note]**
Related: #412
Related: #641

**[note]**
This was the sole usage of font.path.
Has been introduced with commit d8b760aee which looks uncritical.

#### Requirements / Checklist

Use the specified filename directly to detect the filename instead of the round trip through Fontforge that somehow sometimes fails.

Patch for example `SourceHanCodeJP-Regular.otf` or `Noto Sans CJK Bold.otf`

#412 #641

_Edit: Typo_

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -69,7 +69,7 @@ class font_patcher:
         self.sourceFont.encoding = 'UnicodeFull'  # Update the font encoding to ensure that the Unicode glyphs are available
         self.onlybitmaps = self.sourceFont.onlybitmaps  # Fetch this property before adding outlines. NOTE self.onlybitmaps initialized and never used
         if self.args.extension == "":
-            self.extension = os.path.splitext(self.sourceFont.path)[1]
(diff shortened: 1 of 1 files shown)


Finii:
I did try Fontforge `HEAD` which still returns `None`.

ryanoasis:
Tested locally just to confirm. Seems to work well. Thanks! 🏅

Issue #703, opened 2021-12-03
Overpass Nerdfont kerning is incorrect


InfoSec812:
#### 🗹 Requirements

- [ ] A brief but descriptive _title_ of your issue
- [ ] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [ ] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [ ] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

#### 🎯 Spacing issues in Overpass Nerdfont

When I use Overpass Nerdfont the spacing between letters is mishandled.

![image](https://user-images.githubusercontent.com/99691/144656035-761f1f72-db04-4151-8c61-dd81ea8047f1.png)

#### 🔧 Your Setup

- Overpass Nerdfont Mono
- It happens in all of them on Linux as well as in my IDEs
- Linux

ryanoasis:
It's strange because even the unpatched Overpass is displaying strangely in gnome terminal for me just like your screenshot.

![Selection_097](https://user-images.githubusercontent.com/8083459/146798792-96dbcf50-010d-4a2d-8e17-ddfcf1b96741.png)

However something like gedit both patched and unpatched Overpass look fine 🤔 

![Selection_098](https://user-images.githubusercontent.com/8083459/146798767-f2414230-c9dc-4bf5-8977-ed7f67ecab0f.png)

InfoSec812:
Hmmm. I have no issues with the unpatched Overpass Mono from Google Fonts. Also, I have issues with the patched version in EVERY application, not just Gnome terminal.

InfoSec812:
Example:

Konsole with Overpass Mono (unpatched)
![image](https://user-images.githubusercontent.com/99691/146974444-76094e9f-1eae-4a52-a0e5-df1850bc5b46.png)

Konsole with Overpass Mono (patched)
![image](https://user-images.githubusercontent.com/99691/146974487-b83707d9-2b00-4fef-818b-456795bad698.png)

Finii:
Hmm, dowloaded Overpass-Mono from G-Fonts and patched with HEAD. Shows ok in Tilix.

![image](https://user-images.githubusercontent.com/16012374/146987128-f8e2c049-ac55-459c-a6d1-e426b6d2afe3.png)


[...]

PR #695, opened 2021-11-29
font-patcher: Improve 'Mono' compatibility with Windows

Finii:
#### Description

**[why]**
Some of the patched Mono fonts do not turn up in the font chooser
of Windows `CMD` and `PowerShell` (and probably more).

**[how]**
For some reasons Windows does not identify the fonts as being strictly
monospaced, so they are hidden in that font choosers.

For the monospaced fonts we set now the Panose proportion 'monospaced'.
Windows seems to honor the Panose properties.

It is not clear why we need to set the old Panose props, especially as
Cascadia Code does not (!) set them and is still detected as monospaced.

Anyhow, the way Windows detects if a font is monospaced is a mystery (at
least for me), and this works, so ;-)

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [x] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [x] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### What does this Pull Request (PR) do?

Only for `font-patcher --mono`:

Set/overwrite `OS/2` - `Panose` - `Propotion` to `Monospaced`
_WHEN_ `OS/2` - `Panose` - `Family Kind` is `Any` or `Latin-Text`

Set `OS/2` - `Panose` - `Family Kind` to `Latin-Text`
_WHEN_ it was `Any` before

#### How should this be manually tested?
Try to select font in MS-Windows `cmd` application font chooser.

#### Any background context you can provide?
Cheat sheet for Panose: https://forum.high-logic.com/postedfiles/Panose.pdf

#### What are the relevant tickets (if any)?
For example #586

PR #732, opened 2021-12-22
Bugfix monospaced glyph scaling

Finii:
#### Description

Even when patched for Monospace (with `-mono`) fonts have a hard time to be recognized as monospaced font sometimes.

#### Requirements / Checklist

The first commit just adds a warning mechanism, that shows us when an inserted glyph is not scaled in the expected way. Where expected means: Fits into the normal font width. We do not patch doublewidth glyphs in, jet (afaik).

Then the scaling mechanism (`ScaleGlyph`) is expanded to handle more complex situations and do so automatically.

This new mechanism is than used for the _Font Awesome_ glyphs.

With just the first commit you can see the extend of the problem. Only Font Awesome is affected:

image

* Patch a font that is not recognized as monospaced by some applications (like Gnome Terminal, Windows CMD, ...) and see if it is afterwards recognized.
* Check glyphs from `FONTA_SCALE_LIST` in `fontforge` for example, if they scale to a reasonable size. They will be smaller then before, which is the goal of this, but people might notice.
* Check font metadata, for example with
    * `ttfdump FontName.ttf  | grep 'advanceWidthMax|xAvgCharWidth'`
    * `showttf FontName.ttf | grep 'advanceWidthMax|avgWidth'`

Keep in mind that this is just relevant for 'Nerd Font Mono' fonts, i.e. patched with `-mono`. Without the parameter the behavior of `font-patcher` is not changed.

Bottom of #695, for example https://github.com/ryanoasis/nerd-fonts/pull/695#issuecomment-997337500

Instead of taking ONE 'main glyph' that rules the scaling of all the special glyphs, we form groups of glyphs that have to be scaled together. Like all the selected (blue) glyphs here, but as individual groups:

image

The combined bounding box of the group is determined (like when all glyphs in the group are stamped on top of each other and then the bounding box is found), and all members of the group are scaled such that the combined bounding box is maximized to the allowed limits.

For example the group of four arrows (in line 5) is scaled together, but independently to all other glyphs.

Here an example that shows how Font Awesome is ... ~buggy~ strange, the selected (blue) glyphs are clearly too big for the font itself, that does maintain good glyph size consistency otherwise.

image

For example #703 (part of)


Finii:
I tried to track down when this first occured. But even when `ScaleGlyph` has been only that, one glyph (and not a dict that contains a lot of information, and ALL glyphs have been scaled by the ScaleGlyph, already then the very problematic glyphs (mars, venus, merkur, etc) were there and were too big.

So obviously this has never been checked. This is one reason why the runtime check-and-warning is introduced.

Furthermore this PR is rather big (in code change), unfortunately. There is only one symbol font, and only 24 (?) glyphs that are affected, so the question is legitimate: Do we want such a big change.

Alternatively one could add Font Awesome piecewise, with different 'ordinary' ScaleGlyphs. But all in all that is not easier to comprehend, and so I ditched that way (and the commits). This is a hard decision, because for me it is a question of maintainability and understand-ability of the code, and it is not clear what is better (for 24 glyphs :sob:). In the end my choice fell on the more complex data structure but more straight forward definition and use. But of course we can also have a look on the other design choice and or develop a completely different approach (that I could not come up with).

_Ugg, I just see that the CI does not run through... Python version stuff I presume, will fix_

Finii:
This exposes a scaling glitch that is in font-patcher:

image

These are glyphs that are not in the `ScaleGlyph` set. I guess the reason is some rounding issue, still investigating.

Finii:
Scaling glitch fixed with a7f91ac

The errors were always there, just invisible because we did not check :-D

This PR is now ready :-)

Finii:
```shell
$ fontforge --script ~/git/nerd-fonts/font-patcher --mono --complete ~/Download/Overpass_Mono/static/OverpassMono-Regular.ttf
```
Left side: **before** this PR, right side **after** this PR.

#### Gender / planet glyphs rescaled smaller (and independent from other glyphs):

image

#### Rounding error fix of non-`ScaleGlyph`-Glyph:

image

#### Metadata diagnostics:

```shell
$ ttfdump Overpass\ Mono\ Regular\ Nerd\ Font\ Complete\ Mono_prior.ttf | grep 'advanceWidthMax|xAvgCharWidth' 
	 advanceWidthMax:		 1517
	 xAvgCharWidth:		 1232
$ ttfdump Overpass\ Mono\ Regular\ Nerd\ Font\ Complete\ Mono_after.ttf | grep 'advanceWidthMax|xAvgCharWidth'
	 advanceWidthMax:		 1257
	 xAvgCharWidth:		 1232
```

_Note: 1232 * 1.02 = 1256.6 (for overlap 2% on powerline stuff)_

font-patcher
```diff
@@ -817,6 +842,20 @@ class font_patcher:
...
+                    print("  Warning: Scaled glyph U+{:X} wider than one monospace width ({} / {} (overlap {}))".format(
+                        currentSourceFontGlyph, int(xmax - xmin), self.font_dim['width'], overlap))
+
+            # reset selection so iteration works properly @TODO fix? rookie misunderstanding?
+            # This is likely needed because the selection was changed when the glyph was copy/pasted
+            if symbolFontStart == 0:
+                symbolFont.selection.all()
+            else:
+                symbolFont.selection.select((str("ranges"), str("unicode")), symbolFontStart, symbolFontEnd)
+
```
ryanoasis, line 858: This can be removed how with #736 ?

font-patcher
```diff
@@ -817,6 +842,20 @@ class font_patcher:
             # does not overlap the bearings (edges)
             self.remove_glyph_neg_bearings(self.sourceFont[currentSourceFontGlyph])
 
+            # Check if the inserted glyph is scaled correctly for monospace
+            if self.args.single:
+                (xmin, _, xmax, _) = self.sourceFont[currentSourceFontGlyph].boundingBox()
+                if int(xmax - xmin) > self.font_dim['width'] * (1 + overlap):
+                    print("  Warning: Scaled glyph U+{:X} wider than one monospace width ({} / {} (overlap {}))".format(
```
ryanoasis, line 849: I think this should print on a new line. when progressbars are on it displays on the same line

font-patcher
```diff
@@ -770,10 +793,12 @@ class font_patcher:
                 # Currently stretching vertically for both monospace and double-width
                 scale_ratio_y = self.font_dim['height'] / sym_dim['height']
 
+            overlap = sym_attr['params']['overlap'] if 'overlap' in sym_attr['params'] else 0
```
ryanoasis, line 796: I don't personally like this conditional style but is it common and does PEP8 or other have particular rules regarding?

font-patcher
```diff
@@ -864,6 +903,53 @@ class font_patcher:
         except:
             pass
 
+    def prepareScaleGlyph(self, scaleGlyph, symbolFont):
+        """ Prepare raw ScaleGlyph data for use """
```
ryanoasis, line 907: 👍🏻

font-patcher
```diff
@@ -893,18 +987,9 @@ def get_glyph_dimensions(glyph):
...
-    for i in glyph_list:
-        if isinstance(i, tuple):
-            if unicode_value >= i[0] and unicode_value <= i[1]:
-                return True
-        else:
-            if unicode_value == i:
-                return True
-    return False
-
+def get_glyph_dimensions(glyph):
```
ryanoasis, line 990: nice

font-patcher
```diff
@@ -547,9 +547,38 @@ class font_patcher:
         # Most glyphs we want to maximize during the scale.  However, there are some
         # that need to be small or stay relative in size to each other.
         # The following list are those glyphs.  A tuple represents a range.
-        DEVI_SCALE_LIST  = {'ScaleGlyph': 0xE60E, 'GlyphsToScale': [(0xe6bd, 0xe6c3)]}
-        FONTA_SCALE_LIST = {'ScaleGlyph': 0xF17A, 'GlyphsToScale': [0xf005, 0xf006, (0xf026, 0xf028), 0xf02b, 0xf02c, (0xf031, 0xf035), (0xf044, 0xf054), (0xf060, 0xf063), 0xf077, 0xf078, 0xf07d, 0xf07e, 0xf089, (0xf0d7, 0xf0da), (0xf0dc, 0xf0de), (0xf100, 0xf107), 0xf141, 0xf142, (0xf153, 0xf15a), (0xf175, 0xf178), 0xf182, 0xf183, (0xf221, 0xf22d), (0xf255, 0xf25b)]}
-        OCTI_SCALE_LIST  = {'ScaleGlyph': 0xF02E, 'GlyphsToScale': [(0xf03d, 0xf040), 0xf044, (0xf051, 0xf053), 0xf05a, 0xf05b, 0xf071, 0xf078, (0xf09f, 0xf0aa), 0xf0ca]}
+        DEVI_SCALE_LIST  = {'ScaleGlyph': 0xE60E, # Android logo
```
ryanoasis, line 550: 👍🏻

font-patcher
Finii: I honestly do not know. And I do not really care, if you like traditional `if/else` more I can change it whereever it is now.

font-patcher
Finii: This is a merge error. It has not been in any of my commits.
I usually do a rebase on master instead of merging in master; find that easier to follow and later on it is not so much work to pull it (fast forward will be possible).

I will drop your merge commit and do a proper rebase, it should be vanished then.

Finii:
Force push to remove merge-commit and do instead a rebase on master.

Finii:
Ah, git tricked me also... Did the rebase again with `git am` and now we have it.
Force push.

Also changed the stuff from the review above.

font-patcher
Finii: Changed.

font-patcher
Finii: Changed.

This comes from my usual style that a work as much with `const` stuff as possible, and then you need to assign the thing you want and can not use an `if` structure like this.
Python is no language I use often.

font-patcher
Finii: Code Climate would have complained anyhow ;-) because of code repetition.

(Well, I guess it is `off` for `font-patcher`, but it would have complained.)

font-patcher
Finii: Hmm, and it can not be avoided for more complex Python constructs (i.e. list comprehension)
```python
                sym_dim = get_multiglyph_boundingBox([ symbolFont[g] if g in symbolFont else None for g in group ])
```
That is very hard to write differently and still in a Python way... So I kept it there. Looks super clumsy if done manually.

font-patcher
Finii: :-)

font-patcher
ryanoasis: no worries. thanks!

ryanoasis:
I meant to do a final review/test this weekend... maybe this week 🤞🏻

ryanoasis (approved):
code change and results in fontforge look good. The font I tested (3270) still didn't show up in my Virtualbox Windows 10 cmd _but_ the results in Konsole were promising. I think is good enough to merge. Thanks!

PR #736, opened 2021-12-22
font-patcher: Copy selection instead of continuously regenerating

Finii:
**[why]**
This is a TODO item. Well, two in fact.

The `symbolFont`'s selection is used for two things
- the main loop to iterate over all glyphs to insert
- to select the one glyph that is actually copied over

Because the main loop uses iterators on the selection.
The iterator is not 'stable' but invalidates if the selection
is changed.
The current code therefor restores the old selection before the loop
jumps to the head again.

This design is not very robust.

[how]
We need the selection to copy the symbol glyph.

But we can rewrite the loop that it does not need the selection at every
iteration, but that the selection is copied into a list, and we loop
over that list - which is independent on a later selection state.

#### Requirements / Checklist

Remove two `TODO` items.

Patch one font with and one without this PR. Compare if symbols that we patch in are there and have same size as before.

_Edit: Fill how-to-test paragraph_

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -647,10 +647,8 @@ class font_patcher:
             careful = True
 
         if exactEncoding is False:
-            sourceFontList = []
+            sourceFontList = list(range(sourceFontStart, sourceFontEnd + 1))
             sourceFontCounter = 0
-            for i in range(sourceFontStart, sourceFontEnd + 1):
-                sourceFontList.append(format(i, 'X'))
 
         scale_factor = 0
         if scaleGlyph:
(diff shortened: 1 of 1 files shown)


Finii:
Add two unrelated code simplifications. Cherry pick whatever you like ;-)
Look at the individual commits not the total diff.

These are just items I needed to overcome in my mind when started working on `font-forge`.

Finii:
Another oddity...

image

We check if the hex-string in `copiedToSlot` can be
- converted to a positive int
- starts with 'uni'

Both will never happen, because we fill it with a generated string `format(..., 'X')` in all cases.

font-patcher
```diff
@@ -685,46 +683,34 @@ class font_patcher:
...
-            if int(copiedToSlot, 16) < 0:
-                print("Found invalid glyph slot number. Skipping.")
-                continue
-
             if self.args.quiet is False:
                 if self.args.progressbars:
                     update_progress(round(float(index + 1) / glyphSetLength, 2))
                 else:
-                    progressText = "\nUpdating glyph: " + str(sym_glyph) + " " + str(sym_glyph.glyphname) + " putting at: " + copiedToSlot
+                    progressText = "\nUpdating glyph: {} {} putting at: {:X}".format(sym_glyph, sym_glyph.glyphname, currentSourceFontGlyph)
```
Finii, line 695: I believe we should drop `sym_glyph.glyphname` from the output, as that is already in the to-string-cast `sym_glyph`.
But well, I did not want to change the outside-visible behavior.

Finii:
Maybe the code could also benefit from a style unification. CamelCase versus snail_case? "string" versus 'string'? print() versus sys.write()? etc.
With FontnameParser I tried to follow PEP.

ryanoasis (approved):
Great, thank you!

Actually all of the changes look good to me. really cleans/simplifies things too.

font-patcher
```diff
@@ -688,46 +683,34 @@ class font_patcher:
...
-                # convince that this string really is a hex:
-                currentSourceFontGlyph = int("0x" + sourceFontList[sourceFontCounter], 16)
-                copiedToSlot = sourceFontList[sourceFontCounter]
+                currentSourceFontGlyph = sourceFontList[sourceFontCounter]
                 sourceFontCounter += 1
 
-            if int(copiedToSlot, 16) < 0:
-                print("Found invalid glyph slot number. Skipping.")
-                continue
-
```
ryanoasis, line 704: `custom` still works which is when this was added. probably safe and unneeded so this seems okay to remove 👍🏻

Issue #662, opened 2021-09-09
Unable to patch fonts on Apple silicon


selfagency:
Hello,

I'm trying to patch a font but despite having FontForge installed locally, which is said to be the only prerequisite in the docs, attempting to run the Python script resulted in:

```
» /opt/font-patcher.py         

Nerd Fonts: FontForge module is probably not installed. [See: http://designwithfontforge.com/en-US/Installing_Fontforge.html]
```

So I tried it this way:

```
» fontforge -script /opt/font-patcher.py RecMonoSemicasual-Bold-1.078.ttf 

Copyright (c) 2000-2021. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20201107
 Based on sources from 2021-03-14 03:27 UTC-D.
The requested file, original-source.otf, does not exist
Traceback (most recent call last):
  File "/opt/font-patcher.py", line 929, in <module>
    main()
  File "/opt/font-patcher.py", line 924, in main
    patcher.patch()
  File "/opt/font-patcher.py", line 88, in patch
    symfont = fontforge.open(self.args.glyphdir + patch['Filename'])
OSError: Open failed
```

So then I tried it this way:

```
» docker run -v unpatched:/in -v patched:/out nerdfonts/patcher -c                          

WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
```

And finally, I tried it this way: 
```
» docker run --platform linux/amd64 --rm -v unpatched:/in -v patched:/out nerdfonts/patcher -c 
```

...which simply fails to launch, whereas Docker notes "Attempts to run Intel-based containers on Apple Silicon machines under emulation can crash as qemu sometimes fails to run the container." Docker's recommended remedy is "encouraging container authors to produce arm64, or multi-arch, versions of their containers." So I'm here to encourage you to please [build a multi-arch](https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/) version of the Docker image so that I may patch my fonts in peace.

Thank you!

Finii:
Added hint to download glyphs: done
Docker should work on Silicon now, I believe :thinking: there was some PR, iirc.
And/or use `FontPatcher.zip` from the release page.

sammcj:
[...]

Issue #659, opened 2021-08-27
font-patcher crashes when patching Agave


ThoriumTextile:
#### 🗹 Requirements

- [X] A brief but descriptive _title_ of your issue
- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [X] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [X] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

#### 🎯 Subject of the issue
Whenever i try to patch (no output dir specified) `Agave-Regular-slashed` (from [here](https://github.com/blobject/agave)), font-patcher crashes with the following details/log:
**This happens with the regular Agave version (not slashed) too.**
```
The following table(s) in the font have been ignored by FontForge
  Ignoring 'TTFA'
The requested file, original-source.otf, does not exist
Traceback (most recent call last):
  File "/home/gabri/Downloads/./font-patcher", line 929, in <module>
    main()
  File "/home/gabri/Downloads/./font-patcher", line 924, in main
    patcher.patch()
  File "/home/gabri/Downloads/./font-patcher", line 88, in patch
    symfont = fontforge.open(self.args.glyphdir + patch['Filename'])
OSError: Open failed
```

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
- _Agave_
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
- _kitty_
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
- _Fedora 34, Linux_

ryanoasis:
were you using the font-patcher from this repo? If so how did you download it? It sounds like you are missing the source files that the patcher relies upon.

NuraliMedeu:
Hello! I have the same issue when patching [Kawkab Mono](https://github.com/aiaf/kawkab-mono).

## 🗹 Requirements
- [x] A brief but descriptive title of your issue
- [x] I have searched the issues for my issue and found nothing related and/or helpful
- [x] I have read or scanned the FAQ
- [x] I have read or scanned the Wiki

## 🎯 Subject of the issue
The same as above.

## 🔧 Your Setup
[...]

Issue #495, opened 2020-06-24
font patcher fails


doums:
### 🎯 Subject of the issue

I installed FontForge from my distribution's package manager (`pacman` on Arch Linux).

I tried to patch several ttf fonts: JetbrainsMono (downloaded [here](https://www.jetbrains.com/lp/mono/)) and Inconsolata (downloaded from google fonts), in various style. All fail with the same error message:

```
$ fontforge -script font-patcher --octicons static/Inconsolata-Bold.t
tf
Copyright (c) 2000-2020. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20200314
 Based on sources from 2020-03-18 19:16 UTC-ML-D-GDK3.
The following tables are ignored by FontForge
  'STAT' style attributes table
Use-my-metrics flag set on at least two components in glyph 631
File original-source.otf does not exist
Traceback (most recent call last):
  File "font-patcher", line 924, in <module>
    main()
  File "font-patcher", line 919, in main
    patcher.patch()
  File "font-patcher", line 88, in patch
    symfont = fontforge.open(__dir__ + "/src/glyphs/" + patch['Filename'])
OSError: Open failed
```
I don't know what is the problem ?

#### 🔧 Your Setup

Arch Linux

cfbender:
It doesn't make it very clear, but you cannot just run the script on it's own - you need to first clone the repo with `--depth 1` and run the script from there. You may be able to get away with only downloading the files you need from `src`, but cloning will definitely fix this for you.

Etherian:
I ran into this issue as well.

@ryanoasis, could what tools and resources are needed and where to find them be clarified in the Font Patcher section? For example, it wasn't immediately clear that the `font-patcher` script was in the repo, and I didn't know from where the script was expecting to pull the glyphs until I decyphered the Python code in the error message.

Etherian:
I ran into this issue as well.

@ryanoasis, could what tools and resources are needed and where to find them be clarified in the Font Patcher section? For example, it wasn't immediately clear that the `font-patcher` script was in the repo, and I didn't know from where the script was expecting to pull the glyphs until I decyphered the Python code in the error message.

wimstefan:
[...]

PR #741, opened 2021-12-27
font-patcher: Check if glyph source is available (`--glyphdownload`)

Finii:
**[why]**
When users just download the script (and not the source glyphs) the
script fails with an obscure error message.

**[how]**
Check if the glyphdir exists at all. If not give a hint to download the
glyphs.

Check if the individual glyph font exists and is readable. Bail out if
not.

Add option `--glyphdownload` that tries to download missing glyph fonts.

#### Requirements / Checklist

Add checks before we open a symbol font and give hints to the users if we can not.

* Patch a font _(shall work as before)_
* Remove one (needed) symbol font, patch a font again _(errors out)_
* Remove complete `src/glyphs/` (or use `--glyphdir`), patch a font again _(errors out)_

The two bottom cases a **meaningful** message should be printed.

* Repeat the 3 tests but specify `--glyphdownload`. _(all shall work)_

For example #659 #662 #495

image

_Edit: Add `--glyphdownload` as that has been added as 2nd commit to the PR_

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.1.1"
+script_version = "3.2.0"
 
 version = "2.2.2"
 projectName = "Nerd Fonts"
@@ -35,6 +35,12 @@ except ImportError:
             "[e.g. on Linux Debian or Ubuntu: `sudo apt install fontforge python3-fontforge`]"
         )
     )
+
+try:
+    import urllib.request
+    urllib_present = True
+except ImportError:
+    urllib_present = False
 
 # This is for experimenting
 sys.path.insert(0, os.path.abspath(os.path.dirname(sys.argv[0])) + '/bin/scripts/name_parser/')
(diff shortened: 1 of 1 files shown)


Finii:
Not sure if we want to go that route, but it is requested a lot it seems...:

Autodownload...

image

Finii:
Two notes:

* Of course the first commit should be merged regardless
* The second commit is kind of optional

If we go the route with the second commit there might be problems of `font-patcher` version versus actual repo version - when the glyph paths change. So in reality we should download the repo-glyph stuff from the version (and not MASTER) that the font-patcher actually is. But this is not possible at the moment because the versions are not increased even when the directory structure is changed. This might need more thinking and/or ideas. At least maybe a warning in the code when a download fails, that one needs to use the font-patcher HEAD.

Finii:
Rebase on master, force push

Finii:
All readme files need to be corrected :stop_sign:

Finii:
Does this download the `--makegroups` stuff also? I guess not. Would that be beneficial, or just disable the feature? Who uses that feature?

Finii:
I believe we should promote the `FontPatcher.zip` instead.
This would need to keep a parallel set of sub-files as for the zip generation.
A lot of maintenance burden.

Issue #694, opened 2021-11-29
Some Iosevka variants broken


Finii:
#### 🗹 Requirements

- [x] A brief but descriptive _title_ of your issue
- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

## The patched Iosevka Mono fonts are more or less broken

Some Iosevka Mono fonts can not be opened at all (or it takes more than some minutes), while most other open with error messages of this kind:

```
patched-fonts/Iosevka/Thin-Italic/complete/Iosevka Thin Italic Nerd Font Complete Mono.ttf
PythonUI_Init()
copyUIMethodsToBaseTable()
Internal Error:
Subtable status not filled in for 0th subtable of 'calt' Contextual Alternates lookup 26
Subtable status not filled in for 1th subtable of 'calt' Contextual Alternates lookup 26
Subtable status not filled in for 2th subtable of 'calt' Contextual Alternates lookup 26
Subtable status not filled in for 0th subtable of 'dlig' Discretionary Ligatures lookup 32
Subtable status not filled in for 1th subtable of 'dlig' Discretionary Ligatures lookup 32
Subtable status not filled in for 2th subtable of 'dlig' Discretionary Ligatures lookup 32
Subtable status not filled in for 0th subtable of Reverse Contextual Chaining Substitution lookup 35
Subtable status not filled in for 1th subtable of Reverse Contextual Chaining Substitution lookup 35
Subtable status not filled in for 2th subtable of Reverse Contextual Chaining Substitution lookup 35
Subtable status not filled in for 0th subtable of Reverse Contextual Chaining Substitution lookup 38
Subtable status not filled in for 1th subtable of Reverse Contex
```

I believe fontforge tries to fix the fonts that do not open for extended times like this:
```
...
Attempt to reference lookup 11293 (within a contextual lookup), but there are only 429 lookups in 'GSUB'
Attempt to reference lookup 11805 (within a contextual lookup), but there are only 429 lookups in 'GSUB'
Attempt to reference lookup 12317 (within a contextual lookup), but there are only 429 lookups in 'GSUB'
[...]

Issue #742, opened 2021-12-29
Patched Iosevka font will remove ligature inside it


GulajavaMinistudio:
There is compatibility issue in Nerd Font Patcher for Iosevka 11 or up version. I try patching Iosevka 11.x with font patcher, but causing ligature removed from it. I want to keep ligature inside Iosevka 11, but the font patcher remove it.

And in Terminal progress, there's some error warning , i think this is causing ligature can't work in Iosevka 11 ?

**"Internal Error: Attempt to output 66740 into a 16-bit field. It will be truncated and the file may not be useful."**

![Screen Shot 2021-12-29 at 07 34 53](https://user-images.githubusercontent.com/11639412/147616965-2595a00d-0ac4-4b06-86b6-6da99b0dd62b.png)

And here is my command to running font patcher
`fontforge -script ./font-patcher ./iosevka-bolditalic.ttf --complete`

Any ide to bring back ligature in patched Iosevka 11's font ?

Finii:
Which ligature is it you miss?

GulajavaMinistudio:
> Which ligature is it you miss?

After patching with font patcher, **"calt"** and **"dlig"** is removed from Iosevka. I want to keep it like Iosevka v10 and older. But in Iosevka 11, ligation can't be shown because it's removed.

xsrvmy:
Running into this same issue (I'm getting a different number though)
https://github.com/fontforge/fontforge/issues/3009
I'm assuming the table responsible for generating ligatures is getting cut off. Iosevka has a LOT of features so maybe that's why.

GulajavaMinistudio:
I'm still use font patcher in master branch, and try to use other Iosevka variant called "Iosevka Fixed" that contain no built in ligature and no built in glyph. 
And this run successfully without "truncated error" like previous screenshot with Iosevka and Iosevka Term.

Here is newest screenshot with "Iosevka Fixed" variant
![Screen Shot 2022-01-07 at 22 46 43](https://user-images.githubusercontent.com/11639412/148568860-443809ae-2a55-4775-bf6e-38c1b02e424f.png)

And here package list link for description about Iosevka Fixed. [https://github.com/be5invis/Iosevka/blob/v11.2.4/doc/PACKAGE-LIST.md](https://github.com/be5invis/Iosevka/blob/v11.2.4/doc/PACKAGE-LIST.md)

I think font patcher will get error if there's some built in ligation and glyph, especially in Iosevka or other font like Fira Code.

[...]

PR #717, opened 2021-12-15
Introduce a font name parser

Finii:
**[why]**
A lot of the fonts have incorrect naming after patching. A completely
different approach can help to come up with a consistent naming scheme.

**[how]**
See [README](https://github.com/ryanoasis/nerd-fonts/blob/feature/rewrite-setup_font_names/bin/scripts/name_parser/README.md)

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [x] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [x] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### Any background context you can provide?

In general (MS or Apple, is the same):
https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-ids
https://www.fonttutorials.com/how-to-name-font-family/

About the length limits:
https://adobe-type-tools.github.io/font-tech-notes/pdfs/5088.FontNames.pdf
https://forum.glyphsapp.com/t/overly-strict-font-name-max-length-recommendation-in-naming-tutorial/10164
https://github.com/googlefonts/fontbakery/issues/1488
https://typedrawers.com/discussion/617/family-name/p2

#### Screenshots (if appropriate or helpful)

_Edit: Add all links in Background paragraph_

PR #750, opened 2022-01-03
font-patcher: Remove CV lookups in Iosevka

Finii:
**[why]**
For some reason `fontforge` is unable to pack all the lookup tables of
Iosevka back into the generated font, although they fit obviously in the
source font.

The lookup tables will become garbled and `fontforge` says something like:

        Attempt to output 67666 into a 16-bit field.
        It will be truncated and the file may not be useful.

As this also happens if we do nothing, just open and generate the font
without touching, there is not much we can do.

**[how]**
The only thing we can do is reduce the number of lookup tables that are
in the font file. This is somewhat brute, but at least we end up with a
usable font, albeit with a little bit less features.

Unfortunately the SS (Style Set) tables are not enough, but the CV
(Character Variants) suffice. So all CV lookups are riped out.

Fixes: #694 

#### Requirements / Checklist

Remove the CV lookup tables from all Iosevka variants.
Other fonts in the `src/unpatched-fonts/` are not affected/changed.

Fontforge ~~bug~~ unexpected behavior. Even in interactive mode we can not open and directly (unmodified) the Iosevka fonts in opentype format. See #694

#694 and possibly #742

Thanks @xsrvmy for the hint in https://github.com/ryanoasis/nerd-fonts/issues/742#issuecomment-1003851925

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -60,6 +60,7 @@ class font_patcher:
         except Exception:
             sys.exit(projectName + ": Can not open font, try to open with fontforge interactively to get more information")
         self.setup_font_names()
+        self.remove_too_many_lookups()
         self.remove_ligatures()
         make_sure_path_exists(self.args.outputdir)
         self.check_position_conflicts()
@@ -433,6 +434,26 @@ class font_patcher:
         self.sourceFont.sfntRevision = None # Auto-set (refreshed) by fontforge
         self.sourceFont.appendSFNTName(str('English (US)'), str('Version'), "Version " + self.sourceFont.version)
         # print("Version now is {}".format(sourceFont.version))
+
+
+    def remove_too_many_lookups(self):
+        # Iosevka has so many lookup tables that fontforge is unable to save the file
+        lookups = self.sourceFont.gsub_lookups
+        # We identify Iosevka not by name, but by the amount of lookups.
+        # Maybe we want to identify by name?
+        # The number of lookups is not relevant if the save can be done or not,
+        # it depends on the space the tables need.
+        if len(lookups) < 410:
+            return
+        removed = []
+        for l in lookups:
+            # Arbritarily choose to remove CV. SS is too small to help much.
+            hit = re.search('cv[0-9]+', l)
+            if hit:
+                self.sourceFont.removeLookup(l)
+                removed.append(hit.group())
+        if removed:
+            print("Removing lookup: ", ','.join(removed))
 
 
     def remove_ligatures(self):


Finii:
It can be discussed if that is the right soluion, or if CV is used by anyone (urgently needed).

Maybe we could remove other tables instead. But it's hard to decide on _which_ and make the decision future proof.

Finii:
Used _number of lookup tables_ to identify Iosevka...

For comparison here the number of lookup of all fonts in `src/unpatched-fonts` that have more than 41:
```
 42 src/unpatched-fonts/Noto/Serif/
113 src/unpatched-fonts/Hasklig/
115 src/unpatched-fonts/Monoid/ 
145 src/unpatched-fonts/Lilex/  
170 src/unpatched-fonts/VictorMono/                          
198 src/unpatched-fonts/FantasqueSansMono/                   
221 src/unpatched-fonts/VictorMono/                          
385 src/unpatched-fonts/CascadiaCode/                                   
403 src/unpatched-fonts/FiraCode/                            
403 src/unpatched-fonts/JetBrainsMono/Ligatures/             
429 src/unpatched-fonts/Iosevka/
```

So that is obviously not a good metrics for total lookup size. But ok-ish for our purpose?

xsrvmy:
Iosevka can be manually compiled with cv's selected. So a manual patch is always possible.
On the topic of Iosevka, can you make the script set the WWS family and subfamily? Otherwise the system group the font under Iosekva instead of Iosevka NF

Finii:
> can you make the script set the WWS family and subfamily?

#717 ... it\s ready albeit the not-ready state ;)

No wait, WWS you say? Normally 717 should be enough, no?

xsrvmy:
WWS family has to have Nerd Font appended to it, or Iosevka nerd font will be treated as a variant of Iosevka rather than a seperate font.

Finii:
But you did have a look at 717?

It sets ID 1 & 2 (Family and SubFamily) and 16 & 17 (Typographic Family and Typographic Subfamily/Style).

WWS Family ... normally is not set:

> WWS Family Name. Used to provide a WWS-conformant family name in case the entries for IDs 16 and 17 do not conform to the WWS model.

_[Source](https://docs.microsoft.com/en-us/typography/opentype/spec/name#name-ids)_

And 717 is specifically designed to make 16/17 (and 1/2) to obey the customary rules)

That is why I ask. Did you try 717 and still need WWS? What constellation?

xsrvmy:
On my systems it seems that WWS is overriding typographic family

xsrvmy:
> Used _number of lookup tables_ to identify Iosevka...
> 
> For comparison here the number of lookup of all fonts in `src/unpatched-fonts` that have more than 41:
> 
> ```
>  42 src/unpatched-fonts/Noto/Serif/
> 113 src/unpatched-fonts/Hasklig/
> 115 src/unpatched-fonts/Monoid/ 
> 145 src/unpatched-fonts/Lilex/  
> 170 src/unpatched-fonts/VictorMono/                          
> 198 src/unpatched-fonts/FantasqueSansMono/                   
> 221 src/unpatched-fonts/VictorMono/                          
> 385 src/unpatched-fonts/CascadiaCode/                                   
> 403 src/unpatched-fonts/FiraCode/                            
> 403 src/unpatched-fonts/JetBrainsMono/Ligatures/             
> 429 src/unpatched-fonts/Iosevka/
> ```
> 
> So that is obviously not a good metrics for total lookup size. But ok-ish for our purpose?

Does this work for the ss0x ttfs?

Finii:
I do not understand the question?

Stylistic Sets are untouched.

Only Iosevka has that many tables.

Hmm.

??

xsrvmy:
I was talking about Iosevka's ttfs that have ss pre-applied. Because their file size is much smaller I wasn't sure if detecting via look up table size would work. I tried this branch on them and it still removes the cv's though, so there is no concern.

Finii:
Oh. In `src/unpatched-fonts/Iosevka/` there are no `otf` fonts, only `ttf`s.

It detects the number of lookup tables, not size (which is much more involved to determine).

:-)

Finii:
It's not only `fontforge` that has a hard time to handle Iosevka.
Also `ttfdump` crashes (`showttf` works).

image

It's of course not Iosevka's fault, but ... it is no easy font either.

GulajavaMinistudio:
Screen Shot 2022-01-05 at 23 50 16

Trying the latest commit from repository. It still error to patching latest Iosevka version.

Finii:
Seems you are on `master` but the fix has not yet been pulled. You need to switch the branch or cherry-pick the fix-commit?

Or am I wrong?

GulajavaMinistudio:
> Seems you are on `master` but the fix has not yet been pulled. You need to switch the branch or cherry-pick the fix-commit?
> 
> Or am I wrong?

Oh my bad, i forget that the fix is still in pull request. I will try with switching branch soon.

GulajavaMinistudio:
> Seems you are on `master` but the fix has not yet been pulled. You need to switch the branch or cherry-pick the fix-commit?
> 
> Or am I wrong?

I'm still use font patcher in master branch, and try to use other Iosevka variant called "Iosevka Fixed" that contain no built in ligature and no built in glyph. 
And this run successfully without "truncated error" like previous screenshot with Iosevka and Iosevka Term.

Here is newest screenshot with "Iosevka Fixed" variant
Screen Shot 2022-01-07 at 22 46 43

And here package list link for description about Iosevka Fixed. [https://github.com/be5invis/Iosevka/blob/v11.2.4/doc/PACKAGE-LIST.md](https://github.com/be5invis/Iosevka/blob/v11.2.4/doc/PACKAGE-LIST.md)

I think font patcher will get error if there's some built in ligation and glyph, especially in Iosevka

xsrvmy:
No idea if the ligation is the real issue. There are some ligatures that have CVs as well.

GulajavaMinistudio:
Trying font patcher from branch "bugfix/Iosevka-lookups" , still not successful to patch Iosevka Font with custom build. 
Screen Shot 2022-01-08 at 07 22 41

Still showing 
```sh
"Done with Patch Sets, generating font...
Internal Error: Attempt to output 65574 into a 16-bit field. It will be truncated and the file may not be useful."
```

Here i attach Iosevka custom build in this comment, maybe you can try to patch and debugging font patcher with it. 
[ttf-iosevka-mayukai-custombuild.zip](https://github.com/ryanoasis/nerd-fonts/files/7831986/ttf-iosevka-mayukai-custombuild.zip)

Finii:
> I think font patcher will get error if there's some built in ligation and glyph, especially in Iosevka

Ligatures are no problem, and I persoanlly like the 'programming' ligatures - and use Nerd Fonts with ligatures ever since.

The problem is that Iosevka has **a lot** lookup tables, and when we (i.e. `fontforge`) re-packs them runs out of space.
In my opinion ligatures are supported by almost all fontends people can use, and should not be removed without need. On the other hand most programs struggle to support advanced OpenType features like Stylistic Sets (SS) and Alternative Glyps - thats why I thought IF we remove something, let it be those.

Finii:
> Here i attach Iosevka custom build in this comment, maybe you can try to patch and debugging font patcher with it.

Hmm, patching is no problem with Regular. And that files has only 157 tables, so it would not need the `bugfix/Iosevka-lookups` anyhow (i.e. the branch does not handle that font any different then before).

Dropping of tables would be shown where the red arrow is:

image

and comes out without error:

image

I tried with `fontforge`:
```
 Version: 20201107
 Based on sources from 2021-01-15 15:55 UTC-ML-D-GDK3.
```
and HEAD (self built).

----

Ah, you have _Italic_ in the name, so you probably did this?

image

which also has no issues here

image

Finii:
This is how it looks if tables are removed:

image

Maybe you can include the top in your screenshots, i.e. the invocation (command line) and the first few lines until patching starts.

GulajavaMinistudio:
> Maybe you can include the top in your screenshots, i.e. the invocation (command line) and the first few lines until patching starts.

Here's my additional screenshot and command that i use. 
`fontforge -script ./font-patcher /Users/gulajavministudio/Documents/WebstormProject/Iosevka-11.2.4/dist/iosevka-mayukai-sonata-terminal-ss03-v574/ttf/iosevka-mayukai-sonata-italic.ttf --careful --complete`
Screen Shot 2022-01-09 at 01 46 44

If i use --careful tag, the result is failed with internal error "truncated" like before.
Screen Shot 2022-01-09 at 01 48 16

---

But if i use without --careful tag, the result is successfully patched.
Screen Shot 2022-01-09 at 01 52 09
The result is success
Screen Shot 2022-01-09 at 01 52 19

Finii:
Can reproduce.

Finii:
At least the old and your new issue appear at the same spot when writing the subtables...:

image

Finii:
I more and more think this is a `fontforge` bug.

The GSUB table _is_ big, one can not use `short` to represent indices within them...:

```
True Type Font File Dumper: v 0.5.5
Copyright 1996-1998 ollie@ms1.hinet.net
Dumping File:iosevka-regular.ttf


Offset Table
------------
         sfnt version:           1.0
         number of tables: 18
   0. 'DSIG' - checksum = 0x00000001, offset = 0x004f3018, len =         8
   1. 'GDEF' - checksum = 0x13de42bd, offset = 0x0000012c, len =       718
   2. 'GPOS' - checksum = 0x0050d227, offset = 0x000003fc, len =    455900
   3. 'GSUB' - checksum = 0xa79d9c51, offset = 0x0006f8d8, len =    179962
   4. 'OS/2' - checksum = 0x71108694, offset = 0x0009b7d4, len =        96
   5. 'cmap' - checksum = 0x01d678f7, offset = 0x0009b834, len =      5772
   6. 'cvt ' - checksum = 0x337614d2, offset = 0x004f206c, len =       200
   7. 'fpgm' - checksum = 0x622f037f, offset = 0x004f2134, len =      3596
   8. 'gasp' - checksum = 0x00000010, offset = 0x004f2064, len =         8
   9. 'glyf' - checksum = 0xbd7c4a19, offset = 0x0009cec0, len =   3977248
  10. 'head' - checksum = 0x1ae21d3f, offset = 0x00467ee0, len =        54
  11. 'hhea' - checksum = 0x049b52f3, offset = 0x00467f18, len =        36
  12. 'hmtx' - checksum = 0x82097cfe, offset = 0x00467f3c, len =     91238
  13. 'loca' - checksum = 0x87347f18, offset = 0x0047e3a4, len =     96724
```
GPOS is even bigger, is processed before GSUB, and has no issue although it is generated by the same function!

Continuing to examine whilst doing Zoom :yawning_face:

Finii:
Found. Creating patch over there.

Finii:
`fontforge` would be fixed with https://github.com/fontforge/fontforge/pull/4883, but then the question is how long will it take to be pulled, released, distributed, ...

I'm not sure what we should do with this issue. This PR does not work reliably, maybe I can come up with an intermediate solution.

Finii:
Fontforge merged the bugfix into MASTER, so in principle we could generate the patched Iosevka with a self-build `fontforge`.

But somehow we need a solution for self patcher that use an older `fontforge`, at least a warning or something...

Finii:
Fixed via #884

PR #752, opened 2022-01-05
Handle TTCs gracefully

Finii:
**[why]**
When a True Type Collection file (`.ttc`) is used as font source this is
not handled and just the first file in the collection is processed and
saved. But the user is not informed.

When the target file format is True Type Collection, **no file at all** is
written.

These are two distinct cases, because you can in fact open a `.ttc` and
save the first font (patched) when specifying a different extension via
`-ext`. Or open a normal font and specify `ttc` as extension i.e. target
file format.

**[how]**
Check if a collection is to be opened. As we currently have no code to
loop through all fonts (and just the first font is processed) a message
is issued and we exit. Typically a user would want all the fonts and
would have to 'explode' the collection into multiple single font files
beforehand.

Prevent the target to be `ttc`, as that is not handled in fontforge at
all. To save TTCs a different API function is to be used. Unfortunately
fontforge does not care and just does nothing.

`font.generateTtc()` would have to be used with `ttc` extensions...

Anyhow. As the looping through all fonts is missing anyhow, and I feel
the usefulness is very slim, we just prevent silent failures with this
commit.

#### Requirements / Checklist

* Normal patching should not be changed.
* try `fontforge --script font-patcher [...] -ext ttc` _(shall fail with meaningful message)_
* try `fontforge --script font-patcher somefont.tcc` _(shall fail with meaningful message)_
* try `fontforge --script font-patcher somefont.tcc -ext ttf` _(WOULD work for the first embedded font, now prevented)_

https://github.com/ryanoasis/nerd-fonts/discussions/735

https://github.com/ryanoasis/nerd-fonts/issues/175
https://github.com/ryanoasis/nerd-fonts/pull/664 (!!!?)
https://github.com/powerline/fontpatcher/pull/6

We could also wield [#6](https://github.com/powerline/fontpatcher/pull/6) ... in a next step ;) If you think its needed. The change today would be much smaller than back then I guess. But there are so many PRs waiting ;)

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -55,6 +55,9 @@ class font_patcher:
             sys.exit("{}: Font file does not exist: {}".format(projectName, self.args.font))
         if not os.access(self.args.font, os.R_OK):
             sys.exit("{}: Can not open font file for reading: {}".format(projectName, self.args.font))
+        if len(fontforge.fontsInFile(self.args.font)) > 1:
+            sys.exit("{}: Font file contains {} fonts, can only handle single font files".format(projectName,
+                len(fontforge.fontsInFile(self.args.font))))
         try:
             self.sourceFont = fontforge.open(self.args.font, 1) # 1 = ("fstypepermitted",))
         except Exception:
@@ -72,6 +75,8 @@ class font_patcher:
             self.extension = os.path.splitext(self.args.font)[1]
         else:
             self.extension = '.' + self.args.extension
+        if re.match("\.ttc$", self.extension, re.IGNORECASE):
+            sys.exit(projectName + ": Can not create True Type Collections")
 
 
     def patch(self):


lilyball:
FWIW https://github.com/powerline/fontpatcher/pull/6 is mostly just whitespace. If you ask GitHub to hide whitespace in the diff it just becomes `+11 -2`. The change itself is fairly simple and can probably be ported to the current script without much difficulty.

Finii:
But is there interest? Code that is never used will break sooner or later.
I have a lot bought (commercial) fonts and some free ones, but never stumbled over `ttc`. Maybe because I prefer `otf`.
Any instances of `ttc`s?

Finii:
> If you ask GitHub to hide whitespace

Did not know that feature. Nice. (Python with its bloody indent-levels is bad for diffs :grimacing:)
Looks really simple, will give a shoot.

Finii:
It's a bit more envolved, ... need to clean up code first :unamused:
Envy your nicer/more modern naming convention :smile_cat:

See #783

PR #758, opened 2022-01-10
Update readme.md

Alyetama:
#### Description

Changing the current directory with bare `cd` might disrupt an automated installation process (which is what these commands are for). Adding parentheses to the command `(cd <path>; <command>)`  prevents switching the active directory for the shell and only switches it for the command inside the parentheses.

#### Requirements / Checklist

- Improves the README instructions.

- By running:
```
 (cd ~/.local/share/fonts && curl -fLo "Droid Sans Mono for Powerline Nerd Font Complete.otf" https://github.com/ryanoasis/nerd-fonts/raw/master/patched-fonts/DroidSansMono/complete/Droid%20Sans%20Mono%20Nerd%20Font%20Complete.otf)
```

NA

NA

NA

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -283,7 +283,7 @@ git sparse-checkout add patched-fonts/JetBrainsMono
 
 ```sh
 mkdir -p ~/.local/share/fonts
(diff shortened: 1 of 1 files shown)


Finii:
Thank you for the PR.

I wonder, is the readability better with the parens? Is the command intended to be used _as is_ in a script?

And why do you stick to `cd` instead of `curl --output-dir` (which might be a problem on systems that have a very old `curl`)?

The subshell has the problem that it is a new shell process that needs to be started ;-)

I always believed that the example there should be easily understandable and copy&paste for dummies.

_(The commit title is a bit ... nondescriptive. While the PR title can be changed that is not the case for the commit title (at least not without work.)_

Issue #612, opened 2021-04-22
Hack characters stretched when font size is <= 11 


gnojus:
#### 🎯 Subject of the issue

When font size is below or equal to **11** the patched `Hack` font is stretched and noticeably differs from original font.

Happens in terminals and native (probably gtk) font picker, but not in `libre-office`

#### 🔧 Your Setup

- **Font:** `Hack Regular Nerd Font Complete.ttf`
- **Terminal:** `xfce4-terminal` and `tilda`
- **OS:** Linux, Debian 10, xfce4

#### ★ Optional
I attach a gif displaying this behavior:
![fonts](https://user-images.githubusercontent.com/46261165/115764773-a1da3000-a3ae-11eb-8b90-fac16d2dccf9.gif)

Thanks!

Finii:
Strange.

1. I can not see the current `Hack Regular Nerd Font Complete.ttf` in the font chooser at all
2. If I prepare a patched Hack font manually, I do not see the issue:

![Peek 2022-01-07 16-09](https://user-images.githubusercontent.com/16012374/148564040-fcbc7d86-d77c-40d4-b72d-da5960968306.gif)

Where did you get the `Hack Regular Nerd Font Complete.ttf` that is recognized as monospaced by the font chooser?

Finii:
Also strange, why does the Nerd Font show up as 'Regular' 'Bold' etc in the chooser and the original Hack isn't?
I assume you installed only the `Regular` font files in both cases?

nojus297:
`Hack Regular Nerd Font Complete.ttf` I downloaded from this repo.

It is indeed strange that `Hack Nerd Font` has multiple styles - right now I see the same regular/bold/italic options, even though I have only `Hack Regular Nerd Font Complete.ttf` in `~/.fonts` and `fc-list` only lists `Regular` also.

When recording my original gif, I probably had other original Hack variants deleted to make animation easier.

Finii:
Interestingly `xfce4-terminal` shows the 4 variants (although not installed thus they would be autogenerated), but still, no issue:

![Peek 2022-01-07 16-38](https://user-images.githubusercontent.com/16012374/148568078-8b03c43c-0ed3-4843-9a89-7ed3cdd4c17a.gif)

Finii:
Let me search for a 'new' Hack Nerd Font that you could try... I believe I have some in a branch.

nojus297:
Could it be DPI related?
Mine is at 96.

[...]

PR #761, opened 2022-01-11
Feature: correct small rendering

Finii:
#### Description

**[why]**
Some fonts set specific lower resolutions and handle the small rendering
very nicely. But fontforge does not copy that information when it opens
a font; rather some default values are always written. That can destroy
the small font rendering.
    
The responsible settings are the `ppem_to_int` flag and the
`lowestRecPPEM` setting, both in the `HEAD` table.
    
**[how]**
After successfully patching and generating the new font we copy that
settings over from the source font.
    
Instead of using [`fontTools`](https://fonttools.readthedocs.io) (and thereby requiring all users to pull that
dependency) we handle the raw font file changes ourselves.
    
Fixes: 612
    
Reported-by: @nojus297

#### Requirements / Checklist

All these information is given in Issue #612 but a lot other fonts are also affected.

For example
```
fontforge font-patcher src/unpatched-fonts/CascadiaCode/Regular/CascadiaCode-Regular.otf
Copyright (c) 2000-2021. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20201107
 Based on sources from 2021-01-15 15:55 UTC-ML-D-GDK3.
PythonUI_Init()
copyUIMethodsToBaseTable()
No configfile given, skipping configfile related actions
Nerd Fonts Patcher v2.1.0 executing

Adding 56 Glyphs from Seti-UI + Custom Set 
╢████████████████████████████████████████╟ 100%
Adding 198 Glyphs from Devicons Set 
╢████████████████████████████████████████╟ 100%

Done with Patch Sets, generating font...
Changing flags from 0xB to 0x3
Changing lowestRecPPEM from 8 to 6

Generated: CaskaydiaCoveNerdFont-Regular in './Caskaydia Cove Regular Nerd Font.otf'
```
_(Note messages about 'changing' in the bottom, these values are copied over from the source font into the patched font, because `fontforge` does not even read them,)_

### DIFF
--- a/.github/workflows/font-patcher.yml
+++ b/.github/workflows/font-patcher.yml
@@ -47,6 +47,16 @@ jobs:
           sudo ninja
           sudo ninja install
 
+      - name: Compile showttf
+        run: |
+          cd "$GITHUB_WORKSPACE/fontforge-${{matrix.FontForgeRelease.version}}/contrib/fonttools"
+          mkdir build
+          cd build
+          cmake -Wno-dev -GNinja ..
+          ninja showttf
+          echo "SHOWTTF=$(realpath showttf)" >> $GITHUB_ENV
+          cd ../../..
+
       - name: Setup additional dependencies
         run: |
           pip install fonttools --quiet
@@ -82,6 +92,14 @@ jobs:
         run: |
           echo "FONT_FAMILY was ${{ env.FONT_FAMILY }}"
           [[ "${{ env.FONT_FAMILY }}" == "Hack Nerd Font" ]] && echo "Font Family matches expected" || exit 1
+
+      - name: Spot check font properties
+        run: |
+          ${{ env.SHOWTTF }} -c "$GITHUB_WORKSPACE/temp/Hack Regular Nerd Font Complete.ttf" | grep 'File Checksum.*diff=0\s*$' && echo "TTF checksum ok" || exit 1
+          ORIG_MINPPEM=$(${{ env.SHOWTTF }} -c "src/unpatched-fonts/Hack/Regular/Hack-Regular.ttf" | grep 'lowestppem=' )
+          PATCH_MINPPEM=$(${{ env.SHOWTTF }} -c "$GITHUB_WORKSPACE/temp/Hack Regular Nerd Font Complete.ttf" | grep 'lowestppem=' )
+          echo "${ORIG_MINPPEM} == ${PATCH_MINPPEM}"
+          [[ ${ORIG_MINPPEM} == ${PATCH_MINPPEM} ]] && echo "lowestRecPPEM matches" || exit 1
 
       - name: Patcher monospaced
         run: |
--- a/font-patcher
+++ b/font-patcher
@@ -93,7 +93,8 @@ class TableHEADWriter:
             self.tab_offset = self.getlong()
             self.tab_length = self.getlong()
             if tab_name == b'head':
-                break
+                return
+        raise Exception('No HEAD table found')
 
     def goto(self, where):
         """ Go to a named location in the file or to the specified index """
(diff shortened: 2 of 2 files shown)


Finii:
Instead of the hand-made (but more or less trivial) new class `TableHEADWriter` we could also utilize the excellent [`fontTools`](https://fonttools.readthedocs.io), as @nojus297 suggested [here](https://github.com/ryanoasis/nerd-fonts/issues/612#issuecomment-1009314365).

It's just the NPM explosion (again) and the inconvenience for end users that let me seek another way to handle it.

Many thanks to @nojus297 whose colaboration made this PR possible.

vaijab:
Would there be a way to run the build locally to produce patched fonts? I am so keen to get my hands on properly scaled Hack font.

Finii:
If you have `font-patcher` already somewhere locally (if not, see [here](https://github.com/ryanoasis/nerd-fonts#option-5-clone-the-repo)), you can just download the [changed script](https://github.com/ryanoasis/nerd-fonts/blob/feature/correct-small-rendering/font-patcher) from this branch. Only the `font-patcher` file is changed.

Or you need https://github.com/ryanoasis/nerd-fonts/blob/feature/correct-small-rendering/font-patcher
and all files (recursively) from https://github.com/ryanoasis/nerd-fonts/tree/feature/correct-small-rendering/src/glyphs (in the correct relative directory).

If you use git
```
git fetch
git checkout feature/correct-small-rendering
```

Or whatever ;) I do not know where you are.
If you use a container or Linux-On-Windows, I have no clue ;)

If you need only a very few specific fonts (and tell me which, and wether `--mono` or `--windows`) I might find time to put them somewhere. But please only if you can not do it yourself.

vaijab:
> Or whatever ;) I do not know where you are. If you use a container or Linux-On-Windows, I have no clue ;)
> 
> If you need only a very few specific fonts (and tell me which, and wether `--mono` or `--windows`) I might find time to put them somewhere. But please only if you can not do it yourself.

This is great, thank you so much. I use linux, no problem. I'll figure it out!

nojus297:
Thanks for the fix, and the exclusion of `fontTools` as a dependency is nice. Although this does make me wonder if some tests or checks would help to prevent regression failures in the future.

Finii:
What kind of checks to you have in mind?

The actual code this PR adds has been tested with the help of `fontTools`, and I think adding it as test-dependency is quite ok. But I have a problem introducing normal unit tests here; it would require possibly some restructuring of everything. I would also split the code in smaller files as modules, test the modules independently etc, and then 'build' the end-user `font-patcher` from the individual files.

_Some random thoughts following._

But at the moment the concept here is to have everything (also the built artifacts like patched fonts) IN the repo and not only as release artifacts. For easier access by users I assume.

What I also look upon with mixed feeling is that the release is defined only as [workflow](https://github.com/ryanoasis/nerd-fonts/blob/master/.github/workflows/release.yml). Having even some kind of `make release` Makefile would be more desirable in my opinion, as that allows to produce exactly 'the release' on a local machine easily; and having the github workflow only doing the upload-files-as-release stuff.

And the tests would then be triggered by `make test`, as usual. THAT can be triggered by a commit hook.

Use of `make` is just exemplary, usually I use Meson, but even a central shell script would be ok, that calls the appropriate sub-scripts in `bin/scripts/`.

The question is... is Nerd-Fonts complex enough to change from a github-workflow based release definition to a proper stand alone definition? Hmm.

nojus297:
I don't doubt that your code is correct and tested. I'm just worried that, as the patcher code is already quite complex, some subtle changes may break it in the future. And it will be hard to notice, as it wasn't easy to understand/reproduce the issue.

Therefore, for this case, I had in mind some sort of CI test to simply check if the flags of generated fonts are correct, e. g. with `fontTools`. It could be integrated after the patching workflow.

This was inspired by the fact that most recent font builds/releases do not work at all on my machine, they don't show up in font picker. Regression happens easily.

---

Storing patched fonts in the repository is certainly a bad idea, the `.git` folder is already `7.4G` in size, making cloning really slow.

Also separating patching script from github-releasing workflow is probably a good idea, as is restructuring of the whole project.

By the way, also noticed that `fontTools` is already installed by workflows: https://github.com/ryanoasis/nerd-fonts/blob/877887cac2d6ccc7354a8d95e8c39fff10bf120b/.github/workflows/font-patcher.yml#L52

Finii:
> Therefore, for this case, I had in mind some sort of CI test to simply check if the flags of generated fonts are correct, e. g. with `fontTools`. It could be integrated after the patching workflow.

Added, utilizing fontforge's own toolset ;-)
Thanks for the suggestion.

jirutka:
It fails if the font file is `sfd`:

```
$ font-patcher './src/unpatched-fonts/NerdFontsSymbolsOnly/NerdFontsSymbols 1000 EM Nerd Font Complete Blank.sfd' --complete --ext ttf
...
Done with Patch Sets, generating font...
Traceback (most recent call last):
  File "font-patcher", line 1191, in <module>
    main()
  File "font-patcher", line 1186, in main
    patcher.patch()
  File "font-patcher", line 253, in patch
    source_font = TableHEADWriter(self.args.font)
  File "font-patcher", line 140, in __init__
    self.flags = self.getshort('flags')
  File "font-patcher", line 49, in getshort
    return (ord(self.f.read(1)) << 8) + ord(self.f.read(1))
TypeError: ord() expected a character, but string of length 0 found
```

Finii:
> It fails if the font file is `sfd`

Thanks for the report @jirutka! Fixed.

Issue #762, opened 2022-01-12
Update codepoint ranges


jmederosalvarado:
#### 🎯 Subject of the issue

After the addition of various symbols like `codicons`, codepoint ranges in the wiki are outdated. Could we please update it? It would be great to update the website as well.

I'd be happy to submit a PR if someone pointed out how to extract codepoint ranges from the patcher script.

Finii:
I'm not sure if Ryan's thumps-up means it will be fixed or it would be great to create a PR.

Anyhow, the ranges are given on lines 562+

```python
"Seti-UI + Custom",         'Exact': False,                               'SymStart': 0xE4FA, 'SymEnd': 0xE531, 'SrcStart': 0xE5FA, 'SrcEnd': 0xE631
"Devicons",                 'Exact': False,                               'SymStart': 0xE600, 'SymEnd': 0xE6C5, 'SrcStart': 0xE700, 'SrcEnd': 0xE7C5
"Powerline Symbols",        'Exact': True,                                'SymStart': 0xE0A0, 'SymEnd': 0xE0A2, 'SrcStart': None,   'SrcEnd': None  
"Powerline Symbols",        'Exact': True,                                'SymStart': 0xE0B0, 'SymEnd': 0xE0B3, 'SrcStart': None,   'SrcEnd': None
"Powerline Extra Symbols",  'Exact': True,                                'SymStart': 0xE0A3, 'SymEnd': 0xE0A3, 'SrcStart': None,   'SrcEnd': None
"Powerline Extra Symbols",  'Exact': True,                                'SymStart': 0xE0B4, 'SymEnd': 0xE0C8, 'SrcStart': None,   'SrcEnd': None
"Powerline Extra Symbols",  'Exact': True,                                'SymStart': 0xE0CA, 'SymEnd': 0xE0CA, 'SrcStart': None,   'SrcEnd': None
"Powerline Extra Symbols",  'Exact': True,                                'SymStart': 0xE0CC, 'SymEnd': 0xE0D4, 'SrcStart': None,   'SrcEnd': None
"Pomicons",                 'Exact': True,                                'SymStart': 0xE000, 'SymEnd': 0xE00A, 'SrcStart': None,   'SrcEnd': None
"Font Awesome",             'Exact': True,                                'SymStart': 0xF000, 'SymEnd': 0xF2E0, 'SrcStart': None,   'SrcEnd': None  
[...]

PR #778, opened 2022-02-04
Feature: stable codepoints

Finii:
#### Description

The Font Logos and Octicons codepoints depend on the parallel existence
of FontAwesome. I.e. Font Logos is shifted of Octicons or FontAwesome is
also present in the patched font; Octicons is shifted if FontAwesome is
present.

This means that people, although using a Nerd Font, can expect the
symbols in different locations. The reason is clear; people that just
want one or some symbols and use a specifically patched font will be
able to use the original symbol font codepoints.

But I guess that these uses are nonexisting. Almost all will use
'complete' patched fonts and that is what we deliver and document.

To make the documentation less complicated we should fix the code point
ranges that a specific symbol set will be patched in at.

The two possible places where Octicons / Font Logos ends up are there
since they have been added back in 2015/6 (commits 9620d47ae, f933b5a2).

#### Requirements / Checklist

Make the codepoints of symbol sets stable, i.e. independent of other also incuded symbol sets.

It also drops `SrcEnd` from the patch table with 159820b. This improves the maintainability of the code and should be cherry-picked even if the complete PR is not merged.

Generate `--complete` font before and after this changes, fonts are identical (maybe use fontforge font compare, or `cmp`) (I did this).

Came up with #762 i.e. on https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points.


Finii:
Please consider 159820b independently, that commit just improves code maintainability with no functional change.

Issue #786, opened 2022-02-16
patched-fonts folder contains obsolete files


Finii:
#### 🗹 Requirements

- [ ] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [ ] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [ ] I have searched the [Wiki](https://github.com/ryanoasis/vim-devicons/wiki) for help

#### 🎯 Subject of the issue

When fonts files get new names the github action adds the new files, but does not remove old ones. It is written under the assumption that names do not change and the expectation that it will overwrite any old font file,

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_

#### ★ Screenshots (Optional)

![image](https://user-images.githubusercontent.com/16012374/154203592-2e2e6aec-bbb9-480a-a20e-a2861498188c.png)

_Three year old next to 10 days old_

PR #787, opened 2022-02-16
patch-em-all: Purge destination dirs if possible

Finii:
**[why]**
When the file names of the source files change which happened for example
with these commits:
```
  ac432eb20 Updated Inconsolata source to latest upstream version of 2.001 (WIP #289)
  9c5ad2c78 Updated Inconsolata source to latest upstream version of 2.001 (WIP #289)
```
or the patched font files naming changes because of any other reason (*):

The patched-font directory will contain the new files along with the
unchanged old ones (because they where not overwritten).

Typically when manually updating the patched-fonts this is not a
problem, as the maintainer can clean this up by hand (if it is noticed).
But with a github action we might want to have that automatized.

To not deter the usability of the script for end-users or for patching
single fonts of a collection we do NOT want to purge 'all old files'
because we can not know if they are really old or not.

**[how]**
For each directory that we process from the source fonts we check if all
font files therein match our search criterion (pattern, $2). If we are
going to patch _all_ files that are in that source directory we delete all
font files in the destination directory; expecting that all files will
be recreated.
If we do _not_ patch _all_ files, we can do nothing, because we can not
decide if the existing files originate from one of the
not-to-be-processed source font files or are zombies.

Fixes: #786

(*) For example when the font naming changes because
- The FontnameParser is used where `Nerd Font` moves forward before the style in the font name
- We will finally create font files without blanks in them, but conventional `Name-Weight.ttf`

#### Requirements / Checklist

Remove font files from `patched-fonts` that are not created on the most recent complete patch run.

Inconsolata has some orphan font files, run the patched `patch-em-all` and see those vanish.

#786

### DIFF
--- a/bin/scripts/gotta-patch-em-all-font-patcher!.sh
+++ b/bin/scripts/gotta-patch-em-all-font-patcher!.sh
@@ -62,12 +62,18 @@ echo "$LINE_PREFIX Total source fonts found: ${#source_fonts[*]}"
 function patch_font {
   local f=$1; shift
   local i=$1; shift
+  local purge=$1; shift
   # take everything before the last slash (/) to start building the full path
   local patched_font_dir="${f%/*}/"
   # find replace unpatched parent dir with patched parent dir:
   local patched_font_dir="${patched_font_dir/$unpatched_parent_dir/$patched_parent_dir}"
 
   [[ -d "$patched_font_dir" ]] || mkdir -p "$patched_font_dir"
+  if [ -n ${purge} -a -d "${patched_font_dir}complete" ]
+  then
+    echo "Purging patched font dir ${patched_font_dir}complete"
+    rm ${patched_font_dir}complete/*.[to]tf
+  fi
 
   config_parent_dir=$( cd "$( dirname "$f" )" && cd ".." && pwd)
   config_dir=$( cd "$( dirname "$f" )" && pwd)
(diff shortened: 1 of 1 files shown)


bin/scripts/gotta-patch-em-all-font-patcher!.sh
```diff
@@ -62,12 +62,18 @@ echo "$LINE_PREFIX Total source fonts found: ${#source_fonts[*]}"
...
   # take everything before the last slash (/) to start building the full path
   local patched_font_dir="${f%/*}/"
   # find replace unpatched parent dir with patched parent dir:
   local patched_font_dir="${patched_font_dir/$unpatched_parent_dir/$patched_parent_dir}"
 
   [[ -d "$patched_font_dir" ]] || mkdir -p "$patched_font_dir"
+  if [ -n ${purge} -a -d "${patched_font_dir}complete" ]
+  then
+    echo "Purging patched font dir ${patched_font_dir}complete"
+    rm ${patched_font_dir}complete/*.[to]tf
```
Finii, line 75: It does not purge out `readme`s or any other files.
But maybe we want to do that (purge out **all** files), so that the `patched-fonts/` dir just contains all autogenerated content.

If so we can change the new line 75 accordingly.

On the other hand, in that case the workflow could just wipe the `patched-fonts/` dir before calling `patch-em-all`; and we do not need any of these changes (i.e. this PR)
The drawback of THAT solution is, that it fixes it for our github workflow, but not for any self patching user.

Finii:
What I did not check is if the https://github.com/EndBug/add-and-commit action works like ordinary (command line) `git add`.

On the command line, adding a directory marks all missing files in that directory as deleted on the stage and in the later commit.
The action allows also `remove:` so I'm not sure they changed that behavior (I believe: no, they just call `git add` and `git rm` without any special logic added).

Finii:
Here is an instance where the present behavior (keeping old files laying around after renaming etc) results in an Issue...
https://github.com/termux/termux-styling/pull/77

> This seems to have gone unnoticed here as because the old Fura* files remain (unupdated) in the Nerd Fonts repo (by mistake, I presume), the download script didn't break

Issue #765, opened 2022-01-17
Remove the empty glyph from 1000-em variants at codepoint U+0045 (Latin capital letter `E`)


arnavyc:
#### 🗹 Requirements

- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [X] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [X] I have searched the [Wiki](https://github.com/ryanoasis/vim-devicons/wiki) for help

#### 🎯 Subject of the issue

All 1000-em variants contain an empty glyph at codepoint U+0045 (Latin capital letter E)

#### 🔧 Your Setup

- _Which font are you using?_ - Fira Code with fontconfig (Arch package [`ttf-nerd-fonts-symbols-mono`](https://archlinux.org/packages/community/any/ttf-nerd-fonts-symbols-mono/))
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_ GNOME terminal
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ Arch Linux

#### ★ Screenshots (Optional)

Looking at TTF in fontforge (https://github.com/ryanoasis/nerd-fonts/issues/581#issuecomment-837335900)

![image](https://user-images.githubusercontent.com/13703064/149715648-40e3137f-119c-4e4c-8b4b-ead93ca796fa.png)

#### Suggestion

Remove the empty glyph from 1000-em variants at codepoint U+0045 (Latin capital letter `E`). This will also fix #581.

Finii:
See #753, marked there as 'fixed' but maybe only the main issue not the `'E'`

> Additionally:
All 1000-em variants contain an empty glyph at codepoint U+0045 (Latin capital letter E) !??

WhyNotHugo:
Duplicate of #581?

Finii:
Thank you for the input.

TO references that Issue already in the description.

But didn't Ryan already address that? It's still open :thinking:

Finii:
It IS solved, see https://github.com/ryanoasis/nerd-fonts/issues/581#issuecomment-1057969427

Doublechecking...
:-o

What the heck?

Finii:
Double checked doing the patching, there are no `E`s in the font.

[...]

Issue #581, opened 2021-02-14
Symbols-1000-em Nerd Font Complete.ttf has broken letter 'E'


fwsmit:
#### 🎯 Subject of the issue

The capital letter 'E' from 'Symbols-1000-em Nerd Font Complete.ttf' is not working correctly. It seems like either it only contains that letter, or that letter is prioritized over other fonts. Anyhow, this leads to the letter 'E' not showing up in mono fonts in firefox. See below for an example.

Arch bug report: https://bugs.archlinux.org/task/69568?project=5&string=ttf-nerd-fonts-symbols-mono
Forum post: https://bbs.archlinux.org/viewtopic.php?pid=1955896

#### 🔧 Your Setup

- 'Symbols-1000-em Nerd Font Complete.ttf'
- Program: firefox
- OS: Arch linux.
- Package: ttf-nerd-fonts-symbols-mono 2.1.0+36+gd0bf73a1-2

#### ★ Screenshot

![example](https://user-images.githubusercontent.com/23078054/106174382-44f72100-6195-11eb-912d-0a5670384213.png)

StaticRocket:
bumping the issues because it's still prevalent
![external-content duckduckgo com](https://user-images.githubusercontent.com/35777938/114814079-6d002480-9d81-11eb-8aaa-4d1f6d313c39.png)

polyzen:
In the meantime you could use a more tailored font configuration file, eg.:
https://github.com/polyzen/dotfiles/blob/master/workstation/.config/fontconfig/fonts.conf

I haven't used 10-nerd-font-symbols.conf and don't have this issue.

Strahinja:
It also seems as if capital "E" is represented by blank space only in "1000 em" version and not in "2048 em" version (previews by [fontpreview](https://github.com/sdushantha/fontpreview)):
![https://i.imgur.com/tPxgY6b.png](https://i.imgur.com/tPxgY6b.png)
![https://i.imgur.com/okEx2TR.png](https://i.imgur.com/okEx2TR.png)
![https://i.imgur.com/pMoP0Pj.png](https://i.imgur.com/pMoP0Pj.png)
![https://i.imgur.com/8vn62eo.png](https://i.imgur.com/8vn62eo.png)

Strahinja:
[...]

Issue #479, opened 2020-05-24
[Suggestion] Improve icons-only "Symbols Nerd Font"


hpfr:
The implementation of the icons-only font in #84 is very useful, and I think it could be improved.

1. Based on 3b5dff7, it appears the icons-only font is just a ttf file in `src/glyphs`. Instead of just providing the ttf in the repo, could you somehow leverage the standard patcher and just provide an empty font as input? Or maybe you could add an option to the patcher like `--no-input`? I'm not sure of the best way to go about it since I'm not familiar with the code, but this way, the icons-only font won't get out of date and will keep up with all the other Nerd Fonts. Right now, I think the icons-only ttf is missing the Crystal icon added in `original-source.otf` in 3bdc094, based on the fact that it was added 4 months ago and the ttf's are from 2 years ago. Scripting them like the other Nerd Fonts would eliminate this problem. Also, should you generate otf's instead of ttf's since they have more features? Since it's just icons, you get to choose, right? I guess you could even do both.
2. Generate proportional and monospaced variants as with standard Nerd Fonts, i.e. "Symbols Nerd Font" and "Symbols Nerd Font Mono". This would allow users to select the proportional font for use outside of monospaced contexts (terminal emulators) to get full-size icons. For example, in window manager status bars.
3. Provide these two font variants in the releases instead of burying them in the repo for users to find ;D.

Thank you for your work on this excellent project!

polyzen:
> Right now, I think the icons-only ttf is missing the Crystal icon added in original-source.otf in 3bdc094, based on the fact that it was added 4 months ago and the ttf's are from 2 years ago.

This PR has the symbols fonts rebuilt for recent changes: https://github.com/ryanoasis/nerd-fonts/pull/461

> Scripting them like the other Nerd Fonts would eliminate this problem.

👍

ryanoasis:
Yes good suggestions and you are totally right that right now they are buried and easily out dated

hpfr:
Cool, it looks like 1 and 3 will be taken care of thanks to #461 in the next release. Is that coming up anytime soon?

Does the `2048` variant address point 2? I'm guessing that's full-width non-monospaced icons?

ryanoasis:
> Is that coming up anytime soon?

😟 I don't want to promise anything, I'm catching up on PRs but yeah the releases have been really poor lately. Hoping to automate things and just release monthly or something. Since you asked

[...]

PR #801, opened 2022-03-03
Bugfix: Automatic symbols only font generation

Finii:
#### Description

**[why]**
The symbol only fonts `Symbols-1000-em Nerd Font Complete.ttf`
and `2048-em` in `NerdFontsSymbolsOnly/` are generated from
some 'almost' empty source fonts, that are assumable in turn generated
from the sfd font descriptions in `src/glyphs/`?

The process is not documented and we have issues in the generated font
(for example the glyph for capital `E` is defined (and empty) #581 #765).

**[how]**
Use the existing font definitions from `src/glyphs/*.sfd` directly as
source font. That needs a change in `font-patcher` because the empty
fonts have no glyphs that can be used to orient the scaling upon. In
that case scale on the source font definitions EM.

Then we need patch-em-all to also patch `*.sfd` fonts.

And finally we need `patch-em-all` to take a font specific command line
switch for `font-patcher` (compare 9e2bc9a2 of #723) to instruct it to
create a `ttf` rather than a `sfd` front file.

In the sfd file we additionally set the Panose type.
And the UnderlinePosition is adjusted to match the current patched font.

[note]
Also fix wrong glob pattern in patch-em-all `*.[o,t]tf`. The comma is
for sure some leftover from a '{}' shell pattern, that is not used
anymore.

#### Requirements / Checklist

Autogenerate the symbols only fonts from the `sdf` templates and not the `ttf` templates.

Automatic generation from `ttf` files: #479
Issues with `ttf` files: #581 #765


Finii:
I _guess_ the empty `E` letter comes from the problem that `font-patcher` needs (*) at least one valid glyph in the basic range (`0000` - `017F` to determine the font metrics. So probably Ryan added an empty `E`, patched, and then removed the `E` - all manually?

That does not make much sense either ;-) Whatever.

_*: Well, it does not need it anymore with the PR, if no glyph is existing the EM is used_

Finii:
I consistently typed `front` instead of `font` in the commit messages :->

Fixed and force push to get rid of that embarrassment.

Finii:
Rebase on master, resolve conflicts by changing to `-iregex`, force push

Issue #610, opened 2021-04-17
Noto Fonts "Mono" variant is too widely spaced


rrobertobt:
#### 🎯 Subject of issue

The "NotoSansMono Nerd Font ***Mono***" seems to have been spaced more widely than the **non-mono** version "NotoSansMono Nerd Font"
The fonts were downloaded from the official Nerd Fonts website.
This issue also happens with RobotoMono Nerd Font: #597 

Screenshots are below

#### 🔧 My current set-up

- _Noto Sans Mono Nerd Font Mono 11_
- _GNOME Terminal, Alacritty_
- _Fedora Linux 34 x86_64_

#### ★ Screenshots

- **Font on GNOME Font Viewer - Opening the "Noto Sans Mono Regular Nerd Font Complete _Mono_.ttf" file**
![mono](https://user-images.githubusercontent.com/44622828/115101893-d65c7080-9f04-11eb-95ee-e65c7b12798b.png)
The space between characters is too wide

- **Font on GNOME Font Viewer - Opening the _Not Mono_ variant**
![non-mono](https://user-images.githubusercontent.com/44622828/115101985-85994780-9f05-11eb-9961-f0d6bfdb4fe2.png)
Spacing is correct

- **After setting font on GNOME Terminal**
![Screenshot from 2021-04-16 22-29-27](https://user-images.githubusercontent.com/44622828/115102003-a3ff4300-9f05-11eb-88c5-7ad9872aff22.png)

- **After setting font on Alacritty**
![alacritty](https://user-images.githubusercontent.com/44622828/115102023-ba0d0380-9f05-11eb-9bd3-bc5fe63c470a.png)

David-Else:
Exact same thing here with Literation Sans
![Screenshot from 2021-10-18 19-43-06](https://user-images.githubusercontent.com/12832280/137788618-2d49325a-d38a-410b-b1a8-072f1ea0106a.png)

pravin:
Bumping up. Anybody have ideas on how to fix this?

ZetaNumbers:
> Bumping up. Anybody have ideas on how to fix this?

I am using "NotoSansMono Nerd Font" inside the terminal

Finii:
Hmm, easy to fix :roll_eyes:, MR created.

PR #830, opened 2022-05-14
Fix Noto too wide

Finii:
#### Description

**[why]**
The 'monospace' width is determined by examining all the 'normal' glyphs
and taking the widest one.

'Normal' means `0x00`-`0x17f`: the Latin Extended-A range.

Unfortunately Noto has the '1/2', '1/4', '3/4' that are all wider then the
normal (i.e. letter) glyphs.

**[how]**
Exclude a small sub-range from the 'find the widest glyph' that contain
no glyphs one would call 'letter'.

#### Requirements / Checklist

Fix `Noto * Nerd Font Mono` variants.

* #610

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -654,6 +654,8 @@ class font_patcher:
         #
         # 0x00-0x17f is the Latin Extended-A range
         for glyph in range(0x00, 0x17f):
+            if glyph in range(0x7F, 0xBF):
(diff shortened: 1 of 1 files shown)


Finii:
Highligthed are the glyphs that are excluded from width-determination with the PR.

image

PR #837, opened 2022-05-28
Add Braille support

### LINKED ISSUE #482: Support for Braille Characters?
#### 🎯 Subject of the issue
Just a request/suggestion :)
Would be nice to have an option to add a braille font (https://aur.archlinux.org/packages/ttf-ubraille/)
Some popular tools such as gotop use this font to render pixelart, such as graphs, in the terminal.
An example is shown below.

image

jnaskali:
#### Description

Adds braille support to render pixelart (e.g. in terminal applications) requested in #482.
Thanks @Chaz6 for [the implementation idea](https://github.com/ryanoasis/nerd-fonts/issues/482#issuecomment-727270731)!

#### Requirements / Checklist

- [x] Verified the license of [UBraille.ttf](https://yudit.org/download/fonts/UBraille/UBraille.ttf) to be [GPL 2](https://yudit.org/download.html)

Adds UBraille.ttf to glyph sources and updates font-patcher so that it copies symbols from 0x2800 to 0x28FF.

Check that fonts are created correctly.

Closes #482.

example image
Screenshot by [dominikheinz](https://github.com/dominikheinz) from the related issue.


font-patcher
```diff
@@ -609,6 +615,7 @@ class font_patcher:
...
             {'Enabled': self.args.octicons,             'Name': "Octicons",                'Filename': "octicons.ttf",                                   'Exact': self.octiconsExactEncodingPosition,  'SymStart': 0xF27C, 'SymEnd': 0xF27C, 'SrcStart': 0xF4A9, 'SrcEnd': 0xF4A9, 'ScaleGlyph': OCTI_SCALE_LIST,  'Attributes': SYM_ATTR_DEFAULT},  # Desktop
             {'Enabled': self.args.codicons,             'Name': "Codicons",                'Filename': "codicons/codicon.ttf",                           'Exact': True,                                'SymStart': 0xEA60, 'SymEnd': 0xEBEB, 'SrcStart': None,   'SrcEnd': None,   'ScaleGlyph': None,             'Attributes': SYM_ATTR_DEFAULT},
+            {'Enabled': self.args.braille,              'Name': "Braille",                 'Filename': "UBraille.ttf",                     'Exact': False,                               'SymStart': 0x2800, 'SymEnd': 0x28FF, 'SrcStart': None,   'SrcEnd': None,   'ScaleGlyph': None,             'Attributes': SYM_ATTR_DEFAULT},
```
Finii, line 618: I guess we need a ScaleGlyph here, to have a working `mono` version?
About the ScaleGlyph ... without one the patching will maximize the size of each glyph individually to fill the 'cell' completely. But with Braille you want to scale all glyphs _identical_ to keep the relative size/position differences intact.

Further the `'Attributes': SYM_ATTR_DEFAULT` seems an odd choice, as that means `'align': 'c', 'valign': 'c', 'stretch': 'pa'`. That does only make sense if a `ScaleGlyph` is active? And maybe we want a `careful` here?

Also maybe correct the whitespaces, the vertical alignment (nothing I like normally, but we have it here) is broken as you see from the diff. Maybe a tab instead of spaces?

Finii:
Great work creating this PR :+1: 

Source of the glyphs mentioned in commit: :heavy_check_mark: 

As this is not in a PUA, I think maybe it should be `careful` to preserve pre-existing glyphs. It is possible to have it in the `attributes` even it is not used by any patch set at the moment. (Compare diff comments on attributes below).
"Think maybe" => open question

Finii:
Additional note

If you edit the commit, maybe you could add to the commit message the `Fixes: #837` line (on a line of its own at the bottom).
That makes it easier to go from `git blame` to the Issue and the MR.

Finii:
Same font packaged here: https://aur.archlinux.org/packages/ttf-ubraille

The font contains this copyright notice
`Created by Vyacheslav Dikonov with PfaEdit 1.0 (http://pfaedit.sf.net). Free to use and distribute.`

Same font is mentioned here https://fontforge.org/archive/sfds/index.html as 'GNU licensed'

Note that GNU FreeFont uses a Braille set by Steve White [link](https://www.gnu.org/software/freefont/sources/).

font-patcher
jnaskali: Thank you for looking through the code. I amended the vertical alignment and moved the issue number to a line of its own in the commit message.

I'm afraid I don't know enough about fonts to touch the attributes, so (everyone) feel free to improve things.

Finii: Here the scale effect for illustration:

image

Lets fix it...

Finii: Fixed in additional commit:

image

Finii: After some thought I believe we do not want `careful` here, at least not a partial one.
If there is any glyph in the range it will not match our new glyphs in style and size.
I guess best is to not be careful here.

Finii:
For me this looks good to merge now; playing the ball to @ryanoasis

font-patcher
Finii: OH NOO, it is not working, the scaling is uniform but the centering is individual? Now we have a lot centered single dots!

Examining...

Finii: Well, `ScaleGlyph` can not (by design) enforce consistent shifting/centering.

This needs some thought.

Finii:
It's a whole lot more involved than I initially thought. Split that up into individual commits that introduce the changes step by step.

Looks good now:
image

The changes are (mostly) only active when patching in the Braille font.

Also noticed that one glyph in the source font is missing: `dots-67` is empty

image

Although it should have dot 6 (2nd colum, 3rd from top) and dot 7 (1st column, lowest row) active - as the tiny `fontforge` preview shows. I have no clue about Braille (maybe it is not needed), but it can lead to 'broken' displays in the usecase shown in the top of the Issue (use as graphical elements).

Finii:
Rebase on master, force push

Finii:
All the readmes need an update for the new parameter :stop_sign:

musjj:
Can this potentially fix the issue of gaps between the top and bottom of braille glyphs? Some terminals [custom render](https://wezfurlong.org/wezterm/config/lua/config/custom_block_glyphs.html) them to make sure the gaps are filled, but can this issue be solved from the font level?

I noticed the commit regarding padding, so I tried out the PR out of curiosity to see what differences does it make, but it seems that this patch just makes the gaps even bigger.

Before (old screenshot, ignore the colors):
image
After:
image

Finii:
Ah, what you would like is a stretch-x-y glyph rescale (like the powerline glyphs).

That would be nice in graphic scenarios like above, but look funny in normal (i.e. displaying actual braille text) ones, because the dots will get distorted. Hmm.

I can try a bit with that image above, but ... ping me if it takes too long, I might have forgotten to do it :-}

musjj:
Cool, looking forward to it :)
Do you think the dots need to be distorted? For reference, this is how it looks like in Wezterm, with the custom glyph rendering:
image

And the PR implementing this: https://github.com/wez/wezterm/pull/919

Finii:
> Do you think the dots need to be distorted?

Well, the 'cells' of some fonts are rectangular, more high than wide, like your cursor in the image above.
But some fonts have almost square 'cells'. If the dots are to 'fill' the 'cell' completely they must be scaled differently in X and Y, thus loosing the aspect ratio, thus round things can become ... ellipses?

image
_Example of more squarish cells_

Finii:
> And the PR implementing this: [wez/wezterm#919](https://github.com/wez/wezterm/pull/919)

Ah, they do not patch in a scaled source, they create the glyphs on the fly from circles drawn in the 'cell'.
That would also be possible, but more complex. Normally Nerd Fonts just aggregates glyphs, not creates new ones.

Finii:
@musjj Could you copy & paste that image into a comment here, so that I could try it?

I tried with this image, but that looks a bit fuzzy:

```
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⢈⣈⣌⣌⣬⡮⡦⠲⠳⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⣈⣌⣮⣮⣿⣿⣿⣿⠻⠓⠑⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣌⣬⣮⣮⣮⣎⣌⣌⣮⣿⣿⣿⣿⣿⣿⣿⣿⣝⣝⣝⣮⣮⣮⣮⣎⣌⣌⣌⣮⣮⣮⣮⣮⣮⣮⣮⣮⣾⣿⣿⡷⡷⠷⠳⠓⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠓⠁⠀⠀⣑⣽⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣟⢙⢙⢉⢈⢈⢈⢈⠈⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣈⣌⣮⣿⣿⡷⢳⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⢻⠳⠳⠓⠑⠑⠑⠑⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⢀⢈⢈⣌⣿⣿⣿⣿⣿⣮⣮⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣟⠻⡳⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣯⣎⢈⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⣠⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠏⠱⣷⣟⠳⠆⠀⣳⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣞⠙⠳⡷⣷⣿⣿⣯⢌⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠰⣷⣿⣿⣿⣿⡿⠳⠑⠑⠑⠳⠳⡳⡷⡷⡷⡷⠷⠀⠀⠀⠱⡇⠀⠀⣰⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣎⠈⠀⠐⠱⡷⣿⣯⢌⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠑⠳⡳⠳⠁⠀⠀⠈⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⣬⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣯⠌⠀⠀⠀⠀⠱⡳⣏⠈⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠑⡳⣧⣮⣎⣌⣌⣌⣌⣌⣌⣌⣬⣮⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠌⠀⠀⠀⠀⠀⠀⠁⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⢈⢈⢘⢹⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⠱⣷⣿⣿⣿⣿⣯⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣈⣮⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠗⠀⠀⠰⣿⣿⣿⣿⡿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣬⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⠓⠁⠀⠀⠀⠀⣱⣿⣿⣿⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⢀⣬⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠷⠓⠀⠀⠀⠀⠀⠀⠀⣰⣿⡿⠃⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⣨⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⠷⠳⢓⣽⣿⣿⡷⠳⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⣸⠷⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⣼⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡿⠓⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⡣⣦⣎⣌⢈⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⣸⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⠏⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⡳⣿⣿⣯⣎⢈⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣯⠱⣷⣿⢏⠀⠀⠀⠀⠀⠀⠀⠀⠀⢈⠈⠀⠀⠀⢈⢈⢈⢹⣷⣿⣿⣯⣎⠈⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣯⠈⠱⡷⣏⢈⠀⠀⠀⠀⠀⠀⠀⠑⣳⣯⣎⠈⢙⣿⣿⣿⣿⣿⣿⣿⣿⣿⣎⠈⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠰⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣮⣌⢘⠑⠁⠀⠀⠀⠀⠀⠀⠀⠀⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⢎⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠰⣿⣿⣿⡿⠱⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⡷⠳⠁⠀⢀⣨⣿⣿⣿⣿⣿⣿⠷⠳⠳⠳⡷⣿⣿⣿⣿⣿⠎⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠐⣷⣿⠃⠀⠐⣷⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣟⣍⣌⣌⣌⣌⣮⣿⣿⣿⣿⣿⣿⣿⠃⠀⠀⠀⠀⠀⠰⣿⣿⣿⣿⣏⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠑⠀⠀⠀⣰⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣷⣿⣿⣿⣿⣎⡾⠃⠀⠀⠀⠀⣿⣷⣿⣿⠟⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣸⣿⣿⣿⣿⡿⠓⠳⡳⣷⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⣿⢍⠀⠀⣷⣿⣿⣟⣝⢈⢈⣌⠀⠀⠀⠓⣰⣿⣿⠇⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⣬⣿⣿⣿⠷⠑⠀⠀⠀⠀⠀⠐⠱⠳⡷⡗⠱⡳⡷⣿⣿⣿⣿⢏⠀⠀⠐⠑⠑⠑⠑⠑⠀⠀⠀⠀⠀⣼⣿⠿⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⢈⣌⣮⡿⡷⠳⠑⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠱⣿⣿⡿⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣠⣿⠷⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠲⠳⠑⠑⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣼⠷⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣈⡿⠓⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠐⠁⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⣀⡶⠓⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀
```
(I created that with https://image2ascii.com/)
Maybe their algorithm expects the dots at other codepoints?

Here is the source image, with very smooth borders:

dragon

musjj:
Here you go:

```
  ⣴⣶⣤⡤⠦⣤⣀⣤⠆     ⣈⣭⣿⣶⣿⣦⣼⣆         
   ⠉⠻⢿⣿⠿⣿⣿⣶⣦⠤⠄⡠⢾⣿⣿⡿⠋⠉⠉⠻⣿⣿⡛⣦      
         ⠈⢿⣿⣟⠦ ⣾⣿⣿⣷    ⠻⠿⢿⣿⣧⣄    
          ⣸⣿⣿⢧ ⢻⠻⣿⣿⣷⣄⣀⠄⠢⣀⡀⠈⠙⠿⠄   
         ⢠⣿⣿⣿⠈    ⣻⣿⣿⣿⣿⣿⣿⣿⣛⣳⣤⣀⣀  
  ⢠⣧⣶⣥⡤⢄ ⣸⣿⣿⠘  ⢀⣴⣿⣿⡿⠛⣿⣿⣧⠈⢿⠿⠟⠛⠻⠿⠄ 
 ⣰⣿⣿⠛⠻⣿⣿⡦⢹⣿⣷   ⢊⣿⣿⡏  ⢸⣿⣿⡇ ⢀⣠⣄⣾⠄  
⣠⣿⠿⠛ ⢀⣿⣿⣷⠘⢿⣿⣦⡀ ⢸⢿⣿⣿⣄ ⣸⣿⣿⡇⣪⣿⡿⠿⣿⣷⡄ 
⠙⠃   ⣼⣿⡟  ⠈⠻⣿⣿⣦⣌⡇⠻⣿⣿⣷⣿⣿⣿ ⣿⣿⡇ ⠛⠻⢷⣄
     ⢻⣿⣿⣄   ⠈⠻⣿⣿⣿⣷⣿⣿⣿⣿⣿⡟ ⠫⢿⣿⡆    
      ⠻⣿⣿⣿⣿⣶⣶⣾⣿⣿⣿⣿⣿⣿⣿⣿⡟⢀⣀⣤⣾⡿⠃    
```
Rendered in Vim with Wezterm:
image

Finii:
Rebase on master, actually rebase on the code added by #916 which has the same problems.
Force push.

Finii:
Well, I do not know if this is really useful.

The braille glyphs are designed to always have one row of dots empty, so that one can distinguish between 'letters'.
Here is how the font originally looks like:

image

Dots have the **blue** diameter.
The leftmost (pale) dot is never there.
Between dots we have a gap of **green**
And the **brown** gap in the beginning and end of the glyph together are also one **green**.

This is in direct contradiction to the image shown above, where there are no visible vertical gaps from the columns of not-existing points.

Here you see the 'empty columns of dots' (arrows), and that they are missing in Wezterm:

image

This just talks about the vertical gaps, not the horizontal (line to line) ones.

Braille rendered with a font without char-to-char gaps will be unreadable.

Finii:
Lets assume we do not care if Braille is readable, because ... we are Nerds and Braille is not read from a screen but from a Braille line reader.

To get the padding right...

Horizontally we have `(2 * blue + 1 green) * padding% = 1 green`
Vertically we have `(4 * blue + 3 green) * padding% = 1 green`

This is not working if horizontal and vertical padding is the same; which is how it is currently with the patcher.

And for that we would still scale the glyphs not preserving aspect ratio, thereby ending up with ellipsis.

I guess the only good solution (for a graphical display) would be to generate the glyphs while patching the font with adapted gaps. This would also be possible but a totally different approach, rather like Wezterm.

musjj:
That seems like an interesting approach. Not really sure how fonts work, but thinking about it again, is it actually possible to get rid of the horizontal space between the lines? Aren't they rendered and controlled by the terminal instead of the fonts?
Maybe this isn't possible after all without custom glyph rendering done by the terminal.

Finii:
> is it actually possible to get rid of the horizontal space between the lines?

We get rid of it for the powerline glyphs, right? ;-)

image

They always touch top and bottom of the line (at least in terminal emulator environments).

clort81:
> Lets assume we do not care if Braille is readable, because ... we are Nerds and Braille is not read from a screen but from a Braille line reader.
> 
> To get the padding right...
> 
> Horizontally we have `(2 * blue + 1 green) * padding% = 1 green` Vertically we have `(4 * blue + 3 green) * padding% = 1 green`
> 
> This is not working if horizontal and vertical padding is the same; which is how it is currently with the patcher.
> 
> And for that we would still scale the glyphs not preserving aspect ratio, thereby ending up with ellipsis.
> 
> I guess the only good solution (for a graphical display) would be to generate the glyphs while patching the font with adapted gaps. This would also be possible but a totally different approach, rather like Wezterm.

In particular, braille characters are useful as box-drawing characters.
 
I ask the reader to now take comprehension of a few pertinent facts:

* Fonts are made for people with functional vision.  We view font  glyphs on screens, with our eyes.
* Blind people do not read braille on screens, with their eyes, which don't work, because they are blind.
* The Braille glyphs add 256 (2^(4x2)) distinct drawing glyphs, which allow us far greater flexibility in drawing to terminal than other glyphs in the BMP range.
* The style of braille glyphs is most useful for drawing when they are connecting blocks, as opposed to dots.   

Which looks better to you?  Dot-braille or block-braille?
dotbraille_vs_block_braille-menus
dotbraille_vs_block_braille-art
[fancy_monitor2.txt.gz](https://github.com/ryanoasis/nerd-fonts/files/10473468/fancy_monitor2.txt.gz)

Finii:
@clort81 Interesting design.
And they (the blocky ones) would not need to be generated for each font but can be patched in using `xy` scaling. They would be oblong and not square, just fitting the 'Braille grid' into one target font cell.

Finii:
https://github.com/microsoft/cascadia-code/issues/647

Finii:
The future of this is unclear.
The Braille glyphs have a function in unicode, and that is representing Braille.

This PR heads in the direction to re-assign the codepoints to 'some graphics glyphs', what is not exactly right in Unicode context.

So probably this will be only a self-patch option or something.

Moving this to parallel Universe to get it out of the mainstream development.

PhMajerus:
If you're interested, Unicode accepted another block of octants for 2×4 pseudo-pixels.

image

U+1CD00 to U+1CDE5 (it isn't the complete 256 semigraphic characters because some patterns already exist as quadrants or half-blocks).
The octants are documented at https://www.unicode.org/L2/L2021/21235r-terminals-supplement.pdf
Their planned inclusion in Unicode 16.0 is shown at https://www.unicode.org/alloc/Pipeline.html

Updating apps using braille to use the new octants should be easy, it should even be possible to write a small conversion utility that can be piped to that switches existing Braille to the new Octants without touching existing utilities that use Braille.

Finii:
@PhMajerus Thanks for the update. That certainly looks interesting.
I need to have a closer look.

flexiondotorg:
@Finii I wanted to revive some old fonts by adding Nerd Fonts and, as it turns out, Braille.

I have taken the relevant parts from this patch, updated them and applied them to `nerd-font-patcher` 3.1.1 (script version 4.8.2) in a custom Nix package. Here is the 13-line patch:

- https://github.com/wimpysworld/nix-config/commit/141b76ed2b0fcbe54e5fc82e3e3b99287e48858b

I've successfully patched the fonts I wanted :slightly_smiling_face: 

Would you like me to submit an updated pull request here?

Finii:
@flexiondotorg Thanks for the work on this. I would rather keep this 'the' PR about Braille, instead of having two PRs with diverging discussions. I will pull your commit in a minute.

Also wanted to mention, after skimming the comments above, that this PR's problems are not technical (anymore, as the needed algorithm changes are already merged) but conceptual.

Just some issues
* Shall Braille glyphs be usable for Braille (which means gaps for graphics)
* If for graphics use, scaling 'pa' will never work nicely - also 'xy' does not scale nicely on dots that become ellipses
* Blocks instead of dots will have overlap or inter-line and inter-glpyh gap problems depending on terminal
* Unicode accepted another block of octants for 2×4 pseudo-pixels
* Cascadia ended up adding both (blocks in new unicode block, and dots as classic Braille)
  https://github.com/microsoft/cascadia-code/pull/723#issuecomment-2048212464

When simply patching like you proposed, depending on target font we will get extraordinary width vertical gaps in graph renderings, i.e. when the font's cell is more square than oblong-tall, see for example Arimo:

image

And another example LiterationMono, comparison how the box drawing glyphs work that introduce a bit of overlap.

image

### Summary

I guess this works for some fonts but not in general, depending on cell aspect ratio.
And all the hurdles above are unsolved.

Possibly easier would be to autogenerate the glyphs with the right proportions instead of scaling an existing font (for the Braille glyphs), as then we can have dots but avoid big gaps and non-circular dots independent of cell aspect ratio.

Finii:
And to not forget, Braille-6-7 is still broken in 'upstream' `UBraille.ttf` (i.e. empty)

image

Interestingly it is ok in upstreams `UBraille.sdf` :grimacing: 

image

flexiondotorg:
Thanks for the summary @Finii :+1: 

It looks like I got lucky, as the font I most wanted patching (Fixedsys Excelsior) seems to have the ideal aspect ratio for Braille glyphs, and Fixedsys Core is pretty close. The following are from Blackbox, which doesn't have built-in glyphs and uses vte-gtk4 0.76.3.

## Fixedsys Excelsior

Fixedsys Excelsior Test

## Fixedsys Core

Fixedsys Core Test

flexiondotorg:
> And to not forget, Braille-6-7 is still broken in 'upstream' `UBraille.ttf` (i.e. empty)

There is an alternative Braille font I found that I have not tested yet.

- https://www.fontspace.com/braille-cc0-font-f96789

Is CC0 licensed acceptable?

Finii:
> Is CC0 licensed acceptable?

Sure. But I just exchanged the `ttf` for the `sdf`. Force pushed that just a minute ago :grimacing: 

Also removed the mentioning of "GPL" because UBraille is certainly not GPL; `Yudit` is.

Finii:
Original statement (in Russian, autotranslated) of UBraille creator [Vyacheslav](http://luc.devroye.org/fonts-56976.html)

```
This font is probably the world's first free Braille font of
the UNICODE standard. The font contains only dot characters
Braille in the range 0x2800 - 28FF. When creating it, the
following was used pfaedit editor and UCS fonts, also known
as -misc-fixed- for XFree86, as a sample. The font can be used
to display Braille text in yudit and other Unicode text editors.

Dikonov Vyacheslav, 2002
```


```
This is probably the first free unicode Braille font in the world.
It contains braille glyphs in the Unicode range of 0x2800 - 28FF only.
I used Linux and pfaedit editor to create this font and UCS bitmap 
fonts to verify the glyph shapes. It can be used in yudit and other
unicode editors to display braille texts.

This font and its source is free to modify, use, distribute, embed 
in documents and other larger scale unicode fonts. If you require 
a license, consider it to be GPL.


Vyacheslav Dikonov, 2002 <sdiconov@mail.ru>
```

Source: https://fontforge.org/archive/sfds/UBraille.tar.bz2

So is it GPL? Probably not strictly. GPL would impose some restrictions on us. :thinking: 

But to be safe, the CC0 font is certainly the better choice if we do not self-autogen the glyphs later on.

Finii:
Waaaah :sob: 

UBraille is even more broken?!? Comparsion of UBraille and BrailleCC0:

image

## Braille 3478

image
_left: BrailleCC0, right: UBraille_

https://en.wikipedia.org/wiki/Braille_pattern_dots-34#Plus_dots_7_and_8

image

Good that we dropped UBraille!

PhMajerus:
I think Braille should not be the same as the 2×4 pseudo-pixels that will be standardized in Unicode 16.0.
However, I also agree that Braille has been historically used as pseudographics in terminals for years, and there may be something to do with them as well to better accomodate that existing use.

With Unicode 16, we'll have separated 2×2 quadrants, which are mosaics like the standard quadrants, but not in a touching grid, as well as both sextants and separated sextants.
However, while Unicode 16 also adds octants, it does not add the equivalent separated octants.

You can probably see where I'm going here… to me, Braille could be the separated octants, using the same design as separated quadrants and sextants (grid-compatible with them).
See my suggestion in Cascadia's issues for the full detail and examples: https://github.com/microsoft/cascadia-code/issues/739

Finii:
> Braille with the Separated Octants style

I am inclined to implement it in this way, as also pointed out by musjj's and flexiondotorg's images.
It will make it useless for real Braille use, though. I think Braille glyphs are for seeing people that want to check Braille software output for example.

I know I will regret that if we get numerous issues that we broke all those Issues in from devs who write software for Braille hardware :grimacing: 

But in principle I guess for a Nerd Font it makes most sense to have the glyphs serve as 'generic graphic elements' (i.e. separated octants).

Ah, and further I do not see the Unicode 16 graphics building blocks added here soonish.
In principle I would rather spend the time to switch away from `fontforge` to be able to patch variable fonts...

GoldPigg:
@Finii I attempted to resolve this long-standing issue. Below is my solution:

- [braille-support](https://github.com/ryanoasis/nerd-fonts/compare/master...GoldPigg:nerd-fonts:braille-support)

If you need it, you can **pull it directly**, and note that I have already performed a **rebase**.

To address this issue, I introduced support for **dynamically generated glyph fonts** into FontPatcher. In fact, this is much simpler than it sounds. For details, please see below:
- [Allow generate symfont dynamicly](https://github.com/ryanoasis/nerd-fonts/commit/956fa7313109a14b52770064f9f2d3778755220a)

Perhaps NerdsFonts' philosophy is only-patch, but I believe introducing this feature wouldn't cause any harm and could offer simpler solutions to certain issues (like the one at hand).

Below is an example based on the `ComicShannsMono` font:


I sincerely hope this helps you.

Finii:
@GoldPigg :+1:
This looks very nice (I mean especially the implementation).
You create the classic round 'dots', what do you think about the octants style proposed above?

Finii:
Closed in favour of #1980

Finii:
Just to inform all followers of this PR: Braille got added via the other PR just now.

Issue #656, opened 2021-08-19
Please add missing from font-logos


Minecodes:
font-logos added new logos for Linux distros like Parrot OS, Kali Linux, Pop OS or Deepin OS.
Can you please add them? Because a Pull request has added them already ([github.com/ryanoasis/nerd-fonts/pull/551](https://github.com/ryanoasis/nerd-fonts/pull/551)).

siduck76:
the next release will have the updated icons

ryanoasis:
if you aren't particular about which font you use, you can try out https://github.com/ryanoasis/nerd-fonts/releases/tag/2.2.0-RC which has a small subset of fonts patched w/ the latest changes.

siduck:
> if you aren't particular about which font you use, you can try out https://github.com/ryanoasis/nerd-fonts/releases/tag/2.2.0-RC which has a small subset of fonts patched w/ the latest changes.

oh I use jetbrainsmono font , I'll wait :(

ryanoasis:
@siduck Added JetBrainsMono to the [2.2.0 RC](https://github.com/ryanoasis/nerd-fonts/releases/tag/2.2.0-RC) if you want to try it out. Thanks

Minecodes:
@ryanoasis does FiraCode have that update already?

ryanoasis:
@Minecodes sorry, what's the context? which update?

Minecodes:
> @Minecodes sorry, what's the context? which update?

for the new symbols like the Kali Linux icon

ryanoasis:
Ah, no FiraCode isn't updated. The RC is only doing a limited set of fonts.. I suppose we can change that soon. Just wanted to prove it out

siduck:
> @siduck Added JetBrainsMono to the [2.2.0 RC](https://github.com/ryanoasis/nerd-fonts/releases/tag/2.2.0-RC) if you want to try it out. Thanks

thanks! it works :D

![image](https://user-images.githubusercontent.com/59060246/146627697-6ad7ab87-9cfd-44ec-9aa7-572f4fa74616.png)

ryanoasis:
Just FYI all fonts are now being built in the RC

marcomeyerVFX:
are you planning to update to font logos v0.18 before the next release? seems like they are not in the latest RC fonts yet.

siduck:
> are you planning to update to font logos v0.18 before the next release? seems like they are not in the latest RC fonts yet.

[...]

PR #551, opened 2020-12-22
Update font-logos with new linux icons

daniruiz:
#### Description

font-logos is outdated and now comes with some extra Linux-related logos. This PR contains the new font but I've not patched any font, in case I broke something. But I did test it with the fira-code font, especially to check the new kali-linux logo that I've recently added to font-logos in another  [PR](https://github.com/lukas-w/font-logos/pull/56)

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [x] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [x] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### What does this Pull Request (PR) do?
Adds new icons from [font-logos](https://github.com/lukas-w/font-logos)

#### How should this be manually tested?
I created a branch with only fira-code patched, in order to easily test it without rebuilding all the fonts
https://github.com/daniruiz/nerd-fonts/tree/fira%2Bkali/patched-fonts/FiraCode

#### Screenshots (if appropriate or helpful)
image

PR #859, opened 2022-07-07
Update Font Logos to 1.0.1

Finii:
#### Description

Before Font Logos 1.0.0 there was no stable-codepoint guarantee, and the
later revisions (0.18) reassigned the codepoints. That makes updating
for us very hard, because we would have to resort the icons or follow
the codepoint changes.

With 1.0.0 this changed. That new major release introduced not only a
codepoint guarantee, but it also shifted the codepoints from the F100
region to the F300 region, where we put them anyhow (in most cases).
That is one subset less with 'dynamic base codepoint'.

The reassignment of codepoints in Font Logos kept the codepoints stable
for the icons we already have in release 2.1.0. But at the moment the
2.2.0 Release Candidate (which updated Font Logos with commit 557b8da)
have added glyphs, which are on different points as compared with the
Font Logos 1.0.0 release.

What I want to say, we need to get this straight, before our 2.2.0 RC
turns into a Release 2.2.0.

#### Requirements / Checklist

Update Font Logos to the latest release 1.0.0 (with codepoint stability guarantee)

https://github.com/lukas-w/font-logos/issues/72
https://github.com/lukas-w/font-logos/issues/33

#551 Puts new glyphs at codepoints that differ from the now-fixed and 'cast into stone' codepoints of Font Logos 1.0.0.

Discussion on codepoints also here https://github.com/ryanoasis/nerd-fonts/issues/656#issuecomment-1020933693 and following

Font Logos 1.0.0 glyphs at position `F300` and following, no gaps

image

A lot of thanks to @lukas-w who made the new Font Logos release possible and added codepoint stability as one of the objectives.


Finii:
See also #778 for our own codepoint stability. One half of that is obsoleted by the MR.

alesya-h:
`font-logos` has a 1.0.1 release with some fixes. Maybe it would make sense to use 1.0.1 straight away.

Finii:
Thanks for the ping, I'll update this PR.

Issue #858, opened 2022-07-06
Backtick character overlaps with the next character in patched font


krish-r:
Hello,

#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
The backtick (`` ` ``) character overlaps with the next character in the following patched fonts.
* Monolisa (paid)
* Inconsolata (open-source)

(There was already an issue #582 on this topic. However, it was closed and I could not find any solution there.)

Expected behavior:
The backtick (`` ` ``) character should not overlap with the next character like the original font.
 
Example symbols:
`` `i `` (or backtick with any other alphabet should also work)

#### 🔧 Setup

| Font        | Font Version | Format  | Font File Name                                                | Font Source                                                                      |
| ----------- | ------------ | ------- | ------------------------------------------------------------- | -------------------------------------------------------------------------------- |
| MonoLisa    | v1.808       | TTF/OTF | MonoLisa Regular Nerd Font Complete.ttf                       | Self Patched - using Nerd Fonts Docker image with `--complete` option               |
| Inconsolata | v2.012       | TTF/OTF | Inconsolata Regular Nerd Font Complete Windows Compatible.ttf | https://github.com/ryanoasis/nerd-fonts/releases/download/v2.1.0/Inconsolata.zip |

System Information
|              |                                                 |
| ------------ | ----------------------------------------------- |
| Applications | VS Code v1.68.1, Windows Terminal v1.13.11432.0 |
| OS           | Windows 11, Linux (Ubuntu)                      |
 

#### ★ Screenshots

| Font        | Original Font                                                                                                   | Patched Font                                                                                                    |
| ----------- | --------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------- |
| MonoLisa    | ![image](https://user-images.githubusercontent.com/54745129/177486100-c7ce2cd3-e14e-45b5-b24d-5989e0dc8167.png) | ![image](https://user-images.githubusercontent.com/54745129/177485968-42a7f8a9-9533-4ebb-ad1a-42a3afed0b50.png) |
[...]

Issue #582, opened 2021-02-15
backtick for Inconsolata NF overlaps with character afterwards


abelcheung:
## 🔧 Your Setup

- Windows 7
- Inconsolata NF Mono (Windows compatible, TTF)
- Visual Studio code (not sure if non-terminal usage warrants an issue report; trying anyway since it is a regression from unpatched font)

## Issue description

One can refer to screenshots below. Backtick from Inconsolata NF would overlap with char that follows. On that particular position, cursor would not advance at all, behaving like a zero-wdith glyph.

So far no character other than backtick has such behavior observed.

### Using unpatched Inconsolata downloaded from Google

![圖片](https://user-images.githubusercontent.com/83110/107895561-b4a63380-6f2b-11eb-8265-0226ac43df36.png)

### Using Inconsolata NF

![圖片](https://user-images.githubusercontent.com/83110/107895646-fc2cbf80-6f2b-11eb-9c68-a86f33cb3212.png)

Mazyod:
Seems like it's still an issue?

![image](https://user-images.githubusercontent.com/860511/120412628-8d9b3280-c35f-11eb-83d7-d438e47ba9fb.png)

Finii:
Will work on this in 858 because this is closed.

inostia:
Status update on this one? I just encountered it in VSCode.

PR #585, opened 2021-02-26
Update Iosevka unpatched fonts to 5.0.3

FollieHiyuki:
#### Description

Update Iosevka unpatched fonts

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [ ] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [ ] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### What does this Pull Request (PR) do?
Update unpatched fonts for Iosevka to 5.0.3 + LICENSE + README

#### Any background context you can provide?
I got this when running the `gotta-patch-em-all-font-patcher!.sh` script
```
Failed to remove subtable: 'dlig' Discretionary Ligatures lookup 33 contextual 11
```
Also using the tested font patched with `font-forge`, I got glyph overflow (2 pixels size for 1 pixel icon) and don't know what is the cause.
I have no idea what to do with either `config.cfg` or `config.json`.

#### What are the relevant tickets (if any)?
#455

PR #860, opened 2022-07-07
patcher: Fix grave (backtick) zero width

Finii:
#### Description

**[why]**
In ligature-enabled environments the accent grave will be rendered as
zero width, thus overlapping with the next character (in some source
fonts).

The problem is, that the glyph type in the sourcefonts is set to 'auto',
and fontforge exports these as 'mark' - the patched font will be broken.

**[how]**
There is no way we can get that glyph to be 'auto', but we can force it
to be an ordinary 'baseglyph' instead, and that will be respected on
export.

Maybe I should raise an Issue at fontforge... maybe later.

Fixes: #858
Fixes: #582

#### Requirements / Checklist

Always set the glyph type of `grave` to 'just a normal glyph'.

Use some patched font in a ligatures enabled environment, e.g. Writer, Word, VisualStudio, ... type in one "`" and the cursor should advance one slot.

https://github.com/adam7/delugia-code/pull/11

The ~~bug~~ peculiar behavior of `fontforge` is fixed:
* `head` (3253fff88) :heavy_check_mark: 
* `20220308` release :heavy_check_mark: 
* `20201107` release :negative_squared_cross_mark: 

'Before' screenshots are in Issue #585.
After this the glyph will always be 'one width' wide.

image

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -126,6 +126,14 @@ class font_patcher:
             symfont.close()
         print("\nDone with Patch Sets, generating font...")
 
+        # The grave accent and fontforge:
+        # If the type is 'auto' fontforge changes it to 'mark' on export.
+        # We can not prevent this. So set it to 'baseglyph' instead, as
+        # that resembles the most common expectations.
+        # This is not needed with fontforge March 2022 Release anymore.
+        if "grave" in self.sourceFont:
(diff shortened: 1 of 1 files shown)


Finii:
Not all to-be-patched fonts contain a `grave` :grimacing: 
Force push.

PR #867, opened 2022-07-15
fix(jetbrain-mono-font): installation guide

GervinFung:
#### Description

#### Requirements / Checklist

Fix the JetBrain Mono installation guideline

None

I used the command `brew install --cask font-jetbrains-mono` and the font displayed looks weird as shown [here](#Screenshots)
Then I tried `brew install --cask font-jetbrains-mono-nerd-font` according to [this](https://gist.github.com/davidteren/898f2dcccd42d9f8680ec69a3a5d350e?permalink_comment_id=4047567#gistcomment-4047567) and it works as shown [right here](#Screenshots)

None

| Before | After |
| --- | --- |
|  |  |

### DIFF
--- a/src/unpatched-fonts/JetBrainsMono/README.md
+++ b/src/unpatched-fonts/JetBrainsMono/README.md
@@ -22,10 +22,10 @@ Select JetBrains Mono in the IDE settings: go to `Preferences/Settings` → `Edi
     ```console
     brew tap homebrew/cask-fonts
     ```
-2. Install it using the `font-jetbrains-mono` cask:
+2. Install it using the `font-jetbrains-mono-nerd-font` cask:
 
    ```console
-   brew install --cask font-jetbrains-mono
+   brew install --cask font-jetbrains-mono-nerd-font
    ```
 
 ### Manual installation


Finii:
The page you change is in the `unpatched` directory. That directory contains the original (unpatched) fonts; and the README is also the file from the unpatched/original font. Compare with its repo: https://github.com/JetBrains/JetBrainsMono

Usually in the `unpatched` directory everyting is 'untouched' (more or less).

Maybe you wanted to change it in the `patched` directory? https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/JetBrainsMono/Ligatures

How did you end up looking _there_ for installation instructions? I'd just like to understand how people click along, and where the right spot is to intercept them. I know the docu is lacking, esp regarding packaged-by-3rd-party-stuff installation.

(My person problem is that I can not even understand WHY people might want to go the complicated route and install a font (it's just a font, come on!) with some special package software. But maybe I'm too used to get font from anywhere and just install the files with the system font install method (like: just click on it :grimacing:))

GervinFung:
> The page you change is in the `unpatched` directory. That directory contains the original (unpatched) fonts; and the README is also the file from the unpatched/original font. Compare with its repo: https://github.com/JetBrains/JetBrainsMono
> 
> Usually in the `unpatched` directory everyting is 'untouched' (more or less).
> 
> Maybe you wanted to change it in the `patched` directory? https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/JetBrainsMono/Ligatures
> 
> How did you end up looking _there_ for installation instructions? I'd just like to understand how people click along, and where the right spot is to intercept them. I know the docu is lacking, esp regarding packaged-by-3rd-party-stuff installation.
> 
> (My person problem is that I can not even understand WHY people might want to go the complicated route and install a font (it's just a font, come on!) with some special package software. But maybe I'm too used to get font from anywhere and just install the files with the system font install method (like: just click on it grimacing))

Hey I updated the fix according to what you said. I used `brew install` because it's convenient for me, may I know what method you chose to install fonts? Perhaps its just the matter of personal preference

Finii:
Let me add the link to the case here:
https://github.com/Homebrew/homebrew-cask-fonts/blob/master/Casks/font-jetbrains-mono-nerd-font.rb
I also find your referenced gist of use. Hmm.
Hope I will not forget to act on this :grimacing: 

I think this is a more general problem, but you solution is at least better then before. Thank you for the PR.

----

How I install fonts? Usually I buy fonts for a project at MyFonts or some other retailer. Usually afterwards you can download a zip that contains the family. That is the same with free fonts, often. And like the inidividual font zip files on our Release page.
Then unzip and double click on the font files. Done. Instead of double-click you can just copy the files from the command line, if that is your workflow. No need for any font-install-manager-magic (see below).

What is harder is to keep the fonts at bay. I strive to have the minimum number of fonts installed at one time. For example I HATE all these 250? Google Noto Fonts that are autoinstalled with Ubuntu. Font chooser dialog is unusable. The first thing to uninstall after fresh setup.
A long time I used FontyPython to manage the installed fonts, but that seems to be gone. Now I use https://github.com/FontManager/font-manager, which is a no-frills solution; or I just move the files in and out of my `~/.local/share/fonts` from my font storage, just with plain `cp` on the command line.
Unlike others I never had any issue with `fc-cache` and I never ever update anything there, it just works. Probably there is a icrontab set up or what, I have no clue. It just work, and I would expect modern OSes to just work.

On Windows it's the same, usually I copy directly. But sometimes Windows does not let go of a font file and you can not delete it. This different approach with file-detetion is really annoying. And Windows sometimes seems to keep old fonts cached. Well, what people say they have on Linux with `fc-cache` (where I never had the slightest glitch)..

Whatever! If you are still reading this.. thanks ;-)

Finii (approved):

GervinFung:
Thanks for merging this PR

PR #887, opened 2022-08-21
List AUR packages for Jetbrains & Victor Mono

avalonv:
#### Description
Added two fonts to the AUR section of the README

These are currently the 3rd and 4th most popular Nerd Fonts packages in the AUR respectively, and I felt they should be included.
#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?
See description.

#### Screenshots (if appropriate or helpful)
image

PR #462, opened 2020-04-10
Adding additional Archlinux links

quinnjr:
#### Description

Adds additional links to Archlinux AUR packages for Nerd Fonts in each language README.

#### Requirements / Checklist

- [x ] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x ] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [ x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [ ] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [x ] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### What does this Pull Request (PR) do?

This PR updates each language README with additional links.

#### How should this be manually tested?

Read and verify each README addition for proper markdown usage.

PR #894, opened 2022-08-26
Add nerd-fonts-hack AUR package to readme

chapa:
#### Description

Adds a link to nerd-fonts-hack AUR package for Nerd Fonts in readme (of all languages).

#### Requirements / Checklist

Read and verify each README addition for proper markdown usage.

I maintain the nerd-fonts-hack AUR package.


Finii:
Hmm, seeing all the translations... #887 did not touch translations :-(

In that PR I checked the 'popularity' level or something, that might be meaningless :->

Thanks for the PR!

chapa:
I saw #462 that did so I though they had to be updated too, didn't they ?

I didn't notice this list was meant to be non-exhaustive, feel free to close without merging if the popularity of the package is not high enough (you can see it [here](https://aur.archlinux.org/packages?O=0&SeB=nd&K=nerd-fonts-&outdated=&SB=p&SO=d&PP=50&submit=Go))

Finii:
The 2.2.0 release aimed mostly to fix bugs in the patcher itself, so the fonts are repatched (with fewer bugs I hope), but that does not mean that the _sources_ have been updated. I do not know how current the `Hack` source is here.
Updating sources can be the aim of 2.3.0 :thinking: 

The popularity is not so high, but the votes are. Whatever that means.
Will pull it after the release process finished - I fear the release will break if I add more commits while it run. :grimacing: 
Patching `Noto` takes ~5h :-( Just changed the release process to make the release official only after all fonts are ready (at the moment the release is published after the first font is ready :grimacing:

Finii:
Merged via a334db418063de67385c5594d36086e324fa98e8

chapa:
Thank you for the explanation, actually even if sources didn't change in 2.2.0 people generally like to have the latest version :wink:

FYI popularity is related to votes and time since when the package has been published (votes increases popularity and time decreases it):
image

Thanks for the merge :ok_hand:

PR #910, opened 2022-09-02
Update Hasklig readme

chezbgone:
The patched Hasklig font was updated to use Hasklig v1.2 in d4d8d9e184a487fd92412ff639b79b0d61908b91, but the readme was not updated. This PR updates the readme with the appropriate information found [here](https://github.com/i-tu/Hasklig/releases/tag/v1.2).

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1102,6 +1102,15 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "chezbgone",
+      "name": "Jason Chen",
+      "avatar_url": "https://avatars.githubusercontent.com/u/7344128?v=4",
+      "profile": "https://github.com/chezbgone",
+      "contributions": [
+        "doc"
+      ]
     }
   ],
   "repoType": "github",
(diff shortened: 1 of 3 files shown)


Finii:
Well, first I checked the version in the source fonts and **found it to be 1.0** :-o

image

And then I looked at the commit you referenced d4d8d9e184a487fd92412ff639b79b0d61908b91, and that commit **is broken**. Although it adds a patched version to the repo, it does NOT update the source font. And the patched fonts are autogenerated for releases, so ... this commit is completely meaningless.

image

Or am I mistaken? Do you use Hasklig Nerd Font? Has it the changes it should have at rev 1.2?

Finii:
Ah it is e2745c3e4

image

Finii:
Ok, version information seems to be broken in the release for `otf`, for `ttf` is is ok

https://github.com/i-tu/Hasklig/releases

And the file sizes are the same as in the release, so probably we have really `v1.2` :grimacing: 

image

Finii:
Thank you for the contribution!

Issue #658, opened 2021-08-21
Power Line half cirles streched


JanzenJohn:
#### Im trying to use the powerline symbols U+E0B6 and U+E0B4

- _Which font are you using_ "Symbols-2048-em Nerd Font Complete.ttf"
- _Which terminal emulator are you using_ None I'm trying to use them in polybar
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ Im using Linux (artix, kernel 5.13)

#### ★ Optional

![image](https://user-images.githubusercontent.com/51709412/130335124-77731407-6cea-427b-b8f7-3b7939b3f2c6.png)
two "half circles" should create cirle ?

Finii:
Hmm, your symbols look more distorted than encoded:

![image](https://user-images.githubusercontent.com/16012374/193281818-ebafd0da-1f33-46f6-8564-af232bd2bd8b.png)

These round Powerline Things were never half-circles, half-circles would look to small, that is the reason why they were designed to be more elliptic than round.

Finii:
Ah, I see the problem, it is in the `Mono` variant...

![image](https://user-images.githubusercontent.com/16012374/193283732-645d68dc-a7da-4423-840d-7a1ef11c4e10.png)

Finii:
I'm not really sure the `Symbols Only Mono` font does make any sense. What do you use it for? And why (not the non Mono one)?

Finii:
Working on this right now....

By the side: How bad is that, the source of the rounded things already has them ... strangely placed:

![image](https://user-images.githubusercontent.com/16012374/209158377-4f9e911d-990b-40cb-87fd-8bebb9e78584.png)

The 'full' blob is a bit higher than the 'line'. _Pink rectangulars added with my screenshot tool._

Finii:
:sob: 

WTF

![image](https://user-images.githubusercontent.com/16012374/209161381-d16e5a06-9d3f-48a4-a0b7-b6c243089ce6.png)

Not only are the 'line' things further down then the 'full' things (but at least both pairs are equal!),
no, the round-on-right LINE is as if it is about the same position as the FULL

but the round-on-left LINE is right aligned as the FULL, making it considerably more rightish that expected.
**It should be in the middle, like the other LINE thing**

Finii:
This was a tough one, but I think the solution is rather nice. It was about the 4th or 5th different approach. I guess the new code describes best what we want to achieve here.

Thanks for reporting!

[...]

PR #916, opened 2022-09-07
Fix unequal weather icon scale

Finii:
**[why]**
The weather icons have some symbols that have a different bounding box but should nevertheless be scaled alike, because for example one is the outer line of a thermometer and one is the matching stem.

**[how]**
Just add a scaleGlyph set.

Fixes: #915
Fixes: #658

#### Requirements / Checklist


Finii:
This is not enough, because while the scaling is kept identical within a group, the **shifting** is not.

This will still not match despite same scale factor:

image

The stem is too high to fit into the outer line.

Finii:
Rebase on master, force push

Finii:
Ahh, setting up the tests and then using `fontforge` _Compare fonts_, this is nice:

image

_Showing the degrees-icon moving up to were it belongs_

Finii:
Pffff, what a PR.

All tests seem to be okay and this is indeed doing what is intened ;-)

PR #833, opened 2022-05-20
Add icons for Emacs and Orgmode

ashfinal:
#### Description

Add icons for Emacs and Orgmode.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?

I used Inkscape to edit the SVG files, then imported them into Fontforge. In the meanwhile, I uploaded the SVG files to [IcoMoon](https://icomoon.io/app) to see if they triggered any warnings. And add them to `original-source.otf` as `e532` and `e533`. Also, I update `font-patcher` to include new icons in patched fonts.

#### What are the relevant tickets (if any)?

Closes #342 

Similar PR #429 

#### Screenshots (if appropriate or helpful)

Changes in `original-source.otf`

fontforge

Patched font:

kfontview

Nvim with patched font:

nvim_devicons

Let me know if I miss something.
Thanks!

PR #922, opened 2022-09-08
Update orgmode glyph

Finii:
**[why]**
The glyph is needlessly detailed (for a font).
There is another, more simple glyph, which is also used by fileicons.

More details see PR #833.

Reported-by: tecosaur <contact@tecosaur.net>

#### Requirements / Checklist


Finii:
The `original-source.otf` is now autogenerated, remove it from the PR. Force push.

PR #929, opened 2022-09-15
Add icon for hacklang

4e554c4c:
#### Description

_Please explain the changes you made here._

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?
I downloaded the Hack icon from [hacklang.org](https://hacklang.org/) whose source is on [github](https://github.com/hhvm/hacklang.org) and is licensed CC BY 4.0. Then edited it to be monochrome/font friendly in inkscape, added it to `original-source.otf` in fontforge as `U+E535`, and modified `font-patcher` and friends to support the changes.

#### What are the relevant tickets (if any)?

Similar PRs: #862 #833 #429

Closes:  #904

#### Screenshots (if appropriate or helpful)
fontforge
neovim

PR #931, opened 2022-09-17
Autogenerate original-source glyph font

Finii:
#### Description

**Add original-source.otf generator script**

**[why]**
When we add a custom glyph (or want to update Seti) the process is
rather laborious and we are needed to change the font and the
accompanying `i_seti.sh` in sync.

**[how]**
We use a data file to map icon (svg) filenames to codepoints and
readable names.

That file is parsed and the font and info file is created (overwritten
in the repo); and could then be easily committed. This can be a CI
workflow.

Having a dedicated mapping file (`icons.tsv`) enables us to have stable
codepoints for the same symbol over time. Changes in codepoint
allocation can be checked in git.

Having the font autogenerated help guarantee that the icons are all
likely scaled. We rescale them all to the same size and mid-position.
That is not needed for font-patcher, because it rescales and shifts
again based on to-be-patched font metrics. But it certainly is better
for a view into the original-source font.

Sizes and position are still roughly equivalent to the hand positioned
glyphs.

#### Requirements / Checklist

Possibility to generate `src/glyphs/original-source.otf` and `bin/scripts/lib/i_seti.sh` from a file that contains a list of codepoints and all the glyph files in `src/svgs/`.

Make backup of `original-source.otf` and `i_seti.sh`.
Run it (from the script dir). Check changed made to both aforementioned files.

First motivation:
https://github.com/ryanoasis/nerd-fonts/pull/833#issuecomment-1133552076

Came up again and again, for example https://github.com/ryanoasis/nerd-fonts/pull/929#issuecomment-1248497415
when we want to update complete Seti...

* #929 
* #922 

image


Finii:
This needs:
- CI integration
- A run to update the two files

Finii:
Ah, `codeclimate` :-}
It's always annoying, but afterwards it often does look better :->
And better `codeclimate` than ESlint ;-)

...fixing up

Finii:
Ah yes! :tada: How beautifully is that working?

First it looks at all `svg`s and simplifies them. That has been previously a manual task...

```
Checking for SVG simplifications
  Files might be MODIFIED by this run ('doit' specified)
Found 58 svgs in ../../src/svgs
Simplification for go.svg (19570 -> 5504) 28%
Simplification for config-folder.svg (14779 -> 3029) 20%
Simplification for grunt.svg (16596 -> 8178) 49%
Simplification for config.svg (14830 -> 3555) 23%
Simplification for octocat-folder.svg (8043 -> 2884) 35%
Simplification for react.svg (8004 -> 5748) 71%
Simplification for typescript.svg (3973 -> 2427) 61%
Simplification for go2.svg (25555 -> 6855) 26%
Simplification for bower.svg (6223 -> 4602) 73%
Simplification for c.svg (3758 -> 2384) 63%
Simplification for electron.svg (6024 -> 4212) 69%
Simplification for separator.svg (4938 -> 2115) 42%
Simplification for handlebars.svg (3101 -> 2097) 67%
Simplification for stylus.svg (4388 -> 2112) 48%
Simplification for coffeescript.svg (4146 -> 3297) 79%
Simplification for c++.svg (4338 -> 3254) 75%
Simplification for git-folder.svg (3107 -> 2471) 79%
Simplification for rss.svg (4227 -> 2175) 51%
Simplification for ms-dos.svg (8842 -> 5174) 58%
Simplification for license.svg (22305 -> 4062) 18%
Simplification for javascript.svg (4807 -> 2681) 55%
Simplification for npm-folder.svg (4477 -> 2738) 61%
Simplification for twig.svg (13214 -> 3455) 26%
Simplification for less.svg (8921 -> 3189) 35%
Simplification for php.svg (4790 -> 3346) 69%
Found 25 svgs to simplify
```

And as you can see, quite a number were not optimized :->

And then the font is spit out...:

```
[Nerd Fonts]  Glyph collection font generator 2.2.2

Read glyph data successfully with 65 entries (6 aliases)
Generating original-source.otf with 59 glyphs
Generating GlyphInfo i_seti.sh
Finished
```

And both of these changes are automatically added to the PR. :grinning:

Finii:
Here is what came out ...

The font got smaller, all SVGs got bigger? Yes, they are not optimized with `svgo` anymore, but that does optimize only the filesize of the `svg` and NOT the contents. And we do not care about the file size, we want small in number of nodes and paths, and do not care about `svg`-file-size. Because font file size depends only on number of nodes and paths.

image

And the actual font became even smaller. Without the optimization, created from the (un-simplified source svgs) the size was roughly 120k.

And even the `go` symbol looks ok now, which has been broken a long time with big ugly eyes...

image

Finii:
This should not have happened:

image

I just added some documentation, and a 'new' `original-source.otf` has been automagically committed.
It should only commit a new font, if the font is changed.

This is fixed by the following (force pushed) change:

Finii:
Obviously the fix did not ... fix it. :-(

Finii:
Found it:

image

It detects a difference, because we force 'both' timestamps to the previous font's creation time, but THAT font has different creation and modification time set. Creation time is the time when creation started, and mod time, when it finished, and when that hits a seconds boundary we have to different times. That is the reason why it does only happen sometimes.

Fixing...

PR #297, opened 2018-10-02
Add `space` glyph to Symbols fonts

Mange:
#### Description

Added the `space` glyph from DejaVu Sans Mono Regular into both Nerd
Font Complete symbol fonts and saved them as TTF.

I used fontforge and generated the font as a `TrueType` with default options.

This could, in theory, fix #275.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [x] Scripts execute without error (if necessary):
  - If any of the scripts were modified they have been tested and execute without error, e.g.:
    - `./font-patcher Inconsolata.otf --fontawesome --octicons --pomicons`
    - `./gotta-patch-em-all-font-patcher\!.sh Hermit`
- [x] Extended the README and documentation if necessary, e.g. You added a new font please update the table

#### What does this Pull Request (PR) do?

It adds the space character to the symbols fonts files.

#### How should this be manually tested?

On a Linux machine, check this output with the normal symbol fonts installed (and a fontconfig that automatically falls back to the symbols):

```bash
pango-view -t "Playing some  for you right now"
```

Screenshot showing rendering problem with the space after the symbol character

Then install this new version of the font, reset the cache and try again:

```bash
cp src/glyphs/Symbols-2048-em\ Nerd\ Font\ Complete.ttf ~/.local/share/fonts
fc-cache -fv
pango-view -t "Playing some  for you right now"
```

Screenshot showing that the rendering problem has now disappeared

#### Any background context you can provide?

Pango is one of the most common font rendering libraries on Linux. It's used by GTK/GNOME and a lot of standalone apps, like Rofi, Polybar, and a lot of terminals.

Pango looks for fallback fonts when the current font doesn't contain a given glyph, but it won't check spaces or any other very common "invisible" characters with the assumption that all fonts should have spaces in them.
This makes it so that after Pango switches to Nerd Fonts for a glyph, it won't switch back until it gets a non-space character after it, making it impossible to have spaces after a symbol.
I usually try to work around it by wrapping symbols in parenthesis, or adding a colon after them, but in some apps this isn't really possible to do anything about.

**Issues to look at:**

[...]

Issue #275, opened 2018-06-04
Include space in source / icons-only font


maximbaz:
#### 🗹 Requirements

- [x] A brief but descriptive _title_ of your issue
- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

#### 🎯 Subject of the issue

I admire this project, having one font with all icons is simply perfect. However what I'm missing and what I would like to request is a pure icons-only font, a font with no latin characters and other stuff, but only icons in private use area (maybe also space character `0020` is needed).

There are a couple of reasons to have such font. 

Firstly, the "original" fonts tend to use the same code points, and thus overlap — for example, both `Font Awesome` and `Material Design Icons` start with code `f000`, so it is impossible to install two fonts and use icons from both fonts in the same place.

Secondly, some apps do a very cool trick and allow selecting a special font only for a certain code points range — `kitty` terminal is one such example. The reason is that [you don't have to patch fonts anymore](https://github.com/kovidgoyal/kitty/blob/c9818e7bb1511625b952e7c1546ae724e27c90e1/kitty/kitty.conf#L39-L44), you can use your favorite coding font and an icons-only font simultaneously. It makes it super easy to change fonts and keep all the cool icons with no effort spent.

Lastly, I noticed you are not providing patched versions of some proprietary fonts due to licensing issues. This tends to be a burden for users of such fonts, as they need to re-patch their fonts with every new release. Having an icon-only font could be a much more pleasant way for these users to enjoy the power of Nerd Fonts.

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_ `N/A`
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_ `kitty`
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ `Arch Linux`

ryanoasis:
@maximbaz Appreciate your layout and description here.

[...]

PR #942, opened 2022-09-26
Add blank char to symbols only fonts

Finii:
**[why]**
It could be beneficial to have at least the 'blank' glyph defined, next to all the special glyphs (symbols).

This is needed by some (old) font rendering engines (Pango, IE11).

Fixes: #297

#### Requirements / Checklist

This is a retake on #297, which accommodates all the auto-generation stuff we do now.

#275


Finii:
Would this not break the width of 'blank' for people who use the symbols only font as font fallback? Then they would use the artificial blank provided here, instead of the original/text font's blank?

Finii:
I guess we close this. It's not really needed.

Issue #850, opened 2022-06-27
Patcher reduces the spacing between lines making fonts difficult to read.


BoltsJ:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
After patching a font, the line height is reduced which makes 

Expected behavior:
Line height should remain the same after patching.

Example symbols:
N/A

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - `Comic Code Regular Nerd Font Complete.otf`
   - Self-patched using the oci image.
- `gnome-terminal`, Blackbox, `gnome-text-editor`
- Fedora 36 Silverblue

#### ★ Screenshots (Optional)

Unpatched font: 
![image](https://user-images.githubusercontent.com/182554/176006195-c036f961-1855-4777-8516-2877835a49ef.png)

Patched font:
![image](https://user-images.githubusercontent.com/182554/176006340-522025b3-29a1-4cf0-929d-aaab6bac6a5b.png)

bartdorsey:
The same thing happens to me when patching comic code ligatures

Finii:
![image](https://user-images.githubusercontent.com/16012374/180990674-36c57058-2df0-42df-83b9-a7af4c8c34c5.png)

I guess that is very easy to fix. Believe I know where the size is lost.

DanBenHa:
> ![image](https://user-images.githubusercontent.com/16012374/180990674-36c57058-2df0-42df-83b9-a7af4c8c34c5.png)
> 
> I guess that is very easy to fix. Believe I know where the size is lost.

Could you share your insights?

tusharsnn:
I am facing the same issue, is there a known fix yet ?

Finii:
@tusharsnn :+1: 

I'm not sure the `font-patcher` code is correct in this regard anyhow.

:thinking:  Obviously I commented already some time back :grimacing: 

This needs some research in the `src/unpatched-fonts` base, which might take me some time.
Feel free to ping me from time to time :grimacing: 

[...]

PR #943, opened 2022-09-26
Draft: font-patcher: Respect gap values in source font

Finii:
**[why]**
We just ignore specified gaps in the source fonts (i.e. set them to zero). This reduces the line spacing in the patched font (because the gap is missing).

**[how]**
Distribute the gap INTO the cell, so that we can work with zero gap (we need that for the powerline glyphs), and still keeping the powerline glyphs centered about the regular glyphs AND keeping the total line spacing.

Idea-by: Tushar Singh <tusharvickey1999@gmail.com>

Fixes: #850

Reported-by: Joe Bolts

#### Requirements / Checklist

* #850

See Issue.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -738,11 +738,6 @@ class font_patcher:
             self.sourceFont.hhea_ascent = self.sourceFont.os2_winascent
             self.sourceFont.hhea_descent = -self.sourceFont.os2_windescent
 
-        # Line gap add extra space on the bottom of the line which
-        # doesn't allow the powerline glyphs to fill the entire line.
-        self.sourceFont.hhea_linegap = 0
-        self.sourceFont.os2_typolinegap = 0
-
     def get_essential_references(self):
         """Find glyphs that are needed for the basic glyphs"""
         # Sometimes basic glyphs are constructed from multiple other glyphs.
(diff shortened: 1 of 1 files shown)


tusharsnn:
I tried the script, font looks identical to the original font but now the text is centred as oppose to top/bottom aligned.

Also I didn't have to use the flag to achieve this, may be flag(--preserve-gap) hasn't been added yet ?

Finii:
> I tried the script, font looks identical to the original font but now the text is centred as oppose to top/bottom aligned.

What does that mean? Of course the glyphs will not touch the top of the 'cell' anymore if we put half of the gap 'above' the glyph.

> Also I didn't have to use the flag to achieve this, may be flag(--preserve-gap) hasn't been added yet ?

This is correct. I just threw in the code changes, with no (new) option.

tusharsnn:
> What does that mean? Of course the glyphs will not touch the top of the 'cell' anymore if we put half of the gap 'above' the glyph.

I'm sorry if I sounded confusing. I meant that it is now correctly aligned with line gaps as expected.

Finii:
Ok, so I will trudge on with the finer details of the PR, the 'main goal' seems fulfilled. Thank you.

DanBenHa:
Weird, still eats the gaps when I patch it:

Original:
image

Patched:
image

Finii:
What is your terminal?
But you `font-patcher` has the gap code from this PR?
Maybe I should put "my" patched font somewhere so we all talk about the same font file.

DanBenHa:
> What is your terminal? But you `font-patcher` has the gap code from this PR? Maybe I should put "my" patched font somewhere so we all talk about the same font file.

That's alacritty. Yes, I checked out the feature branch and verified the code is there. Yes, if you could provide your test font that would be helpful as a reference.

Finii:
Gap comparison...

Orange numbers is your screenshot image from above.
Yellow numbers is my screenshot, size is the same and number 113 aligned.


Peek 2022-09-29 09-30

Do you have a different `alacritty`?

tusharsnn:
Perhaps alacritty seems to modify font properties and provides it own 'settings' to customize them ?

```yml
font:
  # Normal (roman) font face
  normal:
    family: Noto Mono for Powerline
    # The `style` can be specified to pick a specific face.
    #style: Regular

  # Offset is the extra space around each character. `offset.y` can be thought of
  # as modifying the line spacing, and `offset.x` as modifying the letter spacing.
  offset:
    x: 0
    y: 0
```
*source: https://gist.github.com/Cogitri/e8d5c63818443f3c8f13cd7760fe77aa*


Take a look at:
1) alacritty/alacritty#189
2) https://superuser.com/questions/1619887/how-to-add-padding-change-space-between-lines-in-alacritty-terminal

Finii:
Just compared with `kitty`, which renders it different but same :grimacing: 
Especially it does not 'invent' a **bold** face when the font does not have one

Peek 2022-09-29 09-51

_Shown my `alacritty` and `kitty`_

Finii:
Hmm, obviously it IS possible to nudge `alacritty` into the right direction

image

Lets make this more sane.

Finii:
Well, for ME the current code **is** working with `alacritty`:

Peek 2022-09-29 10-50
_Original and Nerd Fonts version alternatring_

The gap is displayed correctly by `alacritty`. The patching relocates 1/2 of the gap to the top and 1/2 to the bottom.

The thing is just, that `alacritty` shows the gap above lines, not below. Which is what I would have expected, but that does not matter in fact. (With 'original' the `y` touches the cell bottom, so the gap is in the top of the cell. Also note that the `f` is in fact as high as the cell, gap ignored - 'original' shows the (full) gap above the `f`)

But what you can clearly see, the line spacing is not touched.

Here the intermediate steps...

Original -> Gap removed -> Gap redistributed top/bottom -> adding symbols -> Original -> ...

Peek 2022-09-29 10-58

Issue #947, opened 2022-09-29
Something wrong with 3270 letter "small m"?


arkenoi:
#### 🔧 Your Setup

Affects the whole 3270 font family, and rendering is the same in all applications.

I use Linux (Fedora 36 template under Qubes OS 4.1)

#### ★ Screenshots (Optional)

![2022-09-29-121040](https://user-images.githubusercontent.com/6000406/192992126-e74f9af8-b3d5-4bed-8317-d591073db5b7.png)

arkenoi:
Just cross-checked if non-gnome apps are affected. Yes, they are.

Finii:
Thanks for the report. Checking now..

Finii:
Hmm, only `otf` is affected?!

![image](https://user-images.githubusercontent.com/16012374/192999991-e2c6082f-43aa-4da8-8167-0fe2e0b50d11.png)

Finii:
WAT?!

![image](https://user-images.githubusercontent.com/16012374/193000699-8253af69-50ec-43e5-85c0-664847b6caa7.png)

Finii:
This seems to be a `fontforge` problem:

![image](https://user-images.githubusercontent.com/16012374/193003256-56224669-983e-478b-96cc-1b77f3312088.png)

Happens with March 2022 release, but not with March 2020 release (of `fontforge`). Hmm.

Finii:
All the `ff` code around the error message is old. This will take some time to fix in `ff`:

![image](https://user-images.githubusercontent.com/16012374/193003973-e1542f54-1eb4-4ed0-96c9-eaf049200d3e.png)

Finii:
The corruption already occurs when merely opening the file.

Finii:
Bisected to 

![image](https://user-images.githubusercontent.com/16012374/193017345-390df3b8-403b-4df5-b9d0-6453d28a7a88.png)

This PR https://github.com/fontforge/fontforge/pull/4685

Finii:
We can fix this _for us_ by just updating the font. The current version has no problems as far as I can see.
In this way I can avoid to debug `fontforge` :grimacing: 

The patch results can be downloaded here (for trials):
https://github.com/Finii/nerd-fonts/tree/feature/update-3270font-patched/patched-fonts/3270

PR #948, opened 2022-09-29
Update 3270font to v3.0.1

Finii:
#### Description

**[why]**
We have only a rather old 3270font: v1.2.23

This new font can be processed with current FontForge without problems. And we have the benefit of an update.

**[how]**
Because the paths change (Medium->Regular etc) we need to purge the patched-fonts/ directories manually. The automatic purging does work only if at least the paths are kept.

**[note]**
https://github.com/rbanffy/3270font

Fixes: #947

#### Requirements / Checklist

Update 3270 to latest version, thereby fixing `fontforge` problem (see Issue).

Have a look on the produced 3270 fonts, both ttf and otf.

See Fontforge problem mentioned in https://github.com/ryanoasis/nerd-fonts/issues/947#issuecomment-1262154567


Finii:
The patch-reults can be examined here:
https://github.com/Finii/nerd-fonts/tree/feature/update-3270font-patched/patched-fonts/3270

Finii:
This needs a CI `release` run afterwards to prevent empty in-repo directories for this font patched.
It does not add the (newly) patched fonts in itself.

PR #958, opened 2022-10-09
Direct GitHub link

mrienstra:
... Unless you're expecting more topics to match in the future?

Old:
https://github.com/search?q=nerd+fonts&type=Topics

New:
https://github.com/topics/nerd-fonts

#### Requirements / Checklist

---

PS: Not particularly a fan of having the website live in a branch, mostly just because it means it can't be searched via the GitHub website.


Finii:
This definitely looks better, thank you!

> Not particularly a fan of having the website live in a branch

Any suggestions? The handling is always a but awkward, but I can live with it.
And then, there would be so much to do, and there is only so much time in a day...

Finii:
@all-contributors please add @mrienstra  for documentation

Issue #512, opened 2020-08-11
Can't patch font "Cascadia Code PL".


rediffusion:
Aloha! 🙂

### 🔧 Your Setup

«Windows 10 2004»
Windows Terminal > WSL > Ubuntu-20.04

**Manual** has taken from Youtube tutorial "[Scott Hanselman](https://youtu.be/oHhiMf_6exY)".

---

### apt install python3-pip

# OUTPUT:

> Reading package lists... Done
> Building dependency tree
> Reading state information... Done
> python3-pip is already the newest version (20.0.2-5ubuntu1).
> 0 upgraded, 0 newly installed, 0 to remove and 2 not upgraded.

### sudo apt-get install software-properties-common

# OUTPUT:

> Reading package lists... Done
> Building dependency tree
> Reading state information... Done
> software-properties-common is already the newest version (0.98.9.1).
> 0 upgraded, 0 newly installed, 0 to remove and 2 not upgraded.

### sudo add-apt-repository ppa:fontforge/fontforge

# OUTPUT:

>  More info: https://launchpad.net/~fontforge/+archive/ubuntu/fontforge
> Press [ENTER] to continue or Ctrl-c to cancel adding it.
> 
> Ign:1 http://ppa.launchpad.net/fontforge/fontforge/ubuntu focal InRelease
> Err:2 http://ppa.launchpad.net/fontforge/fontforge/ubuntu focal Release
>   404  Not Found [IP: 91.189.95.83 80]
> Hit:3 http://security.ubuntu.com/ubuntu focal-security InRelease
> Hit:4 http://archive.ubuntu.com/ubuntu focal InRelease
> Hit:5 http://archive.ubuntu.com/ubuntu focal-updates InRelease
> Hit:6 http://archive.ubuntu.com/ubuntu focal-backports InRelease
> Reading package lists... Done
> E: The repository 'http://ppa.launchpad.net/fontforge/fontforge/ubuntu focal Release' does not have a Release file.
> N: Updating from such a repository can't be done securely, and is therefore disabled by default.
> N: See apt-secure(8) manpage for repository creation and user configuration details.

### sudo apt-get update

# OUTPUT:

> More info: https://launchpad.net/~fontforge/+archive/ubuntu/fontforge
> Press [ENTER] to continue or Ctrl-c to cancel adding it.
> 
> Ign:1 http://ppa.launchpad.net/fontforge/fontforge/ubuntu focal InRelease
[...]

PR #960, opened 2022-10-10
font-patcher: Warn if sourcefont is a VF

Finii:
**[why]**
fontforge is not really able to work with OpenType variable fonts, at least not with all. Some support is available as MM, for older formats. But anyhow we do not really create a patched variable font but a fixed font.

People might ignore all the errors Fontforge throws on opening, so an explicit message might be in order.

[how]
It is not possible to detect a VF input font with current fontforge reliably. Instead we search for the existence of one of the tables that are  needed for a variable font. We can not rely on `STAT`, because that might be also used in fixed fonts.

Fixes: 512

#### Requirements / Checklist

Just add a warning when trying to patch a VF.

Patch a normal (static) font, patch a VF font.

#512

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.3.1"
+script_version = "3.3.2"
 
 version = "2.3.0-RC"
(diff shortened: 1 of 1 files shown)


Finii:
image

Fixed in #961

PR #967, opened 2022-10-14
Draft: Scale powerline glyphs always

Finii:
#### Description

After this PR the powerline glyphs (well, some of them) will always be scaled down to one cell, as if `-s` (`--mono`) has been supplied. This is only relevant for the non-Mono font variants of course.

#### Requirements / Checklist

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -246,6 +246,7 @@ class font_patcher:
         self.setup_version()
         self.get_essential_references()
         self.setup_name_backup(font)
+        self.font_monospaced = is_monospaced(self.sourceFont)
         if self.args.single:
             self.assert_monospace()
         self.remove_ligatures()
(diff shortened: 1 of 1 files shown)


Finii:
This shows the current state for `Nerd Font` (i.e. non-Mono), `Inconsolata` shown:

Peek 2022-10-14 17-25

The 'cell' is the rectangle in the middle. The glyphs are either right or left aligned to the cell border, with some small overlap.
The glyphs are scaled vertically to match the font's line spacing.

Horizontal spacing is untouched. That means that
* a) The glyphs get distorted (aspect ratio changed) (which might only interest for 'Lego' or 'Squares' glyphs)
* b) The length in the non-aligned direction is not controlled

Some of the right aligned glyphs, like flames, extend rather far into previous 'territory'.
Already the inverse-D-shape extends into the previous char. This of course depends on the concrete patched font.

Finii:
Hmm, #661

Finii:
How does this relate to
* #984

Finii:
This is obsoloete / is now part of
* #748

Issue #976, opened 2022-10-20
Ligatures broken


Finii:
Nerd Font used to have a 'ligature removal' feature for the `Nerd Font Mono` fonts.
This seems to be broken with recent fonts (i.e. `v2.2.2` seems ok while `HEAD` is broken).
See also
* #934 
* #973 
* #975 

The reason is the recent fix of the ligature removal code (i.e. https://github.com/ryanoasis/nerd-fonts/commit/5d218b1cd305470093734e32fbd3d5fb827ede6f).

History data:

* Ligature removal introduced 02-2018 (`v2.0.0`) 9917e2db
* Ligature removal broken 08-2018 (`v2.1.0`) https://github.com/ryanoasis/nerd-fonts/commit/0a480bbc8d03d226f33e588bc6c0d001571389bc
* Ligature removal fixed 09-2022 (`v2.3.0-RC`) https://github.com/ryanoasis/nerd-fonts/commit/5d218b1cd305470093734e32fbd3d5fb827ede6f

Fonts with ligature table data:

* Arimo (*)
* Cousine                 
* DejaVuSansMono
* FiraCode            
* FiraMono
* Hasklig                  
* Iosevka           
* Lekton
* Monoid                    
* Noto               
* OpenDyslexic (*)
* Overpass
* ShareTechMono      
* SpaceMono          
* Ubuntu (*)

The tables have been added with a2f56206e and never been updated (which would generally be needed if the source font is updated) (except Hasklug 12cd703e3).

Affected (very quick check) is only FiraCode and Hasklig.

Updated after introduction of the ligature removal are

* FiraCode 78be0e1e7 d23d0e008 338497d1e 2d32a1f44 dfc31db1d d9ea69ac6 ad162ace7
* Hasklig e2745c3e4 ca44d3873 6edb95723
* Iosevka a7d8f517e 052ece39a 8738eb476 09c4776ee 6edb95723

Of the three Iosevka's ligature removal seems to have survived the updates, but maybe not.

Issue #975, opened 2022-10-19
Fira Code Nerd Font Mono is not displaying == sign correctly


Goku-San:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
**==** sign not displaying properly with **FiraCode Nerd Font Mono Medium**.  Few days ago ( Saturday or Sunday ) I updated the Fira Code Nerd Font. Before that everything was displaying correctly.

Expected behavior:
Expect to display properly as with original Fira Code Medium or FiraCode Nerd Font Medium.

If I use **FiraCode Nerd Font Medium** or **Fira Code Medium** than icons in vim are clipped.
So if you can please fix the **==** sign or revert to the patch from 9 days ago, when all was good.

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_ 
       Fira Code Medium Nerd Font Complete Mono.ttf

   - _Where did you get the file from (download link, self patched, source downloaded from link...)_ 
       https://github.com/ryanoasis/nerd-fonts/tree/master/patched-fonts/FiraCode

- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
    **st** terminal with ligatures patch 
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
    Linux - Manjaro XFCE

#### ★ Screenshots (Optional)

![st_correct](https://user-images.githubusercontent.com/22682262/196760515-ed8ab056-ec4c-4902-ac3b-0b126fd57ebf.png)

st call inside config.h 
``static char *font = "FiraCode Nerd Font:style=Medium:size=12:antialias=true:autohint=true";``

![st_incorrect](https://user-images.githubusercontent.com/22682262/196760682-a6f010f0-ff6a-489e-84cc-7b8bac9a8829.png)

st call inside config.h
``static char *font = "FiraCode Nerd Font Mono:style=Medium:size=12:antialias=true:autohint=true";``

[...]

Issue #973, opened 2022-10-18
Hasklug: additional ligature issues


akosradler:
#### 🎯 Subject of the issue

Experienced behavior:
Turning ligatures on for Hasklug deletes the first character from a two char ligature:
The two issues I've found:
- ```\\``` becomes ``` \```
- ```==``` becomes ``` =```
- ```>>``` becomes ```*>```

🔧 Your Setup
Tried with Hasklug Nerd Font Complete Mono Windows Compatible.otf
Terminal agnostic
Using Windows

A similar issue has been solved on this ticket: https://github.com/ryanoasis/nerd-fonts/issues/934 (that one works fine now)

Finii:
I'm tempted to remove to ligature removal feature altogether. Other Nerd Fonts fonts also have ligatures, and people who do not want them can usually turn them off in their client. If the client can not turn it off, it usually does not see them anyhow :grimacing: 

That feature requires tedious manual work, on every update :roll_eyes:

akosradler:
Not sure if this helps, but Hasklig is essentially just Source Code Pro with ligatures. At least in this case, if someone does not want them, they can just switch to SCP.

PR #977, opened 2022-10-20
Sunset ligature removal

Finii:
#### Description

**[why]**
Keeping the ligature removal tables up to date with source font updates is quite a maintenance burden.

Usually if a terminal supports ligatures at all it has an option to disable them.

Present ligatures do not prevent monospaced fonts and monospaced detection.

It was only present in `v2.0.0`, and the code has been broken with `v2.1.0`.

**[note]**
Updated example ligature removal table to the one for Iosevka. Left stubs in relevant code passages to facilitate finding this commit.

Fixes: #976
Fixes: #975
Fixes: #973

#### Requirements / Checklist

Remove the ligature removal tables, so that all patched fonts still contain all ligatures that they originally had.


Finii:
See also #1191

Issue #1005, opened 2022-11-26
Cannot patch nerd font Meslo LGS with codicons


hgoumner:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

I cannot patch the nerd font Meslo LGS Regular with the codicons package due to the following error:

`  File "/home/hristo/Documents/nerd-fonts/font-patcher.py", line 1427, in <module>
    main()
  File "/home/hristo/Documents/nerd-fonts/font-patcher.py", line 1407, in main
    patcher.patch(sourceFonts[-1])
  File "/home/hristo/Documents/nerd-fonts/font-patcher.py", line 302, in patch
    self.copy_glyphs(SrcStart, symfont, patch['SymStart'], patch['SymEnd'], patch['Exact'], patch['ScaleGlyph'], patch['Name'], patch['Attributes'])
  File "/home/hristo/Documents/nerd-fonts/font-patcher.py", line 956, in copy_glyphs
    currentSourceFontGlyph = min(possible_codes)
                             ^^^^^^^^^^^^^^^^^^^
ValueError: min() arg is an empty sequence
`

This is the command i use: `fontforge -script font-patcher.py MesloLGS\ NF\ Regular.ttf --custom codicon.ttf`

And this is the structure of the directory as explained [here](https://github.com/mortepau/codicons.nvim#how-to-patch-fonts)

![image](https://user-images.githubusercontent.com/11947421/204089950-124b78f5-c3e6-4185-a3af-a7373aa43210.png)

Expected behavior:

I expect to have a patched font being written in the directory.

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - Meslo LGS NF Regular.ttf
   - _Where did you get the file from (download link, self patched, source downloaded from [link...)_
[...]

PR #1006, opened 2022-11-26
font-patcher: Fix crash on some custom sets

Finii:
#### Description

**[why]**
When a custom set is to be patched and that set contains the `.notdef` glyph the script crashes.

**[how]**
The glyphs has an unicode codepoint of -1, a code that we explicitely not patch. That leads to confusion when we try to determine the codepoint to be used for that glyph, because it has none.

Exclude negative codepoint glyphs when we determine the selection of to-be-patched glyphs.

As last resort for broken custom sets skip over glyphs that come out of order (which should be impossible).

Fixes: #1005

#### Requirements / Checklist

Allow patching with custom sets that contain the `notdef` glyph.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.2.0"
+script_version = "3.2.1"
 
 version = "2.3.0-RC"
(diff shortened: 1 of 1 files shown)


Finii:
Forgot to increase the `font-patcher` internal version number, which is important for docker containers...
Force push.

Issue #1018, opened 2022-12-02
Bold arrow is off in Iosevka Nerd Font Mono Regular


uloco:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
When using bold text, the arrow is not aligned correctly horizontally (it's too close to the right number here)
It is also bigger and looks generally out of place.
(This also happens with _Medium_, _Oblique_, _Thin_ and _Light_ fonts)

Expected behavior:
Bold text should align correctly

Example symbols:
⇣1⇡1

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_
     - Iosevka Nerd Font Complete Mono.ttf
   - _Where did you get the file from (download link, self patched, source downloaded from link...)_
     - installed with homebrew like so `brew tap homebrew/cask-fonts && brew install --cask font-iosevka-nerd-font`
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
  - iterm2
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
  - macOS Ventura 13.0

#### ★ Screenshots (Optional)

Regular (works)  
<img width="47" alt="Screenshot 2022-12-02 at 11 48 29" src="https://user-images.githubusercontent.com/8818340/205276333-822c9f72-ad4a-4c87-9184-a53ec51fb368.png">

Bold (broken)  
<img width="47" alt="Screenshot 2022-12-02 at 11 49 41" src="https://user-images.githubusercontent.com/8818340/205276400-3f2c29e2-582f-4b7e-985f-f9619e2a8432.png">

Finii:
The arrow seem to be `21E1` and `21E3`:

![image](https://user-images.githubusercontent.com/16012374/205484920-3c40d697-dc55-4b09-a158-df1306457070.png)

These glyphs are not patched in by Nerd Fonts.

Opening the two fonts, they look both ok, each is 1000 wide:

[...]

PR #1019, opened 2022-12-04
Fix Iosevka Family names

Finii:
#### Description

**[why]**
The Iosevka font has a lot of different families. A lot users install just all "Iosevka Nerd Font" families, and this can break in a lot different ways.

I will try to collect Issues possibly caused by this in PR #1019.

**[how]**
Just turn the feature on in `font-patcher` (via patch-em-all's config).

#### Requirements / Checklist

Enable `--makegroups` for Iosevka on release runs.

For example in Gnome Terminal, this is how all the Iosevka Families are shown at the moment:

image

And this is after this PR has been pulled:

image

Note that the `Nerd Font` part is missing in the 'now' screenshot :-o This not even breaks all the patched Iosevkas, because they mix up all families, but also interfere with parallel installed Iosevka-original :-(

### DIFF
--- a/src/unpatched-fonts/Iosevka/config.cfg
+++ b/src/unpatched-fonts/Iosevka/config.cfg
@@ -1,1 +1,2 @@
 config_has_powerline=1
+config_patch_flags="--makegroups"


Finii:
Possible phenotypes of the Family name issue:
* #1018

Issue #1040, opened 2023-01-09
No bleeding edge FontPatcher.zip


Finii:
> in order to avoid the 30 GB of monorepo history and whatnot from here

Well, in the meantime we have a single zip file that holds the last release script with all additional files needed for patching:
https://github.com/ryanoasis/nerd-fonts/releases/download/v2.3.0-RC/FontPatcher.zip

But well, that is not `master`. And it is very badly documented.
We need to
* promote that more (who wants to write docs?)
* install some CI that keeps one version of that zip as 'rolling release' on `master`

_Originally posted by @Finii in https://github.com/ryanoasis/nerd-fonts/issues/595#issuecomment-1375984265_

b-:
I mentioned this in the PR, but it's marked as a draft because the step performing the actual release is most certainly not correct for this repository. I'd like to ask for help with that part, but I'm pretty sure the zip file this spits out is correct.

It's a little bit rough in that I'm checking out the unpatched fonts and then deleting them, rather than being more specific about what to checkout, but I don't know if this is such a big deal. At the very least I'm not checking out the whole huge repo and its eternal(ly large) history

b-:
Now that I think about it, it probably would make sense to have the action simply add the zip to the repo and push that. I think that's what you're suggesting, @Finii ?

I'm not certain how to add to such a bare repo, though. Maybe we can just `cp fontpatcher.zip nerd-fonts/fontpatcher.zip && git add fontpatcher.zip && git commit -m"Update fontpatcher.zip" && git push`?

Finii:
Yes.

Thanks for the RP.
I'm almost in bed already, just some fast rough notes...

You should use actions instead of writing your own git commands. Like https://github.com/actions/checkout
See how the already existing workflows do it.

There is already a script that creates the release zip file, why not use that for inter-release zips?
https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/archive-font-patcher.sh

Of course there are also actions to commit a new file to the repo. Like, see
https://github.com/ryanoasis/nerd-fonts/blob/master/.github/workflows/packsvgs.yml

One would need to think about the trigger more closely.

If you want to test the workflow I would recommend forking this repo and doing the experiments in your fork.

b-:
Thanks for the notes! I'll work on them.

> You should use actions instead of writing your own git commands. Like https://github.com/actions/checkout See how the already existing workflows do it.

[...]

PR #1041, opened 2023-01-09
Release master fontpatcher

b-:
#### Description

Fixes https://github.com/ryanoasis/nerd-fonts/issues/1040

@all-contributors please add @b- for infra

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/-/contributing.md)
- [ ] Verified the license of any newly added font, glyph, or glyph set _(N/A)_

#### What does this Pull Request (PR) do?

 - Adds a Github Actions workflow to release _just_ an updated fontpatcher.zip from master

PR #1044, opened 2023-01-10
Release master fontpatcher

b-:
#### Description

Fixes https://github.com/ryanoasis/nerd-fonts/issues/1040

@all-contributors please add @b- for infra

#### Requirements / Checklist

- [ ] Verified the license of any newly added font, glyph, or glyph set _(N/A)_

 - Adds a Github Actions workflow to release _just_ an updated fontpatcher.zip from master, upon push of font-patcher or its dependencies or glyphs to master 

 - make sure the freshly cooked font-patcher.zip contains all required dependencies (but not anything else), and functions as expected.

https://github.com/ryanoasis/nerd-fonts/pull/1041

https://github.com/ryanoasis/nerd-fonts/issues/1040

### DIFF
--- a/.github/workflows/release_master_fontpatcher-zip.yaml
+++ b/.github/workflows/release_master_fontpatcher-zip.yaml
@@ -3,6 +3,7 @@ name: Release font patcher from master
 on:
   create:
   push:
+    branches: master
     paths:
       - font-patcher
       - src/glyphs/**
--- a/.github/workflows/release_master_fontpatcher-zip.yaml
+++ b/.github/workflows/release_master_fontpatcher-zip.yaml
@@ -3,7 +3,7 @@ name: Release font patcher from master
 on:
   create:
   push:
-    branches: master
+    #branches: master
     paths:
       - font-patcher
       - src/glyphs/**
@@ -17,26 +17,41 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    
+    outputs:
+      changed: ${{ steps.updated-or-not.outputs.updated }}
     steps:
-    
+    - name: install zipcmp
+      run: sudo apt-get install -y zipcmp
     - name: Grab the script and its dependencies from the repo
       uses: snow-actions/sparse-checkout@v1.2.0
       with:
         patterns: |
+          FontPatcher.zip
           font-patcher
           src/glyphs
           src/archive-font-patcher-readme.md
           bin/scripts/archive-font-patcher.sh
           bin/scripts/name_parser
-          
     - run: chmod +x 
         font-patcher
         bin/scripts/archive-font-patcher.sh
         bin/scripts/name_parser        
     - run: bin/scripts/archive-font-patcher.sh
+    - uses: actions/upload-artifact@v2
+      with:
+        name: FontPatcher.zip
+        path: archives/FontPatcher.zip
+    - id: updated-or-not
+      run: |
+        if zipcmp archives/FontPatcher.zip ./FontPatcher.zip ; then
+          echo "updated=false" | tee -a $GITHUB_OUTPUT
+        else
+          echo "updated=true" | tee -a $GITHUB_OUTPUT
+        fi
     - run: cp archives/FontPatcher.zip . -f
     - uses: EndBug/add-and-commit@v9
+      if: steps.updated-or-not.outputs.updated == 'true'
+      #needs to be quoted
       with:
         fetch: false
         add: "FontPatcher.zip"
--- a/.github/workflows/release_master_fontpatcher-zip.yaml
+++ b/.github/workflows/release_master_fontpatcher-zip.yaml
@@ -3,6 +3,13 @@ name: Release font patcher from master
 on:
   create:
   push:
+    paths:
+      - font-patcher
+      - src/glyphs/**
+      - src/archive-font-patcher-readme.me
+      - bin/scripts/archive-font-patcher.sh
+      - bin/scripts/name_parser
+      - .github/workflows/release_master_fontpatcher-zip.yaml
   workflow_dispatch:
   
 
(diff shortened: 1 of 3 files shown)


b-:
Round 2, with less spam

.github/workflows/release_master_fontpatcher-zip.yaml
```diff
@@ -0,0 +1,45 @@
+name: Release font patcher from master
+
+on:
+  create:
```
Finii, line 4: Not sure why we want to trigger on `create`?

.github/workflows/release_master_fontpatcher-zip.yaml
```diff
@@ -0,0 +1,45 @@
...
+  
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    
+    steps:
+    
+    - name: Grab the script and its dependencies from the repo
+      uses: snow-actions/sparse-checkout@v1.2.0
```
Finii, line 24: In other workflows with equivalent step we use `Bhacaz/checkout-files@v2`.
Maybe use that here also or also switch the other ones?
Can you give details? The less actions we use the easier are updates ;)

I guess this one uses patterns while the other one needs concrete files (but I did not check).

.github/workflows/release_master_fontpatcher-zip.yaml
```diff
@@ -0,0 +1,45 @@
...
+          font-patcher
+          src/glyphs
+          src/archive-font-patcher-readme.md
+          bin/scripts/archive-font-patcher.sh
+          bin/scripts/name_parser
+          
+    - run: chmod +x 
+        font-patcher
+        bin/scripts/archive-font-patcher.sh
+        bin/scripts/name_parser
```
Finii, line 36: I guess the action does not restore the access rights?

If you want to `+x` the scripts in `name_parse` is not some `-R` option or a glob (`*`) needed for `chmod`?

.github/workflows/release_master_fontpatcher-zip.yaml
```diff
@@ -0,0 +1,45 @@
...
+        font-patcher
+        bin/scripts/archive-font-patcher.sh
+        bin/scripts/name_parser        
+    - run: bin/scripts/archive-font-patcher.sh
+    - run: cp archives/FontPatcher.zip . -f
+    - uses: EndBug/add-and-commit@v9
+      with:
+        fetch: false
+        add: "FontPatcher.zip"
+        message: "[ci] update FontPatcher.zip"
```
Finii, line 43: We always start capitalized (i.e. `s/update/Update/`), but I can also change that.

.github/workflows/release_master_fontpatcher-zip.yaml
```diff
@@ -0,0 +1,45 @@
+name: Release font patcher from master
```
Finii, line 1: Maybe we can find a better name here (also for the yaml file).

Esp this is not a 'release' but some snapshot.
Maybe `Create font-patcher snapshot` and `font-patcher-snap.yml`? Hmm :thinking: 
Maybe `Create font-patcher archive` and `font-patcher-zip.yml`?

All other workflows have the `.yml` suffix, I think we should stick with that.

Finii (changes requested):
This looks very good!

I have some questions and suggestions, see below.

And then, I would not put `FontPatcher.zip` into the PR, because it will be outdated anyhow when the PR is pulled. It can be added after pulling by triggering the workflow.

Another thing, maybe we should do a `zipcmp` after creating the zip file to see if the contents of the zip _really_ changed, or just the timestamps differ. At the moment someone could push 'run this workflow' multiple times and we get a new commit always.

.github/workflows/release_master_fontpatcher-zip.yaml
```diff
@@ -0,0 +1,45 @@
...
+
+on:
+  create:
+  push:
+    branches: master
+    paths:
+      - font-patcher
+      - src/glyphs/**
```
Finii, line 9: If someone pushes a fixed `svg` to `src/svgs` the `packsvgs.yml` will run and create a new `original-source.ttf` and push it. But I guess that will NOT trigger this workflow unless one adds something somewhere. I have no clue what and where. Just that there are some ... hurdles if workflow commits shall trigger workflows :thinking: 

It is ok how it is now, but maybe you know the answer :-)

Finii: Also after a release.yml run, I guess the zip inside the repo will not be updated although the version in the script has been changed (by the release workflow) :thinking:

b-: > If someone pushes a fixed `svg` to `src/svgs` the `packsvgs.yml` will run and create a new `original-source.ttf` and push it. But I guess that will NOT trigger this workflow unless one adds something somewhere. I have no clue what and where. Just that there are some ... hurdles if workflow commits shall trigger workflows 🤔
> 
> It is ok how it is now, but maybe you know the answer :-)

https://github.com/b-/Iosevka/blob/cc3babd8033f64086341820cfa014d026a14021f/.github/workflows/build.yml#L45

I don’t remember if this step actually functioned or not — I might have been missing something when I set it up. But the idea here is to set an API PAT (Personal Auth. Token) for the repository you want to trigger workflows _on_, add it as a secret to the repo you want to trigger workflows _from_, and then just call the second workflow from the first. 

I’ve done some research, and I don’t think it’s possible to trigger one workflow from another without a custom PAT. Actions Workflows get a token by default anyway, but it specifically has the ability to trigger other workflows disabled — that’s the issue there

.github/workflows/release_master_fontpatcher-zip.yaml
b-: I think this was a mistake on my part for something else — I have to double check, but I think a different action I used previously to clone the repo used tarballs under the hood and didn’t set permissions properly.

We should be able to just remove the line and not have a problem:

This checkout is just a real sparse checkout (it’s great, I _love_ sparse checkout, it’s the only way to wrangle big repos like https://GitHub.com/QMK/qmk) so it shouldn’t mess with the permissions.

.github/workflows/release_master_fontpatcher-zip.yaml
b-: See my below conversation: the action you use to checkout in the other actions, at least I think that’s the checkout action I am thinking of, messed with/reset permissions by grabbing files using GitHub’s HTTPS api. This one uses good ol’ git under the hood. 

I think it would be less “dirty” to use a `git` (the CLI)-based action like this whenever possible (ie., maybe indeed switch other workflows to it?), and you can see in my action runs it still is pretty fast for this big clunker of a repo

That all was the reason I picked another action, anyway — you (understandably) didn’t like the manual git command twiddling I’d started with, but I really didn’t want to clone even the entire worktree when we are just using a small section of it.

.github/workflows/release_master_fontpatcher-zip.yaml
b-: That was for my own testing, forgive me.

.github/workflows/release_master_fontpatcher-zip.yaml
b-: Of the two you suggested, I like

> Maybe Create font-patcher archive and font-patcher-zip.yml?

more. “Snap” gives me these uncomfortably weird Ubuntu tinglies

Finii:
@b- 
Hej :-)
You are still on this? At least the `zip` file is still in the PR :thinking:

b-:
Oh jeez, I wanted to run it again because I wanted to patch a font, and I honestly forgot why it wasn't merged. Let me remove the zip right now!

b-:
> This looks very good!
> 
> I have some questions and suggestions, see below.
> 
> And then, I would not put `FontPatcher.zip` into the PR, because it will be outdated anyhow when the PR is pulled. It can be added after pulling by triggering the workflow.
> 
> Another thing, maybe we should do a `zipcmp` after creating the zip file to see if the contents of the zip _really_ changed, or just the timestamps differ. At the moment someone could push 'run this workflow' multiple times and we get a new commit always.

I agree a zipcmp would be a good idea. I just removed the FontPatcher.zip from the repo, but I want to use a separate branch to figure out the zipcmp part.

b-:
Okay, we got a zipcmp in there now!

I put the conditional on the step level, but I'm not sure if it would be preferable to have a separate job with a step to do the release. Gut feeling suggests to me that it would be, especially what with the pretty pipeline graphics that you see when you have multiple jobs in an actions pipeline. But I had some trouble with the outputs/conditional on the job level, and this definitely works just fine as is.

I also am a little on the fence because using separate jobs means that the runner has a clean slate on the subsequent job, which means duplicating the checkout step and adding artifact upload and download steps.

Finii:
* Rebase on master
* Sqaush 'testing' commits
* Force push

Finii:
I changed the workflow a bit, hope you do not mind.
Tested the stuff and it looks good.
Thanks for the PR, having the archive for 'master' is really a good addition!

Finii:
Hmm, somehow the rebase on master was lost.
rebase on master, force push

Finii:
@all-contributors add @b- for infrastructure

Finii:
Grr, with a shallow checkout **of course** `git describe` does not work :grimacing: 

image

Fixing...

Finii:
Fixed via afa6abe0d9e2289c49db94008373323dd6941256

Finii:
Just wanted to say that I'm always so happy this is exisiting. When I merge something and I see this:

image

Thank you again :-)

b-:
That made my night to read, thank you!

Issue #1043, opened 2023-01-10
OverpassMono isn't spacing lines


marshallishere:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

The OverpassMono doesn't seems to have the spacement that Overpass Mono normally has.

Expected behavior:

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - Overpass Mono Regular Nerd Font Complete Mono.otf
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - The oficial website
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - [foot](https://codeberg.org/dnkl/foot)
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Fedora Workstation 37
#### ★ Screenshots (Optional)

![Captura de tela de 2023-01-10 11-53-07](https://user-images.githubusercontent.com/110747746/211586192-70602b3b-b110-4fc7-8c17-5bdf85209000.png)

Finii:
Just from the look of it I assume the difference comes from the fact that the original has a `linegap` of 266 (open space without anything between lines).

This should have been fixed by #943.

Before I dive into your problem, can you give me the font's version number?

Finii:
Hmm, on `master` the vertical spacing seems to be ok, but the horizontal spacing... ??!

![image](https://user-images.githubusercontent.com/16012374/211621150-bb44db7b-6c23-4537-a0d1-cc3bf6d3776b.png)

Lets see how it turns out if we patch with the latest patcher?

![image](https://user-images.githubusercontent.com/16012374/211621789-d67449fa-7d36-4308-9a19-b02727831faa.png)

Finii:
Overpass Mono is not monospaced :-(

![image](https://user-images.githubusercontent.com/16012374/211625653-ec0841eb-5731-45cc-bf8c-ac0181337699.png)

[...]

PR #1045, opened 2023-01-10
font-patcher: Fix some Nerd Font Mono too wide

Finii:
**[why]**
The 'monospace' width is determined by examining all the 'normal' glyphs and taking the widest one.

'Normal' means `0x00`-`0x17f`: the Latin Extended-A range.

Unfortunately Overpass (Mono) has wide-as-two-letters `IJ` and `ij` ligatures.

**[how]**
Exclude a small sub-range from the 'find the widest glyph' that contain these ligatures. Yes they will kind of break, but what can we do if we want to create a strictly monospaced font?

[note]
Related commit
  fbe07b8ab  Fix Noto too wide

Related: #1043

#### Requirements / Checklist

Exclude IJ and ij ligatures from the width calculation (for NFM), as they are too wide in the supposed to be `monospaced` font `Overpass Mono`.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -944,19 +944,29 @@ class font_patcher:
         # Ignore the y-values, os2_winXXXXX values set above are used for line height
         #
         # 0x00-0x17f is the Latin Extended-A range
+        warned = self.args.quiet or self.args.nonmono # Do not warn if quiet or proportional target
         for glyph in range(0x21, 0x17f):
-            if glyph in range(0x7F, 0xBF) or glyph in range(0x132, 0x134):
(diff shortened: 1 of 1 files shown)


Finii:
Searching systematically for more potentially problematic fonts...

These need a more detailed look at
* Anonymice
* FantastiqueSansMono
* iMWritingMono
* Inconsolata LGC
* Meslo LG M
* Monoid
* OpenDyslexicMono
* ProggyClean TT (seems too narrow)
* Roboto Mono
* Terminess TTF

image

image

image

image

image

image

image

image

image

image

Issue #1046, opened 2023-01-13
Unexpected Subfamily 'Book'


Finii:
Hmm,

```
$ tar zxf liberation-fonts-ttf-2.1.5.tar.gz
$ find *ttf-2.1.5 -name '*.ttf' -exec fontforge ~/git/nerd-fonts/font-patcher {} \;
$ fontforge ~/git/nerd-fonts/bin/scripts/name_parser/query_names *.ttf 2>/dev/null
```
```
Examining 12 font files
 |Filename                                           | | Fullname                                                          | | Family                                                  | | Subfamily                      | | Typogr. Family                           | | Typogr. Subfamily
 |-------------------------------------------------- |-| ------------------------------------------------------------      |-| ------------------------------------------------------- |-| ------------------------------ |-| ---------------------------------------- |-| ----------------------------------------
 |Literation Mono Bold Italic Nerd Font.ttf          | | Literation Mono Bold Italic Nerd Font                             | | LiterationMono Nerd Font                                | | Bold Italic                    | | LiterationMono Nerd Font                 | |
 |Literation Mono Bold Nerd Font.ttf                 | | Literation Mono Bold Nerd Font                                    | | LiterationMono Nerd Font                                | | Bold                           | | LiterationMono Nerd Font                 | |
 |Literation Mono Italic Nerd Font.ttf               | | Literation Mono Italic Nerd Font                                  | | LiterationMono Nerd Font                                | | Italic                         | | LiterationMono Nerd Font                 | |
 |Literation Mono Nerd Font.ttf                      | | Literation Mono Nerd Font                                         | | LiterationMono Nerd Font                                | | Book                           | | LiterationMono Nerd Font                 | |
 |Literation Sans Bold Italic Nerd Font.ttf          | | Literation Sans Bold Italic Nerd Font                             | | LiterationSans Nerd Font                                | | Bold Italic                    | | LiterationSans Nerd Font                 | |
 |Literation Sans Bold Nerd Font.ttf                 | | Literation Sans Bold Nerd Font                                    | | LiterationSans Nerd Font                                | | Bold                           | | LiterationSans Nerd Font                 | |
 |Literation Sans Italic Nerd Font.ttf               | | Literation Sans Italic Nerd Font                                  | | LiterationSans Nerd Font                                | | Italic                         | | LiterationSans Nerd Font                 | |
[...]

PR #780, opened 2022-02-07
Bugfix vertical lines in powerline prompts

Finii:
#### Description

First part of solution:

**[why]**
For some powerline symbols we add a certain amount of overlap into the
previous or next character to cover up a small gap between the symbols
that otherwise can show up as ugly thin (usually colored) line.

But after we carefully design that glyph with a bit overlap (over-sized
and having negative bearings) we remove all bearings. That breaks of
course the glyph and no actual overlap on the left side happens.

**[how]**
Just do not remove negative bearings on overlap-enabled glyphs. As they
are rescaled in both directions anyhow all bearings are wanted and must
be kept.

Second part of solution:

**[why]**
With the previous commit the gap between powerline glyphs became much    
better, but a faint line is still visible for LCD antialiasing (on my    
machine).    
    
Having seen how big the overlap is for Cascadia Code the overlap here    
can be increase maybe a bit.    
    
**[how]**
Increase the overlap to 3 % (from mostly 2 %) for the glyphs that have a    
'full colored edge' on one side.    
    
Fixes: #29  

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?

Increase the `overlap` for some glyphs slightly and correct the needed negative left side bearings for glyphs that should have had an overlap on the left side but had not.

#### Any background context you can provide?

The issue comes up relatively often, I think. I was never bothered by the lines, so I dod not really follow that. Maybe I'm wrong. I have real list of Issues that are related.

#### What are the relevant tickets (if any)?

#29

Also see #779 for `params` future as dict.

#### Screenshots (if appropriate or helpful)

@MIvanchev used `LiterationMono Nerd Font Mono` and neovim-Qt, so I use that here as examples.
Note: Hinting full, Subpixel antialiasing, X11

**Current behavior (before the PR):**

overlap0
image

**After fixing the code:** (first part)

overlap1
image
_Note more pronounced line for round things, as that has only 1%, the other have 2% overlap_

**After increasing the overlap:** (second part)

overlap2
image

_I guess 3% is enough, it is still very very faintly visible. Can be increased if users still complain_

PR #1047, opened 2023-01-13
font-patcher: Fix unexpected 'Book' SubFamily

Finii:
**[why]**
Sometimes we set an empty string as SubFamily name. That ends up as 'Book' which is unexpected.

**[how]**
The translation from empty to "Book" is done by Fontforge, at least with version 20220308.

Make sure we always have a SubFamily, and if we don't that must be a 'Regular'.

**[note]**
This was only a problem with the old naming engine. `--makegroups` got this right always.

Fixes: #1046

#### Requirements / Checklist

https://github.com/ryanoasis/nerd-fonts/pull/780#issuecomment-1381067808

### DIFF
--- a/bin/scripts/name_parser/FontnameParser.py
+++ b/bin/scripts/name_parser/FontnameParser.py
@@ -302,6 +302,9 @@ class FontnameParser:
         sfnt_list = []
         TO_DEL = ['Family', 'SubFamily', 'Fullname', 'Postscriptname', 'Preferred Family',
                 'Preferred Styles', 'Compatible Full', 'WWS Family', 'WWS Subfamily']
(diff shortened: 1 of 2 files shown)


Finii:
@MIvanchev

> I'm having some issues testing because the patcher sets weight to Book vs Regular. Is there a way to change this @Finii

It's a bug, it's fixed with this PR.

Finii:
Currently CI has problems, Azure ...
```
Could not connect to azure.archive.ubuntu.com:80 (20.106.104.242), connection timed out [IP: 20.106.104.242 80]
```

Finii:
Tests ran successfully on https://github.com/ryanoasis/nerd-fonts/actions/runs/3909272670

Merging.

Issue #400, opened 2019-10-31
Patching InputMono overwrites the letter `j`


CPWeaver:
#### 🎯 Subject of the issue

I'm running font-patcher to patch InputMono font. If I use `--complete` then the icon 
`nf-mdi-decimal_increase` overwrites the letter `j`. If I do not include mdi in the patching than the letter `j` is unaffected.

Found mention of this here but no ticket was opened:
https://github.com/ryanoasis/nerd-fonts/issues/164#issuecomment-410839942

#### 🔧 Your Setup

- Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_

InputMono, all variants. e.g. `InputMono-Medium.ttf` 
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_

iterm2 though this also shows up in the OSX Font Viewer
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_

OSX Catalina. It was true on Mavericks as well.

![image](https://user-images.githubusercontent.com/4966467/67987759-d7f08680-fbf3-11e9-8544-2adef4aae273.png)

Happy to provide any additional info if it'll help.

LKloosterman:
+1 As I have this issue too, thought I could live with it but it's very annoying.

I'm on Manjaro (Arch Linux), my terminal emulator is gnome-terminal and I'm using InputMono-Regular if it helps.

tdemarest:
+1 for another macOS Catalina 10.15.4, iterm2 3.3.10beta3, shows up in the Font Book. I only am using InputMono-Regular.ttf.

jackcogdill:
+1

kmalinich:
I had the same issue, and I tried every combination of dropping different command line arguments, but I couldn't get patching with all glyphs to work (without breaking the J character) ....... until I did it like this (a couple extra lines added for context)

(this is on macOS `10.15.6 (19G2021)`, with fontforge, GNU coreutils, GNU findutils and GNU parallel installed via brew and in my `$PATH`) 

Not really sure why this makes it work properly.

```bash
git clone --depth 1 https://github.com/ryanoasis/nerd-fonts
cd nerd-fonts
mkdir -p ~/fonts/InputMonoCompressed/Patched/
[...]

PR #1058, opened 2023-01-19
font-patcher: Cover alternate unicode encodings

nathanielevan:
#### Description

**[why]**
As described in [this comment](https://github.com/ryanoasis/nerd-fonts/issues/400#issuecomment-1397371200), issue #400 recently reoccurred with the latest build of Input font, and it turns out the dotless-j part of the small `j` now points to `U+0237`, which in turn has an alternate unicode encoding to `U+F6BE`; overwriting `U+F6BE` effectively overwrites `U+0237`, and in turn, alters the small `j`. This patch aims to fix that.

**[how]**
In addition to references, the patcher also checks for alternate unicode encodings which are returned by the `glyph.altuni` attribute, adds those to the essential set of glyphs, and in turn recursively searches for their references/alternate unicode encodings, making sure to handle circular references if they appear (for example: `U+2010` and `U+2011` in Input Mono).

Fixes: #400 

#### Requirements / Checklist

Make `font-patcher` also check for alternate unicode encodings in addition to references so they won't be overwritten.

Input Mono so far is the only font I've had this issue with, and this patch successfully prevents `font-patcher` from overwriting or altering those essential glyphs, but should work with most if not all fonts.

#400

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -921,13 +921,11 @@ class font_patcher:
                 # that means it has gone through this function before.
                 # Therefore we skip it to avoid infinite loop.
                 # A unicode value of -1 basically means unused and is also worth skipping.
-                if r[0] in self.essential or r[0] == -1:
-                    continue
-                self.add_glyphrefs_to_essential(r[0])
+                if r[0] not in self.essential and r[0] >= 0:
+                    self.add_glyphrefs_to_essential(r[0])
         for r in self.sourceFont[unicode].references:
-            if self.sourceFont[r[0]].unicode in self.essential or self.sourceFont[r[0]].unicode == -1:
-                continue
-            self.add_glyphrefs_to_essential(self.sourceFont[r[0]].unicode)
+            if self.sourceFont[r[0]].unicode not in self.essential and self.sourceFont[r[0]].unicode >= 0:
+                self.add_glyphrefs_to_essential(self.sourceFont[r[0]].unicode)
 
     def get_essential_references(self):
         """Find glyphs that are needed for the basic glyphs"""
--- a/font-patcher
+++ b/font-patcher
@@ -912,6 +912,23 @@ class font_patcher:
                 self.sourceFont.os2_typoascent += 1
                 self.sourceFont.os2_winascent += 1
 
+    def add_glyphrefs_to_essential(self, unicode):
+        self.essential.add(unicode)
+        # According to fontforge spec, altuni is either None or a tuple of tuples
+        if self.sourceFont[unicode].altuni is not None:
+            for r in self.sourceFont[unicode].altuni:
+                # If alternate unicode already exists in self.essential,
+                # that means it has gone through this function before.
+                # Therefore we skip it to avoid infinite loop.
+                # A unicode value of -1 basically means unused and is also worth skipping.
+                if r[0] in self.essential or r[0] == -1:
+                    continue
+                self.add_glyphrefs_to_essential(r[0])
+        for r in self.sourceFont[unicode].references:
+            if self.sourceFont[r[0]].unicode in self.essential or self.sourceFont[r[0]].unicode == -1:
+                continue
+            self.add_glyphrefs_to_essential(self.sourceFont[r[0]].unicode)
+
     def get_essential_references(self):
         """Find glyphs that are needed for the basic glyphs"""
         # Sometimes basic glyphs are constructed from multiple other glyphs.
@@ -921,8 +938,7 @@ class font_patcher:
         for glyph in range(0x21, 0x17f):
             if not glyph in self.sourceFont:
                 continue
-            for r in self.sourceFont[glyph].references:
-                self.essential.add(self.sourceFont[r[0]].unicode)
+            self.add_glyphrefs_to_essential(glyph)
 
     def get_sourcefont_dimensions(self):
         """ This gets the font dimensions (cell width and height), and makes them equal on all platforms """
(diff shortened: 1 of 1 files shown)


Finii:
Thanks for reporting and of course for the nice PR!
Really appreciate the commit style :tada: etc.

I will have a closer look at the weekend.

Finii:
I wonder if the `altuni` part needs to be handled at all in `self.essential`, of if the check while patching should be "check if this glyph's codepoint or any of its alternative codepoints is in `essential`".

But then, that would probably be more work because we need to check more often. And the altuni bindings will not go away or something, so yes, probably this is best.

font-patcher
```diff
@@ -904,6 +904,21 @@ class font_patcher:
...
+    def add_glyphrefs_to_essential(self, unicode):
+        self.essential.add(unicode)
+        # According to fontforge spec, altuni is either None or a tuple of tuples
+        if self.sourceFont[unicode].altuni is not None:
+            for r in self.sourceFont[unicode].altuni:
+                # If alternate unicode already exists in self.essential,
+                # that means it has gone through this function before.
+                # Therefore we skip it to avoid infinite loop.
+                # A unicode value of -1 basically means unused and is also worth skipping.
+                if r[0] not in self.essential and r[0] >= 0:
```
Finii, line 916: I like list comprehension, but that is maybe a question of personal style

```diff

-        if self.sourceFont[unicode].altuni is not None:
+        altuni = self.sourceFont[unicode].altuni
+        if altuni is not None:
-            for r in self.sourceFont[unicode].altuni:
+            for altcode in [ v for v, s, r in altuni if v >= 0 ]:
                 # If alternate unicode already exists in self.essential,
                 # that means it has gone through this function before.
                 # Therefore we skip it to avoid infinite loop.
                 # A unicode value of -1 basically means unused and is also worth skipping.
-                if r[0] not in self.essential and r[0] >= 0:
-                   self.add_glyphrefs_to_essential(r[0])
+                if altcode not in self.essential:
+                   self.add_glyphrefs_to_essential(altcode)
```

At least maybe add explanation for the `[0]`, by adding a line with
`# altuni is tuple of (unicode-value, variation-selector, reserved-field)`

Finii (approved):
This looks good, thank you.

Feel free to add some more comment in the code and/or use list comprehension if you like :grimacing: 

Tell me if I shall merge or if you want to add something before.
Again your commit style and the thought you put into this :+1: :green_heart:

Finii:
Ah I forgot, would be good to increase the script version.

font-patcher
```diff
@@ -904,6 +904,21 @@ class font_patcher:
             self.sourceFont.hhea_ascent = self.sourceFont.os2_winascent
             self.sourceFont.hhea_descent = -self.sourceFont.os2_windescent
 
+    def add_glyphrefs_to_essential(self, unicode):
+        self.essential.add(unicode)
+        # According to fontforge spec, altuni is either None or a tuple of tuples
+        if self.sourceFont[unicode].altuni is not None:
+            for r in self.sourceFont[unicode].altuni:
```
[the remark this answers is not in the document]
nathanielevan: Right, it is indeed arbitrary -- I "chose" `r` because it was already used in `get_essential_references()`, so I just used the same variable. `r` here simply refers to one of the altuni bindings currently being iterated. What would you suggest I call the variable?

font-patcher
nathanielevan: I'll try this right away, thanks for the recommendation! In all honestly I'm not very knowledgeable with Python so I actually have never heard of list comprehension, until today, that is.

nathanielevan:
Done! Let me know if you have any more recommendations, or whether I introduced bugs etc

Finii:
Did you test it with `Input`? I do not have that avail right away :grimacing: 

Looks very good :heart:

nathanielevan:
I did, and the glyph that was previously broken seems to have been fixed! But if you do have the font in hand you might want to test it just to make sure I didn't mess anything up.

Finii:
Rebase on master, force push

Finii:
Tried to switch this to some kind of tail-recursion, but the code will end up very ugly.

Also noticed that `Input Mono` indeed contains refs **and** alt-codes in the extended latin range.

Very nice code and PR, thanks again.

nathanielevan:
Sweet! I basically rely on Nerd Fonts for my day-to-day computer usage, so a little code is the least I can give back. Should I invoke the contributors bot as described [here](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md#how-to-add-yourself-to-the-contributors-give-yourself-attribution)?

Finii:
\*lauch\* yes please, I just searched for that syntax also, to add you :-)

Finii:
Hey @all-contributors please add @nathanielevan for code

Finii:
> I'm not very knowledgeable with Python

To be honest .. I really struggle to take Python seriously, and Nerd Fonts is the only project where I actually use it.

:thinking: No, `meson` uses it also I and contributed there. But not much.

Finii:
Thinking... Hope my release workflow will not break, because we added your contributor commit.
There was a problem when the repo is touched after the release has been started :thinking: :unamused: 
Wanted to save time by not always pulling again before pushing.
Lets see...
https://github.com/ryanoasis/nerd-fonts/actions/runs/3984511030

Finii:
Thinking... Hope my release workflow will not break, because we added your contributor commit.
There was a problem when the repo is touched after the release has been started :thinking: :unamused: 
Wanted to save time by not always pulling again before pushing.
Lets see...
https://github.com/ryanoasis/nerd-fonts/actions/runs/3984511030

Finii:
:sob: 

image

Issue #1056, opened 2023-01-19
Scaling inconsistency between text and glyphs


dstoliker:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
After upgrading to the latest version of the Hack nerd font via homebrew, there is an an apparent inconsistency between text and glyph scaling. I have a command line prompt configured with Oh My Posh with powerline symbols. After upgrading my font to the latest version, the glyphs no longer line up vertically. The glyphs are now taller. See screenshots below.

Expected behavior:
Text and glyphs should align as they did before the upgrade.

Example symbols:
\ue0b6
\ue0b4

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Hack Regular Nerd Font Complete.ttf_
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - Mac cask homebrew install
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - iterm2
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - MacOS 12.6.2 (Monterey)

#### ★ Screenshots (Optional)

Powerline bar as it appeared before font package upgrade:
<img width="205" alt="before update" src="https://user-images.githubusercontent.com/30450350/213335703-b24cba6d-3ce8-46e8-9c6f-921d4febeca9.png">

After upgrade:
<img width="214" alt="after update" src="https://user-images.githubusercontent.com/30450350/213335736-166c11b5-11f9-4b95-bd72-e7ecb1d24b0a.png">

iTerm2 font settings:
<img width="565" alt="font settings" src="https://user-images.githubusercontent.com/30450350/213335769-c52cc7b7-e110-4915-b798-844a66b8d667.png">

Finii:
Sorry to hear that. Can see the differences...

![image](https://user-images.githubusercontent.com/16012374/213361008-8776149d-31d3-444d-a403-ae001af852f5.png)

[...]

PR #1060, opened 2023-01-20
Rewrite font height calculation

Finii:
#### Description

**[why]**
The initial font-patcher used the WIN font metrics to determine the cell    
height. What has been found was forced into HHEA metrics but without    
observing the USE_TYPO_METRICS flag.    
That has been changed to use the TYPO metric instead of the WIN metric    
when the font wants that. For that the gap value becomes important.    
    
This is the current code. It still has problems to detect the correct    
cell height. A more rigorous approach seem to be needed.    
    
**[how]**    
The baseline to baseline distance is what we need as 'cell height', to    
fill it completely with the powerline glyphs. This is a little bit    
complicated and not really specified, each font rendering application or    
engine can handle the font metrics differently. But there are some    
common approaches.    
    
So we try to come up with the correct and congruent height, comparing    
different metrics and issuing a warning on problematic fonts.    
Afterwards we make all metrics equal (even if they were not before),    
because our goal is clear now and we impose it onto all platforms.    
    
**[note]**    
Useful resources:    
* https://glyphsapp.com/learn/vertical-metrics    
* https://github.com/source-foundry/font-line    
    
Fixes: #1056   

#### Requirements / Checklist

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -946,21 +946,11 @@ class font_patcher:
         our_btb = typo_btb if use_typo else win_btb
         if our_btb == hhea_btb:
             metrics = 1 if use_typo else 2 # conforming font
-        elif abs(our_btb - hhea_btb) / our_btb < 0.03:
-            print("{}: Font vertical metrics slightly off ({:.1f}%)".format(projectName, (our_btb - hhea_btb) / our_btb * 100.0))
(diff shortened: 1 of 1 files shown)


Finii:
Rebase on master, force push

Finii:
Important information is in #1056!

Issue #1081, opened 2023-01-26
JetBrains Mono family naming still broken


Finii:
Hmmm,

![image](https://user-images.githubusercontent.com/16012374/214907162-42b96a61-c66b-4af0-b3a0-93f9eacf1d48.png)

the naming is not ok. Strangeness!

_Originally posted by @Finii in https://github.com/ryanoasis/nerd-fonts/issues/1072#issuecomment-1405355545_

Finii:
`JetBrains Mono` should have been patched with `--makegroups`, but is obviously not.

Lets try...

![image](https://user-images.githubusercontent.com/16012374/214911362-7e4a06f0-3866-4773-8279-f5161adfa53a.png)

Flag missing. And I already know the reason, see #1072... sigh.

Finii:
Fixed, see specified flag:

![image](https://user-images.githubusercontent.com/16012374/214911844-ad1868b1-77ee-4d5c-9fdd-dc9ef7d4658e.png)

Finii:
See b98c2a1921, which was NOT fixed (until now)

PR #1082, opened 2023-01-26
patch-em-all: Fix ignored config.cfg

Finii:
**[why]**
JetBrains Mono is patched without `--makegroups`, although it is specified in the `config.cfg`.

**[how]**
As with the licenses the code expects a specific font directory depth. That is violated with some fonts, for example JetBrains Mono. The config is never found.

Compare commit
  8a749ab21  `patch-em-all: Fix missing licenses`

Fixes: #1081

#### Requirements / Checklist


Finii:
Naming before this:

image

Naming after this (`--makegroups` active):

image

PR #1086, opened 2023-01-27
font-patcher: Use git version tag for font version

Finii:
**[why]**
When changes are made to the `font-patcher` and fonts are patched with that version we can not see which patcher has been used in the fonts afterwards.

Would be good to have the usual version-patchversion number in the fonts in these cases (i.e. `v2.3.3-7` for 7 commits after `2.3.3`).

I did this manually before, but it is always a hassle.

[how]
If the `font-patcher` is run directly from a git repo and git is installed we try to get the latest tag version including patch number.

If and only if that is successful and that version is 'newer' than the version encoded in the `font-patcher` script the git version is trusted more.

#### Requirements / Checklist

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -1588,6 +1588,34 @@ def check_fontforge_min_version():
         sys.stderr.write("{}: Please use at least version: {}\n".format(projectName, minimumVersion))
         sys.exit(1)
 
+def check_version_with_git(version):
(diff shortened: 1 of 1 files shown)


Finii:
We need to check if/when the CI tags the repo. I guess afterwards, so the hardcoded version will win, which is what we want.

Issue #1057, opened 2023-01-19
Arch links need an update


moralground:
The AUR links are mostly broken now that the bulk of the fonts have been imported into Arch Linux's main repository:

https://archlinux.org/packages/community/any/nerd-fonts/

Some fonts remain in the AUR:

https://aur.archlinux.org/packages?K=nerd+font

I think it would be best to stop listing specific ones in the README and just put those two links somewhere for Arch users.

Finii:
Thanks for the information.

Yes, maybe it would be better to just link to the search page. And then, iirc we had that before and people complained that there should be direct links. I assumed the packages to be more stable.

But I guess you are right.

Should the Arch packages still be listed under 'AUR', because, strictly they are not. Hmm.

Ah I see. Yes, I will change it. Thanks again!

Finii:
_Note to self:_
Unrelated but related, the pkgver stuff, need to check if/how this works with `-git` packages.
https://github.com/ryanoasis/nerd-fonts/issues/573#issuecomment-770453386

Finii:
Link to
* #771

PR #1087, opened 2023-01-27
Update Arch package links

Finii:
**[why]**
All/most fonts are now available as Arch Community packages and not as AUR anymore.

**[how]**
Instead of listing all individually link to search page. Make sure the Symbols Only font in Community is also found.

Fixes: #1057

#### Requirements / Checklist


Finii:
Should the link not be https://archlinux.org/groups/any/nerd-fonts/ ? :thinking: (i.e. `any` instead of `x86`)

Anyhow. Also fixed the `gh-pages` with f2a8569

Issue #901, opened 2022-08-30
Incorrect display of character 'k' for "FantasqueSansMono" when "ss01" font feature is on


Frefreak:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
If a webpage enables "ss01" font feature, and render font is chosen as "FantasqueSansMono Nerd Font", then what should be displayed as
a normal `k` will show like a "coffee" icon. 

Expected behavior:

I see text without emoji

Example symbols:
```
jjjjkkkkllll
```

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_ **Fantasque Sans Mono Regular Nerd Font Complete.ttf**
   - _Where did you get the file from (download link, self patched, source downloaded from link...)_ **archlinux's AUR** [package](https://aur.archlinux.org/packages/nerd-fonts-fantasque-sans-mono)
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_  **alacritty (but this issue happens on webpage)**
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_ **archlinux**

#### ★ Screenshots (Optional)

I initially noticed it [here](https://github.blog/2022-08-29-gits-database-internals-i-packed-object-store/):
I override liberation mono and monospace with FantasqueSansMono Nerd Font using fontconfig's config file, so it's very likely what you see will be fine (because it's not rendered as this font). From the devtools image we can see my browser is rendering using "FantasqueSansMono Nerd Font".

![image](https://user-images.githubusercontent.com/7122156/187436836-48c61520-4b5e-4980-ab14-9c314248c40a.png)
![image](https://user-images.githubusercontent.com/7122156/187437578-d3033cc1-b66d-4fb0-b618-924fa798a200.png)

I believe this can be easily reproduced using this html snippet:
```
<style>
* {
  font-feature-settings: "ss01" on;
  font-family: FantasqueSansMono Nerd Font;
}
</style>
[...]

PR #1089, opened 2023-01-28
Do not destroy ligs or SS

Finii:
**[why]**
When a certain 'higher codepoint' glyph is needed for a substitution or ligature rule of a basic glyph and we replace the 'higher codepoint' glyph with a symbol that stylistic set or ligature will be broken.

**[how]**
We can not determine if a certain glyph is the _target_ of a pos-sub rule (at least I could not find a way). What we do is remove all pos-sub entries that _start_ at a symbol-patched glyph [1], but that is not the same.

Instead of walking through all substitution tables we just examine the 'basic glyphs' and also protect all glyphs that they reference through most of the possub tables.

In fact I encountered only "Substitution" entries and never "Ligature" entries, but we handle both alike. "Pair", "AltSub", and "MultSub" are not handled, but could be added if need be.

[1] #711

Fixes: #901

Reported-by: Xiangyu Zhu <frefreak.zxy@gmail.com>

#### Requirements / Checklist

Check all fonts and collect data on which codepoints will be skipped.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -1305,7 +1305,7 @@ class font_patcher:
 
         # end for
 
-        if not self.args.quiet or self.args.progressbars:
+        if not self.args.quiet:
             sys.stdout.write("\n")
 
 
--- a/font-patcher
+++ b/font-patcher
@@ -940,9 +940,17 @@ class font_patcher:
         # Find out which other glyphs are also needed to keep the basic
         # glyphs intact.
         # 0x00-0x17f is the Latin Extended-A range
-        for glyph in range(0x21, 0x17f):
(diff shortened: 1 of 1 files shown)


Finii:
image

image

Finii:
Affected:

* FantasqueSansMono: `E005`, `E006`, `E008`, `E009`
* InconsolataGo: `F40A`
* LiberationMono: `F004`
* RobotoMono: `F6C3`
* Tinos: `F004`
* Ubuntu: `F800` - `F81D`
* UbuntuMono: `F800` - `F81D`

Looks good. Astonished noone ever reported Ubuntu!

Issue #1098, opened 2023-02-01
Nerd Fonts nonMono not selectable with Windows Terminal


Finii:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

After installing

```
Caskaydia Cove Nerd Font Complete Regular.otf
Caskaydia Cove Nerd Font Complete Mono Regular.otf
Caskaydia Cove Nerd Font Complete Windows Compatible Regular.otf
Caskaydia Cove Nerd Font Complete Mono Windows Compatible Regular.otf
```

only the `Nerd Font Mono` variants are visible in Windows Terminal, when not selecting `all fonts`.

This is possibly a regression, I believe this had been fixed before.

They ARE selectable with Gnome-Terminal which is also picky sometimes.

Expected behavior:

Can select all 4

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
  See above
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
  Release zip 2.3.3
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   Windows Terminal 1.15.3466.0
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   Windows 10 no wait 11

#### ★ Screenshots (Optional)

![grafik](https://user-images.githubusercontent.com/16012374/216072630-047f37a1-8c0f-4aab-8c68-c7f3ac9dda3b.png)

Finii:
:sob: 

![image](https://user-images.githubusercontent.com/16012374/216080908-126591e6-e2e8-46dd-8315-16bfc5c877d9.png)

Finii:
Counterexample (old naming engine active)

![image](https://user-images.githubusercontent.com/16012374/216081952-03243536-3ba5-4778-811e-6e7e37a42d53.png)

Finii:
[...]

PR #764, opened 2022-01-14
font-patcher: Make Nerd Fonts Monospaced Again

Finii:
This reverts commit 59c45ba4eff539d2a83a7d7004f056b9860d3253.

In the following **_The Commit_** means the commit that this PR want to revert (remove).

**[why]**
_The Commit_ breaks the non-mono Nerd Fonts for a lot of people.
The issue which is the reason for _This Commit_ was not very good documented and investigation can not be
seen.

From _The Commit_'s TITLE it should have affected the 2048-em Symbols font only, but in fact all patched fonts were changed.
Maybe that was intended, maybe not.

**[how]**
This will make the advance width again equal for all glyphs.
This enables the use of the non-mono variant in more terminals.
For wider symbols a space is now (again) needed after the symbol.
That is expected by a lot applications.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?

Make Non-Mono Nerd Fonts monospaced but allow added symbols to extend over the allotted space to the right (negative right side bearings).

#### Any background context you can provide?

The behavior has been changed with _The Commit_ 59c45ba4e that has been written to fix this issue (at Arch) https://bugs.archlinux.org/task/66212.

Most of the fonts in `patched-fonts/` were not updated after that change, so the problems with that change became not obvious until recently.

If it is a problem or not depends also on the used terminal application. Some scale individual glyphs (like Windows Terminal), some just write them out.

**_What has been the problem and why is the solution a problem_**

A User complained that in their use case (which is not described in detail, but can be in a non-terminal program that shows a status bar somewhere) the added wide symbols overlap with immediately following ordinary letters.

This is the case because obviously the program that shows the symbols and letters follows the _advance width_, which is the same for every symbol in a monospaced font. The bigger symbols extrude to the right more than the cursor will more, so this is expected.
This is the reason why applications that use the symbols typically add a blank after the symbol itself, to preserve monospace grid and prevent overlap. For example vim-devicons adds an `ArtifactFix` (i.e. blank `' '`) when appropriate. Other programs like '`ls` with filetype symbols' also add a blank space directly after a symbol that is to be displayed.

After _That Commit_ the fonts are not monospaced anymore and terminal emulators do one of
* reject the font outright (not monospaced = not usable)
[...]

PR #1099, opened 2023-02-01
Bugfix Panose on "Nerd Font" variants

Finii:
#### Description

**[why]**
Some fonts have invalid (or unset) Panose flags. When we create a `Nerd
Font Mono` font the Panose proportion is set to `monospace`. This
make the font selectable in certain applications that need monospaced
fonts.
    
After #764 the `Nerd Font` variant (without `Mono`) shall (again) be detected as
monospaced font, but the glyphs have a big right side bearing (hang into
the next 'cell'). So we need to set the Panose bits there also.
    
**[how]**
We already have a check if the font is probably monospaced, independent
from Panose. This is used to prevent `--mono` patching on originally
proportional fonts.
    
If we find out with that check that the font is (most probably)
monospaced we also set the appropriate bits in Panose; unless Panose has
valid values that contradict that change.
    
Fixes: #1098

#### Requirements / Checklist

Install `CaskaydiaCove Nerd Font Complete.ttf` on Windows and try to select it in `Windows Terminal` (without 'all fonts').


Finii:
### All Panose related messages
Running on all fonts in `src/unpatched-fonts/`:

PR #1119, opened 2023-02-12
font-patcher: Make sure nf-custom-asm and nf-custom-v_lang icons are included

teatimeguest:
#### Description

The `nf-custom-asm` and `nf-custom-v_lang` icons are present in the [css file](https://github.com/ryanoasis/nerd-fonts/blob/master/css/nerd-fonts-generated.css) and [database script](https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/lib/i_seti.sh), but not in any patched fonts. Also, in the [cheat sheet](https://www.nerdfonts.com/cheat-sheet) they are displayed as blank rectangles since glyphs are not available (see below).

This PR fixes the `font-patcher` script to include these glyphs in font.

#### Requirements / Checklist

cheatsheet

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.5.5"
+script_version = "3.5.6"
 
 version = "2.3.3"
 projectName = "Nerd Fonts"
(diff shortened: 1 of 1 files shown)


Finii:
Thank you for the nice PR! :+1:

Finii:
Maybe we should do something to prevent this in the future.

One of, maybe
* Mention in some readme and or `icons.tsv`
* Make the range in `font-patcher` far longer, to just patch in all icons from `original-source` without need to change patcher

What do you think?

Finii:
I think @all-contributors should add @teatimeguest for code.

teatimeguest:
Thank you for your quick response and merging!

I think it would be very helpful to contribute if there are brief instructions or checklists in `contributing.md` (or elsewhere) for adding a glyph, as it is a bit confusing as to which files should be changed.

(Or, it might be best if CI could detect some inconsistencies between source files, if possible.)

Finii:
Well, maybe it is intentional that adding a new glyph to the core set is hard.

Because a better solution is always to get the new glyph accepted at some of the upstream symbol collection projects.
Adding it here should be the last resort after trying to add it somewhere else was unsuccessful.

One reason for that is that the codepoints that are left are dwindling away.
And I feel unable to decide if a glyph is 'important' or just of interest for 1.5 users, and even they will stop using it in 1/2 year?
Because once it's in there it should stay forever.

> Or, it might be best if CI could detect some inconsistencies between source files, if possible.

Well, by not specifying the end codepoint we would just patch in all codepoints.
I will add something to that effect.

Issue #1106, opened 2023-02-05
Few `Powerline Extra symbols` in `Caskaydia Cove Regular` are monospaced 


PROxZIMA:
#### Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

Few `Powerline Extra symbols` in `Caskaydia Cove Regular` are monospaced.

[`v2.3.0`](https://github.com/ryanoasis/nerd-fonts/releases/tag/v2.3.0) and above releases.
![image](https://user-images.githubusercontent.com/43103163/216806981-9e4c6b56-4363-45c8-ad34-dae55d34b6c3.png)

Expected behavior:
`CascadiaCode Regular` or `Caskaydia Cove Nerd Font Complete Regular` is non-monospaced font. So the above mentioned symbols should be wider.

[`v2.3.0-RC`](https://github.com/ryanoasis/nerd-fonts/releases/tag/v2.3.0-RC) and below releases.
![image](https://user-images.githubusercontent.com/43103163/216807042-423be8bc-15ab-4db2-91eb-cf3869868750.png)

Example symbols:

- `E0C4`: 
- `E0C5`: 
- `E0C6`: 
- `E0C7`: 
- `E0CC`: 
- `E0CD`: 
- `E0CE`: 
- `E0CF`: 
- `E0D1`: 

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Caskaydia Cove Nerd Font Complete Regular_
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - https://github.com/ryanoasis/nerd-fonts/releases/download/v2.3.0/CascadiaCode.zip
   - https://github.com/ryanoasis/nerd-fonts/releases/download/v2.3.0-RC/CascadiaCode.zip
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - kitty
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Arch Linux

[...]

PR #1123, opened 2023-02-12
 Reform PowerlineExtra sizing

Finii:
#### Description

**[why]**
The Powerline extra glyph sizing is not really clear.
    
**[how]**

* Make the triangulars 1 cell wide, as for example Iosevka also does.
* Make the Legos 2 cell wide with pa scaling to make them look nicer.
* Make the Hexagons 2 cells wide and keep their aspect ratio if possible.
* Make small and big Squares also 2 cell wide and keep their aspect ratio
of possible.
    
For the small and big Squares add a tiny bit of border (negative
overlap), because they have no smooth border line over their open and
closed squares, and that might look strange if some touch and some dont.
    
Fixes: #1106

#### Requirements / Checklist

Will follow

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.5.8"
+script_version = "3.5.9"
 
 version = "2.3.3"
 projectName = "Nerd Fonts"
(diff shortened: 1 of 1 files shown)


djdv:
>Opinions?

Just responding to reiterate my comment in the issue in regards to triangle widths.
A neutral +0 on it, since I'd personally be fine with either 1 cell or 2.
Since the pre-patched font already does 1 cell this change make sense.
But if anyone else has strong opinions on it, those can take precedence.

Finii:
This is how narrow symbols look like (note this is an older font, it has the cell size wrong):

image

_Font version 2.3.3_

and I believe the decisions on what is wide and what not is ok:

image placeholder

_Font version 2.3.3-69_

Finii:
The gaps (thin vertical lines) after the right hand side _flames_ and _waveform_ is not visible in terminal:

Screenshot 2023-02-18 at 17 28 29

🤔

Issue #1061, opened 2023-01-21
New material design icon block icons are much smaller and thus less readable.


Dragon-Hatcher:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
Identical icons from the old and new material design blocks render much smaller in the new block and are thus less readable.

Expected behavior:
Identical icons from the old and new material design blocks should render the same.

Example symbols:
`U+F631` (Old) vs `U+F0132` (New)
`U+F503` (Old) vs `U+F0004` (New)

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   -  `Symbols-2048-em Nerd Font Complete Mono.ttf`
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - Downloaded from the release page (version 2.3.1)
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - Kitty
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Linux Mint 21, Cinnamon

#### ★ Screenshots (Optional)

![image](https://user-images.githubusercontent.com/64214386/213830966-6f2913c9-8d76-4829-97ba-a6a29ad9ac62.png)
(Note that Kitty squishes icons into one cell when there is no trailing space and expands them to full size when there is. The first two icons in each section have a trailing space. Periods for scale. Here is the command I ran: `echo -e 'Old: U+F631    | U+F503    |\nNew: U+F0132 󰄲 󰄲 󰄲󰄲| U+F0004 󰀄 󰀄 󰀄󰀄|\n....................................'`)

Finii:
Hmm. Will check the scaling. 
It differs in methodics. 

Anyhow, bigger symbols should be in Symbols Only Nerd Font.ttf (without Mono). Maybe :thinking:

I will have a look. 



________________________________
From: Dragon-Hatcher ***@***.***>
Sent: Saturday, 21 January 2023 02:06
[...]

PR #1133, opened 2023-02-23
font-patcher: Scale all Material Design Icons individually

Finii:
**[why]**
The Material Design Icons have for sure pairs of glyphs that people would like to have scaled identically. Because the sheer number of glyphs and because they are already very nicely and uniformly scaled within their design space the MDI at the new codepoints where all scaled the same with taking the theoretical design space as ScaleGlyph.

But that means all icons get scaled a bit smaller than before, where we individually scaled each Material Design Icon to fill the cell. This lead to numerous complaints.

**[how]**
We take a different approach now, more conventional maybe. Especially in the light that the older bigger icons will get dropped; and people love them.
So the uniform scaling is ditched and the individual scaling is used.

Fixes: #1061

#### Requirements / Checklist

https://github.com/greshake/i3status-rust/pull/1728

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.5.10"
+script_version = "3.5.11"
 
 version = "2.3.3"
 projectName = "Nerd Fonts"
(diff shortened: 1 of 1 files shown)


Finii:
`Nerd Font Mono` before

image

`Nerd Font Mono` after

image

For the non-Mono font there is no change because that will only scale down if the symbol is wider than 2 cells, which is impossible by their design (for all 'normal' to be patched fonts)

PR #1134, opened 2023-02-24
Use fontconfig convention for naming

Jipok:
#### Description

 Rename `10-nerd-font-symbols.conf` to `30-nerd-font-symbols.conf`

#### Requirements / Checklist

Idk

From `/etc/fonts/conf.d/README`:
```
The files are loaded in numeric order, the structure of the configuration
has led to the following conventions in usage:

 Files beginning with:	Contain:

 00 through 09		Font directories
 10 through 19		system rendering defaults (AA, etc)
 20 through 29		font rendering options
 30 through 39		family substitution
 40 through 49		generic identification, map family->generic
 50 through 59		alternate config file loading
 60 through 69		generic aliases, map generic->family
 70 through 79		select font (adjust which fonts are available)
 80 through 89		match target="scan" (modify scanned patterns)
 90 through 99		font synthesis
```
But I'm not sure I chose the right number.


Finii:
Thanks for the PR.

Maybe you should add a `why`, a motivation for the change. I currently can not come up with a scenario where '10' would be too low and conflict with other, later settings.

I also have a hard time match the config to one of the categories. It is not a classical family substitution.

Usually the number is more important for system packages, so that this all merges well whichever packages a user has installed. So the system package maintainer will choose probably a suitable number i.e. priority.
For example the `fonts-powerline` package installs their config also in the 10th.

Jipok:
> It is not a classical family substitution.

Well, it's definitely far from "Font directories".
M.b. `60 through 69		generic aliases, map generic->family`?
This is very similar to how `60-latin.conf` works.

> Maybe you should add a why, a motivation for the change. 

Just to follow the convention.
I'm just figuring out how fontconfig works and trying to improve my own fonts. Which turned out to be very difficult and I constantly stumble upon redundant/broken/unused rules (I use someone else's configuration). If at least the rules names strictly follow the naming convention, it will be easier.

Finii:
> Well, it's definitely far from "Font directories".

True. But it is `10-` and not `00-`, so I do not understand that comment.

> Just to follow the convention.

Also true. But the 'convention' is not helping here. They all do not apply.
I still believe `10` is the most suitable here. And I see no argument which actual problem would be solved with a higher number.

First you say it should be `30`, then `60`. :thinking: 
As this is a _manually installed config_ I guess `10` is always good, and if it does not suit a user they can easily rename the file. Well, apart from the fact that I can hardly imagine any issue from this config being executed before the others; and you did not show an instance where it would make sense.

Please reopen if there are stronger arguments.

Issue #448, opened 2020-03-02
Text list of nerdfont characters


ranebrown:
Is there a full list of all nerdfont characters available similar to http://www.unicode.org/Public/UNIDATA/UnicodeData.txt? I'm looking for something that could be used with Unicode.vim. See https://github.com/chrisbra/unicode.vim/issues/39.

ryanoasis:
Hey, at first I was going to suggest https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/test-fonts.sh but that isn't for parsing just a formatted visually grep-able table.

https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/lib/i_all.sh is probably the closest script to get a list of characters but probably still not quite what you're looking for but likely needs to be tweaked as it's mostly a convenience for having variables represent the glyphs

meased:
I was looking for this as well. I couldn't find it so I made a quick script that parses the files mentioned above and outputs a nicer file with just the icon and the name of the icon. I posted it at https://gitlab.com/dmease/nerd-font-dmenu.

You can generate the file yourself, or just grab `nerdfont.map` from my repo and be done with it.

Jake-Gillberg:
Done again here, in the format like:

```
e717 dev github full                                                                                      
e718 dev nodejs small                                                                                     
e719 dev nodejs                                                                                           
e71a dev hackernews  
```

Finii:
?! Reminds me of https://github.com/ryanoasis/nerd-fonts/issues/1059#issuecomment-1404891287

The next comments have the commands how to create the lists. Very easy. What do you need?
And of course with hex numbers ;-)

mochaaP:
Hey @Finii, I changed it to return the glyphs directly.

I'm wondering can we have a list (preferably csv) in the repository or in the releases? Currently the only stable way to do that is parsing the scripts in `bin/scripts/lib/` or parsing the CSS. If those metadata could be provided, integrations can update their codepoints easily in case a future breaking change occurs.

Finii:
@mochaaP Creating such a list is very easy for Nerd Fonts (*). We just need to define how it shall look like. It can then be autocreated by the CI on any changes and be part of the Release packages.

Must admit did not read all the text above very carefully. Just define some format all people can live with. And then I also like `csv` but nowadays everyone uses `jq` and it _does_ have its benefits, so maybe json? This all depends of course what the **use** of the list is :-D

[...]

PR #1138, opened 2023-03-02
Add Codicons scale list

Finii:
**[why]**
All glyphs of the codicons set are individually maximised in size. That leads to the curious condition that 'circle small' looks bigger than 'cicrle' (because the line width is scaled up more -> looks bold).

Also some other 'subsets' look ugly and can not be used together.

**[how]**
Add appropriate ScaleGroups.

For the circles we also include one full-size circle as reference. To get less than maximal scale for the small circles.

#### Requirements / Checklist

Scale some Codicon icons down (i.e. not up to fill the cell), in `Nerd Font Mono`.

https://github.com/ryanoasis/nerd-fonts/issues/448#issuecomment-1451808965

Left side: before, right side: after.
Note that the small arrows (circled) are rather big instead.

image

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -867,10 +867,17 @@ class font_patcher:
         # For historic reasons ScaleGroups is sometimes called 'new method' and ScaleGlyph 'old'.
         # The codepoints mentioned here are symbol-font-codepoints.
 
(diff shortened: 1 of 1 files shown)


Finii:
image

_More problematic cases_

Issue #1142, opened 2023-03-09
font-patcher: not enough value to unpack (expected 3, got 2)


durka:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

I ran the font patcher script on a standard Ubuntu font and it did not complete.

```
$ python3 ./font-patcher /usr/share/fonts/truetype/ubuntu/UbuntuMono-R.ttf
Nerd Fonts Patcher v2.3.3 (3.5.1) executing
The following table(s) in the font have been ignored by FontForge
  Ignoring 'DSIG' digital signature table
  Ignoring 'VDMX' vertical device metrics table
The glyph named fi is mapped to U+F001.
But its name indicates it should be mapped to U+FB01.
The glyph named fl is mapped to U+F002.
But its name indicates it should be mapped to U+FB02.
The glyph named Omega is mapped to U+03A9.
But its name indicates it should be mapped to U+2126.
('period', (1.0, 0.0, 0.0, 1.0, 0.0, 0.0))
('period', (1.0, 0.0, 0.0, 1.0, 0.0, 323.0))
Traceback (most recent call last):
  File "/tmp/FontPatcher/./font-patcher", line 1732, in <module>
    main()
  File "/tmp/FontPatcher/./font-patcher", line 1712, in main
    patcher.patch(sourceFonts[-1])
  File "/tmp/FontPatcher/./font-patcher", line 268, in patch
    self.get_essential_references()
  File "/tmp/FontPatcher/./font-patcher", line 948, in get_essential_references
    self.add_glyphrefs_to_essential(glyph)
  File "/tmp/FontPatcher/./font-patcher", line 935, in add_glyphrefs_to_essential
    for refcode in [ self.sourceFont[n].unicode for n, m, s in references ]:
  File "/tmp/FontPatcher/./font-patcher", line 935, in <listcomp>
    for refcode in [ self.sourceFont[n].unicode for n, m, s in references ]:
ValueError: not enough values to unpack (expected 3, got 2)
```

[...]

PR #1144, opened 2023-03-10
font-patcher: Fix crash with older fontforge

Finii:
[why]
If a font has references in glyphs that we want to add to the essential set of glyphs, and fontforge is old (i.e. 2020*) the patcher crashes.

[how]
The fontforge function glyph.references returns a three element tuple in current fontforge (i.e. 20230101). But older versions skip the selection bit and return only tuples of two.

As we use only the first tuple element we do not care about the 2nd and possible 3rd element(s) and just thrash them.

Fixes: #1142

#### Requirements / Checklist

Fix running `font-patcher` on older `fontforge`

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "3.5.11"
+script_version = "3.5.12"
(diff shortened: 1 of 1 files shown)


Finii:
OH NO I pushed to master instead here :-(

Issue #1140, opened 2023-03-08
nerdicons.nvim support search icon and copy icon inside neovim


glepnir:
Hi I am not sure i can create this issue for my plugin first . if it's not allowed please close it.  I create a neovim plugin allow search icons in side neovim and easy to copy icon .  [nerdicons.nvim](https://github.com/glepnir/nerdicons.nvim)

![gif](https://user-images.githubusercontent.com/41671631/223701949-88c95719-8371-4572-8573-987e6c17c373.gif)

glepnir:
yeah it's okay.  the `icons.lua` is generate by script. it parse the html source code of nerdfonts website. this json file very useful thanks !!

Finii:
> doest this json file include the `obsolete` icons?

It is the same data as the html (css / cheat sheet), so yes. They are not flagged specifically; but the range is clear (F500-FFFF, roughly speaking).

> is generate by script

A perfect place to learn github workflows and let the script be run anytime its source changes ;-)

...

And sure, if you find a nice place in the docs...

glepnir:
yes i can create a github action for generate the icons.lua . also i can create it by reading json file in neovim by json decode .

thanks :)

Finii:
:-) And thanks for closing this issue.

I'm currently fighting to get the issue count down to be ready for version 3.0.0.

Well, I am quite proud how I pulled down the number in the last time. Its so hard to keep up with incoming issues :grimacing: 

![image](https://user-images.githubusercontent.com/16012374/223712057-eec995f7-2487-47b8-bc4a-d1a040fe3d64.png)

_Number of issues over time, via https://9-volt.github.io/bug-life_

glepnir:
Thank you for your excellent work. 👍

glepnir:
looks like this json file does not have the icon code ..so i can't to know the icon in `F501-FFFF` range. can we have a new json file or something else that i can use .

Finii:
Well, see
* #448 

Unfortunately it is not easy to add the raw code without breaking backward compatibility.

But to solve your problem, can you not just extract the code from the glyph? Depending on your language, but that should be very easy.

![image](https://user-images.githubusercontent.com/16012374/224558016-9a5ce469-ab5d-4918-98a9-57bc0749754a.png)

Or check if the name starts with `"mdi-"`, they are the affected (obsoleted) icons.

Finii:
@glepnir Added codepoint to json data with commit above.

[...]

PR #1147, opened 2023-03-12
generate-css: Add codepoint to glyphnames.json

Finii:
**[why]**
Just having the characters themselves as data for the name is convenient in some cases; but sometimes the codepoint is also nice. And it certainly improves the human interaction / check of the file.

**[how]**
Include the character and the codepoint in the glyph-name data. This is a breaking change. But given that the json file is rather new I believe we can get away with it.

Sorry for the inconvenience for all the early adaptors that already use the old/previous format. But now we are more future proof and can add even more data without breaking old json code.

Related:
* https://github.com/ryanoasis/nerd-fonts/issues/1140
* https://github.com/kovidgoyal/kitty/issues/2972
* https://github.com/chrisbra/unicode.vim/issues/39
* https://github.com/ryanoasis/nerd-fonts/issues/448
* https://github.com/nvim-tree/nvim-web-devicons/issues/192

#### Requirements / Checklist


Finii:
Carry over the fix to the `gh-pages` manually (this is done automatically only on releases) with above commit 9498b25.

Issue #490, opened 2020-06-16
feat: update Octicons to latest


jef:
#### 🗹 Requirements

- [x] A brief but descriptive _title_ of your issue
- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have read or scanned the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] I have read or scanned the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)

#### 🎯 Subject of the issue

Update the Octicons to [10.0.0](https://github.com/primer/octicons/releases/tag/v10.0.0). Released on 06/10/2020.

#### ★ Optional

Link to new Octicons: https://primer.style/octicons/

djensenius:
Any update on this?

Finii:
@djensenius This has a v3.0.0 tag. That release is due ... some days ago.

See https://github.com/ryanoasis/nerd-fonts/discussions/1074. Updated all source fonts, only some few issues open, this is one of them. Hope to finish this month.

djensenius:
Awesome! Thanks @Finii !

Finii:
Hmm, there seems to be no updated opticons font anymore in the original repo.
At least according to Arch, who build the font on their own from the svgs:

![image](https://user-images.githubusercontent.com/16012374/227575914-8b1a58b0-c78d-40ed-a8d8-9cd639986b97.png)

https://aur.archlinux.org/packages/ttf-octicons

They use `fantasticon`, but that breaks the codepoint stability, or not? Checking.
That would mean we need to self-create that font.
Do you have any information on this @djensenius ? Tnx.

Finii:
Hmm, the produced font has completely different codepoints / sorting, this is not useful... :-(

![image](https://user-images.githubusercontent.com/16012374/227619582-021f4a7b-8b1b-4ded-9f84-a3a01d417d8d.png)

Finii:
The mapping here https://github.com/microsoft/vscode-octicons-font seems to be the old one / the one we use.
We can use that to resurrect and update the font generation with current svgs.

Finii:
At the moment the octicons are at `F400` with length `105` which after gap removal turns into end codepoint `F4A9`.
Currently at `F500` Material Design starts...:

[...]

PR #1156, opened 2023-03-25
Update octicons and drop old MDI

Finii:
### This can only be done after MDI-old has been dropped

**[why]**
The octicons got a lot updates.
But they do not have a font anymore.

**[how]**
Keep our old codepoints constant, but add the new icons thereafter.

Fixes: #490

#### Requirements / Checklist

Update the Octicons icons.

If one icon has been dropped (without replacement) re-use its codepoint for a newly added icon.
Append other newly added icons in the end.
Afterwards Octicons will be gapless in `F400` - `F532` in Nerd Fonts.

Note that we reassign codepoints on patching. The original octicons font has many gaps.
To keep also compatible here the font we create here is a bit ... stranger, with a lot of gaps and some gaps filled.
The methodology seems complicated by the goal is easy: We just want to patch in the icon into Nerd Fonts with the aformentioned codepoints and no additional special rules.

This means for the `octicons.ttf`
Existing gaps are preserved.
Dropped icons are replaced by newly added ones.
The remainder of new icons is added in the back, which is FAR back, because of the one stray icon the original (old) font had.


Finii:
Current output:

```
Reading mapping file
Found 172 entries
Fetching octicons archive "v18.2.0.tar.gz"


Unpacking octicons archive
Found 574 svgs
Found 308 icons after de-duplicating

Renamed cloud-download -> download
Renamed cloud-upload -> upload
Renamed clippy -> paste
Renamed mail-read -> read
Renamed primitive-dot -> dot-fill
Renamed primitive-square -> square-fill
Renamed settings -> sliders
Renamed dashboard -> meter
Renamed trashcan -> trash
Renamed paintcan -> paintbrush

Missing octoface
Missing keyboard
Missing gist
Missing file-text
Missing file-pdf
Missing jersey
Missing radio-tower
Missing repo-force-push
Missing mail-reply
Missing arrow-small-right
Missing gist-secret
Missing no-newline
Missing arrow-small-up
Missing arrow-small-down
Missing arrow-small-left
Missing file-symlink-directory
Missing circuit-board
Missing watch
Missing text-size
Missing ellipses
Missing plus-small

Found 151 (of 172, missing 21) and new 157
```

I guess the icons that have been dropped are no problem (but then, what do I know what people use these icons for).
I would just drop them, update the rest, fill in the new gaps with new icons and append the rest of new icons. As they have no codepoints yet we can assign whatever we want.

Finii:
Octicons have 2 sets of icons, one for 16px and one for 24px scaling. Unfortunately not all icons are available in both sizes.

Whatever, here a comparison of old (left) versus new (right)

image

The old size seems to be in-between, stroke size wise:

image

Finii:
@jef @djensenius What do you think, should we use the 16px or the 24 px octicons?
Given terminal font size 11-15 pt I would guess the 16 px size is better (at the usual 96 dpi)

Finii:
Fixed copyright issue, force push (sorry for the frequent pushes)

Finii:
We need to wait until Material Design Icons moved out of the way, because we need their old (dropped with v3.0.0) codepoints partially (`F000` - `F532`).

We just need to do `fontforge generate` in `src/glyphs/octicons/` and commit the changes as commit 2/2.

jef:
> @jef @djensenius What do you think, should we use the 16px or the 24 px octicons?
> Given terminal font size 11-15 pt I would guess the 16 px size is better (at the usual 96 dpi)

Great point. IIRC, the 24px octicons have some that aren't available in 16px or vice versa.

If not, then 16px sounds great!

Finii:
Thanks for the reply. I had the same feeling, so be it. 👍 

> Great point. IIRC, the 24px octicons have some that aren't available in 16px or vice versa.

Both. The `generate` script builds a list of all icons, regardless of size. Then later it tries to find that icon in the preferred size and if not available falls back to the other size.

_From image above, note 16px icons fill in where not 24px icons are available:_
Screenshot 2023-03-26 at 22 57 11

jef:
Amazing! Nice work and thanks for all the updates :)

Finii:
image

Right! :woman_facepalming: 

This PR is just preparatory and commit 1/2. Needs to be completed!

Issue #1170, opened 2023-04-13
Font Patcher: UnboundLocalError: cannot access local variable 'box_keep' where it is not associated with a value


hrqmonteiro:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
Following error when trying to patch font (tried all the methods):

```
fontforge -script font-patcher Roboto-Regular.ttf -c
Copyright (c) 2000-2022. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20220308
 Based on sources from 2022-07-21 00:00 UTC-ML-D-GDK3.
Nerd Fonts Patcher v2.3.3 (3.7.1) (ff 20220308) executing
The following table(s) in the font have been ignored by FontForge
  Ignoring 'hdmx' horizontal device metrics table
Traceback (most recent call last):
  File "/home/hrq/Downloads/nerd-fonts-master/font-patcher", line 1943, in <module>
    main()
  File "/home/hrq/Downloads/nerd-fonts-master/font-patcher", line 1923, in main
    patcher.patch(sourceFonts[-1])
  File "/home/hrq/Downloads/nerd-fonts-master/font-patcher", line 326, in patch
    self.setup_patch_set()
  File "/home/hrq/Downloads/nerd-fonts-master/font-patcher", line 880, in setup_patch_set
    'default': {'align': 'c', 'valign': 'c', 'stretch': 'xy', 'params': {'overlap': 0.02, 'dont_copy': box_keep}},
                                                                                                       ^^^^^^^^
UnboundLocalError: cannot access local variable 'box_keep' where it is not associated with a value
```

Expected behavior:
Not to happen this error?

Example symbols:
I don't get this.

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
  None, trying to patch
[...]

PR #1171, opened 2023-04-14
font-patcher: Fix patching non-monospaced fonts

Finii:
[why]
Any non-monospaced font will not be be patched, the patcher crashes.

[how]
This must have happened when the box drawing characters rescaling feature has been disabled. The default value (False) is not always set.

The box drawing patch has the ability to rescale existing box glyphs. That used to be done when all box glyphs are already existing in the source font. We do not patch in a new glyph set then, but we rescale the existing glyphs to match the possibly new cell size. But that feature is disabled and the attribute 'dont-copy' is never utilized. It is disabled because some existing box sets are rather ... sspecial in their overlap and can not be scaled as we would scale them.

Fixes: #1170

Reported-by: Henrique Monteiro <hrqmonteiro@protonmail.com>

#### Requirements / Checklist

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -774,19 +774,20 @@ class font_patcher:
         """ Creates list of dicts to with instructions on copying glyphs from each symbol font into self.sourceFont """
 
         box_enabled = self.source_monospaced # Box glyph only for monospaced
+        box_keep = False
         if box_enabled:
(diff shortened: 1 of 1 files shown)


Finii:
Now I again forgot to increase the script version :angry:

Finii:
Hmm, and this has already been fixed with 4dc8d393e42, which was not merged yet.
To many things...

Issue #998, opened 2022-11-18
Cascadia Code text height doesn't match original


longnguyen2004:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
The patched Cascadia Code font doesn't have matching text height to the original, which makes it look unusually large

Expected behavior:
The 2 fonts to have identical text height

Example symbols: Letters and numbers

#### 🔧 Your Setup

- _Which font are you using_
   - `Caskaydia Cove Nerd Font Complete Regular.otf`
   - Downloaded from GitHub Releases
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
Windows Terminal
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
Windows 11 22H2

#### ★ Screenshots (Optional)

Original
![image](https://user-images.githubusercontent.com/15729831/202751511-8cd2f967-f4ac-4a78-ad2a-5c5fbf6a8c46.png)
![image](https://user-images.githubusercontent.com/15729831/202752100-0d95e90a-65da-4adc-945e-0f73cd51d9c2.png)

Patched
![image](https://user-images.githubusercontent.com/15729831/202751581-8e34240c-a48c-438c-a2b9-81cc7da7db3f.png)
![image](https://user-images.githubusercontent.com/15729831/202752201-66fd731f-bb2a-4ddb-a0ac-419edcddd82a.png)

Finii:
Here a visualization of the changes:

![Peek 2022-11-18 17-48](https://user-images.githubusercontent.com/16012374/202758247-731f6c8e-4e78-46f6-92f3-3e909cf55dd5.gif)

Could you please share the font details one can see with the font settings dialog when you click on that concrete font?

I guess you compare the static nerd font version with the dynamic default-for-Windows-Terminal version?
Those two differ already when you download them as original, iirc.

longnguyen2004:
Which font settings dialog are you talking about? The one in Windows' Settings app?

Finii:
[...]

PR #1179, opened 2023-04-27
CascadiaCode: Rehint and use ttf

Finii:
**[why]**
A lot people (read: People on Windows) have the variable font (VF) version of Cascadia Code installed - it comes bundled with Windows Terminal.

The static Cascadia Code instances that we use for patching are hinted with ttfautohint which creates small sized glyphs that are visibly very different. People compare the static Caskaydia Cove with the VF Cascadia Code and are surprised.

**[how]**
First switch from the CFF outlines to TTF outlines - that is the original version (i.e. otf -> ttf). It is unknown why we created patched CFF fonts instead of the TTFs. To get as close as possible to the intended look of the glyphs we should stick with the outline type.

Then we need to re-hint all the fonts, to get hints that are comparable to the VF hints. We can not use the hints of the VF because the outlines are different: The VF has (of course) overlapping outlines, while the static ones (as usual) have not.

The re-hinting can be done with VTT or TTH - both showed results that are more like the original VF font. The usual ttfautohint has been used of the static fonts in the font release and can not be used. It is the reason for this whole problem.

* Used VTT 6.35
* Open font file in VTT
* Import all programs
* Generate 'VTT talk' via Tools -> AutoHint -> LightLatinAutoHint
* Save font file as ...

References:
* https://github.com/microsoft/cascadia-code/issues/371
* https://learn.microsoft.com/en-us/typography/tools/vtt/

Closes: #998

#### Requirements / Checklist


Finii:
## How VTT looks like

grafik

## Static font before and after re-hinting

The x-height changes by one pixel at some sizes

grafik

Finii:
Note that VTT makes the font files **considerably** bigger. On the other hand the fonts will be even far more bigger after we patch them. But there is for sure room for improvement.

Issue #1187, opened 2023-05-01
Space Mono Ligature fi looks strange


Fxzzi:
#### 🗹 Requirements

- [ X ] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [ X ] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [ X ] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
Ligatures should be shown with Space Mono NF. When typing `'fi'` there is a space after and they are stuck together as shown here
![image](https://user-images.githubusercontent.com/18248986/235455202-d187722a-4870-42f6-9ba1-2c137ff1c95c.png)

Expected behavior:

The ligatures for => and ==> are not included and / or shown using this font.

Example symbols:

`->=> ==> fi`

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_
   - SpaceMonoNerdFont-Regular.ttf
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - TTF from releases page (packaged in pacman repo)
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - 
   - Kitty
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - 
   - Arch Linux with Kernel 6.3.0

#### ★ Screenshots (Optional)

![image](https://user-images.githubusercontent.com/18248986/235455658-8cd02374-1a9f-4cc7-8f15-593efa22743a.png)

wjcostel:
Maybe the name of the font has changed (on the system)?  I just had this happen to me after an update and I went back to the config files and opened up Font Manager and plugged in the names of the fonts as they appeared in Font Manager.  I'm not entirely sure though because I may have two versions of the same font (nerd font & non-nerd font) and now I'm just telling the dotfiles to refer to the non-nerd font versions.

Finii:
Can you show your kitty conf?

Ah, if I remember correctly I was unable to turn on (?) any ligatures in kitty. Is there a trick?

Hmm, but even in Writer there seem to be no ligs?

[...]

PR #1191, opened 2023-05-02
Prepatched fonts: Revive some ligature removal

Finii:
**[why]**
Some sourcefonts, even that are monospaced, have a `fi` and/or `fl` ligature that maps into one cell. That looks very strange.

**[how]**
Partially revert commit
  148b0c445  Sunset ligature removal

for the cases that have a one-cell `fi`, `fl`, etc ligature, or a `ldot` related ligature - that is active by default. Discretionary ligatures or Stylistic Sets are not changed.

Do the removal on all patched fonts for consistency, not just `Nerd Font Mono`.

**[note]**
On Noto different subtables are needed for Sans, Serif and Sans-Mono. We can not set up different configs for each, so all are tried in all fonts and might fail (this is normal).

Fixes: #1187

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/gotta-patch-em-all-font-patcher!.sh
+++ b/bin/scripts/gotta-patch-em-all-font-patcher!.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # Nerd Fonts Version: 3.0.0
-# Script Version: 1.4.3
+# Script Version: 1.4.4
 #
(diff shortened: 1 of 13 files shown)


Finii:
This direct subtable removal is huddely muddely unclean :unamused: 
Maybe we will find a better solution somewhen, or at least write a helper that finds all the different subtable names in the font weights... Whatever, this seems to do.

Issue #1198, opened 2023-05-03
Nerd Font Mono some glyphs too wide


Finii:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

This has been reported by @jimeh via Gitter.

Experienced behavior:

![image](https://user-images.githubusercontent.com/16012374/235843866-31f3493d-75cb-4698-9391-ac0a3033e83e.png)

Expected behavior:
Glyphs should not extend out of the 'cell' for Nerd Font Mono.

The patcher also complains:
```
Adding 169 Glyphs from Octicons Set
╢█████████████████████████████░░░░░░░░░░░╟ 72%WARNING: Scaled glyph F478 wider than one monospace width (1410 / 1233 (overlap None))
╢█████████████████████████████░░░░░░░░░░░╟ 72%WARNING: Scaled glyph F479 wider than one monospace width (1407 / 1233 (overlap None))
╢████████████████████████████████████████╟ 100%
```

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - 
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - 
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   -

Finii:
Octicons originates in `F000` and is patched into `F400`...

```
        OCTI_SCALE_LIST = {'ScaleGlyph': 0xF02E, # looking glass (probably biggest glyph?)
            'GlyphsToScale': [
                (0xf03d, 0xf040), # arrows
                0xf044, 0xf05a, 0xf05b, 0xf0aa, # triangles
                (0xf051, 0xf053), # small stuff
                0xf071, 0xf09f, 0xf0a0, 0xf0a1, # small arrows
                0xf078, 0xf0a2, 0xf0a3, 0xf0a4, # chevrons
                0xf0ca, # dash
```

[...]

PR #1199, opened 2023-05-03
Bugfix/fix octa scale

Finii:
```
font-patcher: Fix some Octicons scales
    
    [why]
    With the Octicons update some codepoint contents changed.
    Also some symbols scale still wrong.
    
    [how]
    Check fonts and add / drop entries.
    Use GlyphsToScale+ to correct position of Github text icon.
```    

Fixes: #1198

```
font-patcher: Modernize and expand Octicons scale table
    
    [why]
    The scale glyph is not the biggest glyph; the bell-slash is a little bit
    wider than the design cell...
    
    [how]
    Switch to modern ScaleGroups. These determine the virtual combined
    bounding box of all glyphs in the group and scale them 'as one'.
    
    Also add some more groups.
    
    Also correct wrong code of one small stuff.
```

#### Requirements / Checklist

* #1156

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -982,16 +982,17 @@ class font_patcher:
             range(0xf221, 0xf22d + 1), # gender or so
             range(0xf255, 0xf25b + 1), # hand symbols
         ]}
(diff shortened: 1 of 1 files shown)


Finii:
@allcontributors add @jimeh for bug

Issue #1214, opened 2023-05-08
cod-debug-continue-small is not small


Finii:
#### 🗹 Requirements

- [ ] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [ ] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [ ] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

<img width="350" alt="Screenshot 2023-05-08 at 06 30 25" src="https://user-images.githubusercontent.com/16012374/236734346-73bf7cfb-71ce-4307-8951-7a6dcec3bf42.png">

Maybe there are more

Expected behavior:

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - 
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - 
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   -

Finii:
![image](https://user-images.githubusercontent.com/16012374/236834155-7f065748-21e4-4e80-a19d-66559556888d.png)

Finii:
Closed via #1216

PR #1216, opened 2023-05-08
Codicons: Preserve smaller icons

Finii:
[why]
The codicons have 3 obvious (from the name) pairs of regular and small icons. The size difference is slight but visible.

When we maximize all individual icons the differences are mostly lost.

[how]
Put each pair into one ScaleGroup to keep their individual scaling.

Fixes: #1214

#### Requirements / Checklist


Finii:
Why is this closed without commits?

Issue #1213, opened 2023-05-07
f482 went from symlink directory to bookmark slash?


duhdugg:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

`f482` was previously showing a symlink directory icon, and now shows a bookmark slash icon

Expected behavior:

symlink icon

Example symbols:



#### 🔧 Your Setup

- ttf-firacode-nerd package from Arch repo

#### ★ Screenshots (Optional)

Before Upgrade (`ttf-firacode-nerd-2.3.3-3`)

![2023-05-07-162650_61x31_scrot](https://user-images.githubusercontent.com/1406429/236702079-45a2e657-ac36-45aa-87fc-c166ee3c78a9.png)

After Upgrade (`ttf-firacode-nerd-3.0.0-2`)

![2023-05-07-162710_65x28_scrot](https://user-images.githubusercontent.com/1406429/236702090-67807eef-7df4-4c8a-b7ac-3f0ae21781cc.png)

Using the `lsd` command to generate the above screenshots, which shows this codepoint for symbolic link directories.

## Related Issue

Submitted this issue to the octicons project as well: https://github.com/primer/octicons/issues/945

_Edit Fini: Changed codepoint number_

Finii:
Sorry to hear that you lost a beloved icon with the update.

As you already found out the change is due to the Octicons update.

They dropped a lot icons; and the remaining folder icons changed to a more modern look.

That you see now the bookmark-slashed and not a `.notdef` (i.e. questionmark box) is because we reused the freed codepoints for newly added icons. The Octicons do not have native codepoints.

Unfortunately link-folder icons are not very widespread, so the only one left probably is this:

<img width="163" alt="Screenshot 2023-05-08 at 06 42 48" src="https://user-images.githubusercontent.com/16012374/236735605-c39993d8-1f02-42f3-8c8d-ffcf0a2c52f6.png">

But the 'folder' part looks quite different of course. Probably using that would mean to switch all folder related icons to Codicons.

[...]

Issue #1207, opened 2023-05-06
Icon nf-oct-file_directory changed on update to 3.0.0


Mangeshrex:
Hey, I actually love an icon from the version 2.3  nerdfonts as due to the commit of modernizing the fonts, that glyh was uncertainly removed from it. One of my friend managed to get the file for that file from some other repo. 

So, I copied that font to a different file and managed to make a generate a new font with only that icon in the file with the help of fontforge. But the issue is when i'm using a font from the v3.0 the icon is hollow and If I use the older version it is how I expect it to be. 

Expectations (tested on font from v2.3) for v3.0
![Sat May  6 11:39:49 IST 2023](https://user-images.githubusercontent.com/82205152/236604623-31c284e0-3e94-40fb-9f4b-0f00b98de371.png)

This is how it is when using v3.0 fonts 
![Sat May  6 11:11:36 IST 2023](https://user-images.githubusercontent.com/82205152/236604777-bd07db33-525d-4be8-895f-20769954f186.png)

- [x ] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [ x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [ x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

This is the symbol / glyph I want to be like expected above
 

## My Setup 
Arch Linux

Finii:
Is this about the directory/folder icon?

Mangeshrex:
yes

Mangeshrex:
could you please provide some help ? as it's very urgent

Mangeshrex:
also there's an endpoint in nerd fonts with the same endpoint as the one which I created in the separate file. So maybe the nerd font one is overlapping the one I made. So, is it that I have to create a new endpoint and shift this icon to that endpoint ?

Finii:
This in a duplicate of the comment here: https://github.com/ryanoasis/nerd-fonts/issues/1190#issuecomment-1532632029

You seem to be missing the same glyph?

![image](https://user-images.githubusercontent.com/59060246/235865321-750637d5-54b7-471e-b3b2-b0659556a713.png)

Octicons updated their design and a lot of their icons got modernized and unified in that process.

If it's just that it is 'open', maybe try one of their others:

[...]

Issue #1190, opened 2023-05-01
Some glyphs missing after v3.x.x update


luizkowalski:
#### 🗹 Requirements

- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [X] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [X] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
this morning I updated some deps on my computers and some fonts were updated as well but this is not a big deal since this has happened before, but after the update finished I noticed some icons on my neovim were wrong like the `Gemfile` icon and the `X` for closing a buffer:

<img width="754" alt="image" src="https://user-images.githubusercontent.com/112706/235512942-fddd3422-635a-46a7-a0b0-47e2e488ce54.png">

This was weird so I decided to delete all custom fonts I had, clean up the cache, re-download them manually, and reinstall it. To my surprise, nothing has changed

No idea what is going on, this is my text configuration on iTerm2:

<img width="583" alt="image" src="https://user-images.githubusercontent.com/112706/235512978-8d14b991-fa6e-406b-9499-36a84f26b275.png">

Neovim version:

```
:version
NVIM v0.9.0
Build type: Release
LuaJIT 2.1.0-beta3

   system vimrc file: "$VIM/sysinit.vim"
  fall-back for $VIM: "/opt/homebrew/Cellar/neovim/0.9.0/share/nvim"
```

No Neovim config has been changed prior to today or after this.

Expected behavior:
X and missing icons should exist

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - Fira Code Nerd Font and Fira Code Nerd Font Mono
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - Homebrew and manually download
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - iterm2
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   -  MacOS

One more screenshot from other missing icons

[...]

PR #1219, opened 2023-05-09
Improve custom patching

Finii:
#### Description

This changes how custom patching (i.e. with `--custom`) is handled.
When I tried to use the feature myself several shortcomings became obvious.

#### Requirements / Checklist

This is mostly triggered by @siduck 's problem

* https://github.com/ryanoasis/nerd-fonts/issues/1213#issuecomment-1539214446
* https://github.com/ryanoasis/nerd-fonts/issues/1190#issuecomment-1532632029
* #1207

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -898,7 +898,7 @@ class font_patcher:
         CUSTOM_ATTR = {
             # previous custom scaling => do not touch the icons
             # 'default': {'align': 'c', 'valign': '', 'stretch': '', 'params': {}}
-            'default': {'align': 'c', 'valign': 'c', 'stretch': 'pa', 'params': {}}
+            'default': {'align': 'c', 'valign': 'c', 'stretch': 'pa', 'params': {'careful': self.args.careful}}
         }
 
         # Most glyphs we want to maximize (individually) during the scale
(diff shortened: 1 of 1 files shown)


siduck:
@Finii  thanks for adding file-dir icon back man :D

Finii:
This is so utterly disappointing. Create your own set of fonts, but call them differently. Maybe `NvchatFont` or whatever, but DO NOT publish packages with tampered-with `Nerd Fonts` :-(

Spent several hours on you problem and this is how you thank me.

siduck:
> > nvchad users probably and they'd have to do the same
> 
> Why on earth should they want an icon that has been dropped from Octo years ago and does not fit nicely with all others roll_eyes

it looks better imo and the users are already complaining when their fav folder icon got removed! And that icon does matches with other folder related icons we use ( cod* ), it looks like a filled version of the cod-folder, thats why I like it 

image

I wouldnt be bothered much if I was the only one using this icon, it has become a part of our UI since the beginning and made our folders unique ( similar to doom emacs filetree icons ) 

I really appreciate your work tho, no need to get mad! I never meant to be rude / disrespectful in our conversations. Just wanted to bring that icon back on a large scale

siduck:
> The old icon could only move into `nf-custom-` which is where we preserved some other old ones, but that means a codepoint change.

a codepoint change wouldnt do any harm as icon's still preserved!

siduck:
> What will happen when Microsoft updates its Codicons? At the moment they kept the symbols, but what if. Other icon changes will occur in the future with modernization I'm sure.

we use those folders only at few places and they're rarely visible, so not much of big deal. but that file-directory icon was our main one, we used it at the statusline etc too

Finii:
Found other bugs while moving the icon to `custom` ... :roll_eyes: 
Some `custom` icons are ~~broken~~ slightly deformed ?!

I fear it (the folder icon) will be patched a bit more to the right, because the custom icons are all centered (on something) while the octicons were left aligned. Need to check how much.

Finii:
Hmm, it is still left aligned but a bit smaller :thinking: 

image

siduck:
>  custom icons are all centered 

didnt know! I thought only the fonts that end with "Mono" are centered :thinking:

siduck:
> Hmm, it is still left aligned but a bit smaller thinking
> 
> image

oops looks a bit smaller  :fearful:

PR #1222, opened 2023-05-10
Fix the REGULAR bit of the fsSelection being set incorrectly

negset:
#### Description

This commit resolves the issue of the fsSelection of the OS/2 table not being properly set.

According to [Microsoft's documentation](https://learn.microsoft.com/en-us/typography/opentype/spec/os2#fsselection), bit 6 (REGULAR) of fsSelection is recommended to be set only when bit 0 (ITALIC) and bit 5 (BOLD) are clear.
However, the current script sets bit 6 (REGULAR) without following this rule (for example, when the font name is `XXX-BoldItalic.ttf`).

So, I have made changes to ensure that bit 6 (REGULAR) is not set when bit 0 (ITALIC), bit 5 (BOLD), or bit 9 (OBLIQUE) is set.
While the documentation does not specify whether bit 6 (REGULAR) should be cleared when bit 9 (OBLIQUE) is set, I think it should be treated like 0 (ITALIC) and bit 5 (BOLD).

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/name_parser/FontnameParser.py
+++ b/bin/scripts/name_parser/FontnameParser.py
@@ -255,7 +255,7 @@ class FontnameParser:
         elif 'Italic' in self.style_token:
             b |= ITALIC
         # Regular is just the basic weight
-        if len(self.weight_token) == 0:
+        if len(self.weight_token) == 0 and not b & (ITALIC | BOLD | OBLIQUE):
             b |= REGULAR
         b |= WWS # We assert this by our naming process
         return b


Finii:
Thanks for the excellent catch!

What do you think about this change instead:

```diff
-if len(self.weight_token) == 0 and not b & (ITALIC | BOLD | OBLIQUE):
+if len(self.weight_token) == 0 and not b & (ITALIC | BOLD):
```

As `OBLIQUE` is already handled by the (non) empty weight_token. But maybe your code is more clear.
You decide.

Checking if `--has-no-italic` is handled correctly...
Ah yes, that is handled only later in the concrete name-string generation.

Finii:
Interestingly Bit 9 (`Oblique`) is not mentioned :thinking: 

> Bit 6: If bit 6 is set, then bits 0 and 5 must be clear, else the behavior is undefined. Note that, if both bit 0 and bit 5 are clear, that does not give any indication as to whether or not bit 6 will be clear. For example, Arial Light is not the regular style of Arial and would have all bits cleared. 

Anyhow. After a :coffee: I believe your redundant but more explicit code is better :+1:

Finii:
Thanks again!

I will merge this together with the other 3.0.1 patches.

Finii:
@allcontributors add @negset for code

negset:
Thank you for merging my pull request!

PR #1227, opened 2023-05-10
Add old Octicons folder-icon

Finii:
#### Description

#### Requirements / Checklist


Finii:
When the CI build the original-source we get some defects, it is unclear why.
Running locally does not have these.
Needs to be checked out

image

Finii:
I think I give up on the CI; tried a lot of things to get the correct output but the CI-created original-source font is always 'broken' while it is ok running the same script with the same tools locally.

This takes too much time now, maybe fix 'another day'.

Finii:
Documenting what the CI runs always get wrong ...:

image

The left two have minuscule differences, but the right two are really noticeable.

When the CI should be re-activated this **needs to be checked carefully!**

Tried
* Running `fontforge` instead of `fontforge-nox`
* Running script directly and indirectly
* Running with `fontforge` AppImage
* Doing explicit `glyph.importOutlines(filename, simplify=False)`
* Also did try a bit with `default_joinlimit=...` 

All to no avail. The locally run looks different (correct) while the CI run always has the strange moved points.

Finii:
@siduck FYI

Also #1207, note codepoint transition.

siduck:
@Finii thanks a lot man!! Really appreciate it, lemme know if you need any sorts of help any time :)

Issue #1229, opened 2023-05-11
Enlarged glphys in FiraCode Mono


thpani:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
After the v3.0.0 release, the `Mono` variant of `FiraCode` contains enlarged glyphs at codepoints U+276C to U+276F, compared to non-`Mono` variant and parent FiraCode 6.2.

Expected behavior:
Equal glyphs in all 3 fonts, as before v3.0.0

Example symbols:
U+276C to U+276F: ❬❭❮❯

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - `FiraCodeNerdFontMono-Regular.ttf`
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - Homebrew `font-fira-code-nerd-font: 3.0.0`
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - `kitty`, macOS Terminal, also occurs in macOS TextEdit
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - macOS Ventura 13.3.1

#### ★ Screenshots (Optional)

<img width="374" alt="Screenshot 2023-05-11 at 11 29 00" src="https://github.com/ryanoasis/nerd-fonts/assets/82047/75db1555-0d7d-448d-ba70-682ecbd0a01e">

Finii:
Fira Code 6.2 does **not** contain `276C` - `276F` (but just `2770` and `2771`):

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/65e24cc5-0845-4641-b1cb-34d3b52372c9)

The glyphs you see there are inplace replacements via fontconfig.

They are patched in with these settings

```c
{
   'Enabled': True,                                
   'Name': "Heavy Angle Brackets",         
   'Filename': "extraglyphs.sfd",                                     
   'Exact': True,       
   'SymStart': 0x276C,      
[...]

PR #1232, opened 2023-05-11
font-patcher: Do not maximize heavy brackets in Mono

Finii:
#### font-patcher: Do not maximize heavy brackets in Mono
    
**[why]**
The added heavy brackets are maximized within the cell size as all
normal other symbols. But in fact they should not be maximized but
rather be the size of 'normal brackets'.
    
Furthermore they are all scaled individually, making the size
differences less.
    
With some proportional fonts the brackets look tiny.
    
**[how]**
Introduce new y-padding parameter (because a negative overlap also acts
in x direction, what we do not want (and it distorts)).
    
Pad the brackets with 30% (15% top and 15% bottom).
    
This is used for all fonts (monospaced or not) so that the new glyphs
fit nicely with the existing ones. For some definitions of 'nicely', but
that is as good as we can get with automatism. It's not worse than
font-fallback.
    
Fixes: #1229

#### Also done

Fixes some bug fixed along the way :grimacing: 

#### Requirements / Checklist

Insert the heavy brackets more carefully (size wise).

Use the new `generate-sample-set.sh` and then `fontforge` with font-compare.


Finii:
image

Finii:
> This has only effect in Nerd Font Mono fonts.

Hmm, changed the rule to `pa1!` (with `!`), so all fonts are affected, correcting commit msg

PR #1251, opened 2023-05-20
Feature/construct patch metadata

sapient-cogbag:
#### Description
Make `font-patcher` automatically generate per-font-file metadata about patched codepoints.

#### Requirements / Checklist

Make `font-patcher` produce program-ingestable JSON in `<font-name>.patchmeta.json`, alongside the produced font files containing the various patched symbol ranges that were applied to the original font - organised in batches by the name of the patch set/symbol font. 

Run the program and ensure it works with CI. I've done it manually on a couple of fonts, but I don't have enough internet speed to download even `--depth=1` in reasonable time because that includes all of the fonts, so I can't do the final step of checking  actual CI (and the build process is quite slow on my machine).

I originally came up with this idea because of my desire to add some mechanism to generate the unicode ranges for `symbol_map` in my `kitty` terminal `kitty.conf`. In particular, by generating an include-able file from some kind of metadata associated with the symbol-only font in the Arch Linux packaging process. 

However, I realised that metadata was not programmatically accessible, so  I decided to implement a method of generating JSON directly in the nerd-font build process. I think this way, it could also provide a unified access mechanism for other parts of the build and test pipeline, to avoid repetition of all the codepoint ranges.

N/A

N/A

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -1266,6 +1266,15 @@
       "contributions": [
         "doc"
       ]
+    },
+    {
+      "login": "sapient-cogbag",
+      "name": "sapient_cogbag",
+      "avatar_url": "https://avatars.githubusercontent.com/u/4654798?v=4",
+      "profile": "https://github.com/sapient-cogbag",
+      "contributions": [
+        "code"
+      ]
     }
   ],
   "repoType": "github",
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -186,6 +186,7 @@ Thanks goes to these wonderful people ([emoji key](https://github.com/kentcdodds
       <td align="center" valign="top" width="14.28%"><a href="http://maazu.xyz"><img src="https://avatars.githubusercontent.com/u/80816622?v=4?s=100" width="100px;" alt="Maaz"/><br /><sub><b>Maaz</b></sub></a><br /><a href="https://github.com/ryanoasis/nerd-fonts/commits?author=mmaaaaz" title="Documentation">📖</a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://negset.com"><img src="https://avatars.githubusercontent.com/u/15167516?v=4?s=100" width="100px;" alt="negset"/><br /><sub><b>negset</b></sub></a><br /><a href="https://github.com/ryanoasis/nerd-fonts/commits?author=negset" title="Code">💻</a></td>
       <td align="center" valign="top" width="14.28%"><a href="https://github.com/airt"><img src="https://avatars.githubusercontent.com/u/5058439?v=4?s=100" width="100px;" alt="airt"/><br /><sub><b>airt</b></sub></a><br /><a href="https://github.com/ryanoasis/nerd-fonts/commits?author=airt" title="Documentation">📖</a></td>
+      <td align="center" valign="top" width="14.28%"><a href="https://github.com/sapient-cogbag"><img src="https://avatars.githubusercontent.com/u/4654798?v=4?s=100" width="100px;" alt="sapient_cogbag"/><br /><sub><b>sapient_cogbag</b></sub></a><br /><a href="https://github.com/ryanoasis/nerd-fonts/commits?author=sapient-cogbag" title="Code">💻</a></td>
     </tr>
   </tbody>
 </table>
--- a/font-patcher
+++ b/font-patcher
@@ -313,6 +313,65 @@ def create_filename(fonts):
         sfnt_psubfam = '-' + sfnt_psubfam
     return (sfnt_pfam + sfnt_psubfam).replace(' ', '')
 
+class patched_region: 
+    # the source font start is the location in the font we start patching at, inserting glyphs
+    # taken from SymStart to SymEnd.
+    # If not specified, it just patches from the symbol font start location SymStart :)
+    def __init__(self, symbol_font_start, symbol_font_end, source_font_start, is_exact_range):
+        self.patch_start = symbol_font_start # int
+        self.patch_end = symbol_font_end # int
+        self.is_exact_range = is_exact_range # bool
+        if source_font_start is not None:
+            # Calculate the shift from symbol font to source font.  
+            delta = source_font_start - symbol_font_start
+            # Apply the shift, to calculate the actual location that the replaced glyphs are at
+            self.patch_start += delta
+            self.patch_end += delta
+
+        # Calculate a convenient hex value for the start and end of the patched region ^.^
+        # Most things reading unicode codepoints take them as hex with a prefix, so this is
+        # useful. 
+        self.hex_start = format(self.patch_start, "04X")
+        self.hex_end = format(self.patch_end, "04X")
+
+    # Return an object that can be encoded with python's json mechanism (i.e. a dict)
+    def as_json_encodeable(self):
+        return {
+            "patch-start": self.patch_start,
+            "patch-end": self.patch_end,
+            "hex-start": self.hex_start,
+            "hex-end": self.hex_end,
+            "is-exact-range": self.is_exact_range
+        } 
+
+# Metadata for a given patched font file.
+# Right now, it only outputs the patched codepoints categorized by name,
+# but we put all that data inside an outer JSON object with a specific key, so 
+# other information can be added later if desired without breaking compat :)
+class patching_metadata:
+    def __init__(self):
+        self.applied_patches = {} # class 'dict'
+
+    # Claim that a new patch has been added to the font. 
+    # The argument is a dictionary as in the format of font_patcher.patch_set elements. 
+    # This does not check if the patch is enabled - you need to do that or only insert 
+    # patches that are enabled, if you want to do things that way. 
+    def inject_new_patch(self, patch):
+        name = patch['Name']
+        new_region = patched_region(patch['SymStart'], patch['SymEnd'], patch['SrcStart'], patch['Exact'])
+        if name not in self.applied_patches:
+            self.applied_patches[name] = []
+        # Group patches sensibly
+        self.applied_patches[name].append(new_region)
+
+    # Produce json of the list of patches 
+    def emit_json_string(self):
+        json_structure = {}
+
+        json_structure["codepoints"] = {}
+        for patch_name, patched_regions in self.applied_patches.items():
+            json_structure["codepoints"][patch_name] = [region.as_json_encodeable() for region in patched_regions]
+        return json.dumps(json_structure)
 
 class font_patcher:
     def __init__(self, args):
@@ -329,6 +388,7 @@ class font_patcher:
         self.essential = set()
         self.config = configparser.ConfigParser(empty_lines_in_values=False, allow_no_value=True)
         self.xavgwidth = [] # list of ints
+        self.patch_metadata = None # class patched_symbols 
 
     def patch(self, font):
         self.sourceFont = font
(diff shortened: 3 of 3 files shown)


Finii:
Thanks for the PR! 

I'm camping right now so I can not really check it out. But related is maybe the generated https://github.com/ryanoasis/nerd-fonts/blob/master/glyphnames.json
Did you consider that and what were its drawbacks?

sapient-cogbag:
I don't think I saw that when I initially looked. However, as far as I can tell it's a fixed, existing list, which doesn't work so well if people are custom-selecting which symbols to patch in to a font.

It also doesn't embed ranges, which means end users have to recalculate them from the symbols, and it (at least from what I can see) is manually modified/updated, which means it can get out of sync. Also, it merges the symbol font name (and an aliased version) with the character name inside a single JSON key, which makes for more difficult parsing with automated tools like `jq`, and it lacks the name of the actual produced font (because it is not font-specific), as well as the name of the symbol font it pulled from.

By producing JSON with this information embedded directly, it makes for a much easier time for automated tools to pull this information without having to reparse the font files (requiring libraries)  or hardcode the information (e.g. a map of aliases for symbol fonts or hardcoding the produced font name), and avoids duplication and information getting out of date, and lots of effort spent ensuring that codepoints are updated everywhere when updates occur. 

The logical end point would actually be use the raw `glyphnames.json` as a source of truth specifically for per-character name information, and then embed the extracted information in the per-font generated JSON alongside the range information. 

However, people doing custom symbol fonts or modifying ranges would probably need to discard the name information, but the ranges as derived from the patchset being produced directly alongside the font would always be available and correct. 

I also realised two issues with my PR (why I switched it back to draft), one of which I can fix easily, but the other which I can only tentatively fix since I have trouble with the paradigm used to design it, though I could design a partial solution:
- The `copy_glyphs` function seems to do filling/packing behaviour and this means I need to modify my code to account for skipped glyphs. But honestly I have trouble understanding this code - I'm very unfamiliar with fontforge and the whole "selection" mechanism with copies and pastes that seem to be part of the actual font datatypes (as opposed to a separate function) is confusing. Is there some kind of global clipboard?
  - It would be helpful if someone were to separate out the position calculation logic, and copying logic, from all the stuff about making glyphs aligned with the surrounding font, which I really don't understand unfortunately and having it mixed in with the positioning stuff makes it hard for me to figure it out. But this might be related to my serious unfamiliarity with fontforge ;p
- The metadata currently produced doesn't correctly account for multifont output files. This one is easy to fix by creating separate metadata for each font and then producing a JSON map with each font name as a key and the associated patch metadata as it's value. This has a bonus of actually embedding the real system font names of the produced fonts where any program or automated tool can get to them.

Finii:
> I don't think I saw that when I initially looked. However, as far as I can tell it's a fixed, existing list, which doesn't work so well if people are custom-selecting which symbols to patch in to a font.

Are there really people who do that?

> It also doesn't embed ranges, which means end users have to recalculate them from the symbols,

Well, ranges is hard because some sets are rather scattered all over. Some are compressed, some are more scattered after patching in ;-)

> and it (at least from what I can see) is manually modified/updated, which means it can get out of sync.

This is only half true. It is generated from the `i*.sh` files, which are partially autogenerated. It is true it can get out of sync, but if it does the cheat sheet (which is also based in that data) would be broken; so the confidence that the `i*.sh` are correct is rather high.

> Also, it merges the symbol font name (and an aliased version) with the character name inside a single JSON key, which makes for more difficult parsing with automated tools like `jq`

Well, it contains the CSS entity name of the glyph. That is composed by a Nerd Font prefix, the source, and the name (because the name is not unique over multiple sets).

> and it lacks the name of the actual produced font (because it is not font-specific)

Of course ;)

> as well as the name of the symbol font it pulled from

That is an implementation detail and should not matter users consuming the patched fonts. They might be interested in the set's name, but the filename can change and has changed in the past.

> By producing JSON with this information embedded directly, it makes for a much easier time for automated tools to pull this information without having to reparse the font files (requiring libraries) or hardcode the information (e.g. a map of aliases for symbol fonts or hardcoding the produced font name), and avoids duplication and information getting out of date, and lots of effort spent ensuring that codepoints are updated everywhere when updates occur.

I do not understand why someone would need to reparse the font files?

> The logical end point would actually be use the raw `glyphnames.json` as a source of truth specifically for per-character name information, and then embed the extracted information in the per-font generated JSON alongside the range information.

I'm not sure I get the advantage here. And 'range' is kind of a mirage I guess. A glyph can belong to a set (setname), and that set can contain several ranges. I'm not sure 'range' is leading anywhere.

> However, people doing custom symbol fonts or modifying ranges would probably need to discard the name information, but the ranges as derived from the patchset being produced directly alongside the font would always be available and correct.

The  idea you propose here, generating the `json` file from the actual patching process, is a nice idea.
I guess all the workflow that now revolves around the `i*.sh` files can be rewritten to account for that and take the `json` file instead. That will at least automate the translation process - the 'original data' will always be the `patch_set` in the `font-patcher` (well, unless we completely rewrite that also).

> The `copy_glyphs` function seems to do filling/packing behaviour and this means I need to modify my code to account for skipped glyphs. But honestly I have trouble understanding this code

The `patch_set` has these members per entry:
* `Exact`: Keep gaps in the symbol font or (if this is false) skip over empty symbol slots
* `SymStart`: Start this set at this unicode codepoint in the symbol font
* `SymEnd`: End this set at this unicode codepoint in the symbol font (including)
* `SrcStart`: Patch the glyphs in starting at this unicode codepoint in the destination font

Symbolic code would be like this
```
destination_codepoint = SrcStart
For symbol_codepoint = SymStart .. SymEnd
    if symbol_codepoint is empty
        if not Exact
            continue
    else
        put symbol_codepoint into destination_codepoint
    destination_codepoint++
```

> I'm very unfamiliar with fontforge and the whole "selection" mechanism with copies and pastes that seem to be part of the actual font datatypes (as opposed to a separate function) is confusing. Is there some kind of global clipboard?

Yes, snippet of the actual code:
```python
        symbolFont.selection.select(sym_glyph.encoding)
        symbolFont.copy()

        # Paste it
        self.sourceFont.selection.select(currentSourceFontGlyph)
        self.sourceFont.paste()
```
Each font object has a `selection` property, where one can select one or more glyphs with.
`copy()` copies the selected glyph(s) into the global buffer, `paste()` pastes them into the selected glyph slots.
In the code above it selects only one glyph (per font).

> It would be helpful if someone were to separate out the position calculation logic, and copying logic, from all the stuff about making glyphs aligned with the surrounding font, which I really don't understand

Unfortunately that is interrelated very much, because copy-and-pasting the glyphs means the glyphs are translated to a different coordinate system where rounding errors might occur, and there is a considerable amount of code just to prevent the rounding errors breaking the results :grimacing: In fact that is the most complicated part of it all.

> The metadata currently produced doesn't correctly account for multifont output files. This one is easy to fix by creating separate metadata for each font

Well, normally all files in the container are patched the same, so there is no need for that. Except when one specifies `--careful` and the filled glyphs differ per subfont. Which is a rather specialized (uncommon) scenario I guess.

----

In summary I still do not really know what is missing with the already existing `glyphnames.json` file. If you want to have set-names and or names-without-set as fields, that can be easily added.

I also believe generating the `glyphnames` from one `font-patcher` run is a good idea. These `i*.sh` files are not the best to handle this thing. Otoh changes to the `i*.sh` are very rare for the symbolfonts where it is not automated ;)

I would propose that we
* find how to extend `glyphnames` that it is useful for you
* autogen the `glyphnames` from one patcher run (e.g. on SymbolsOnly)
* drop the `i*.sh` and and use the `glyphnames` in their stead

Finii:
Lets see what the PT generates

```console
$ fontforge ../../font-patcher --complete ../../src/unpatched-fonts/Agave/Agave-Regular.ttf
$ cat AgaveNerdFont-Regular.ttf.patchmeta.json | jq
```

```json
{
  "codepoints": {
    "Seti-UI + Custom": [
      {
        "patch-start": 58874,
        "patch-end": 59135,
        "hex-start": "E5FA",
        "hex-end": "E6FF",
        "is-exact-range": false
      }
    ],
    "Heavy Angle Brackets": [
      {
        "patch-start": 10092,
        "patch-end": 10097,
        "hex-start": "276C",
        "hex-end": "2771",
        "is-exact-range": true
      }
    ],
    "Devicons": [
      {
        "patch-start": 59136,
        "patch-end": 59333,
        "hex-start": "E700",
        "hex-end": "E7C5",
        "is-exact-range": false
      }
    ],
    "Powerline Symbols": [
      {
        "patch-start": 57504,
        "patch-end": 57506,
        "hex-start": "E0A0",
        "hex-end": "E0A2",
        "is-exact-range": true
      },
      {
        "patch-start": 57520,
        "patch-end": 57523,
        "hex-start": "E0B0",
        "hex-end": "E0B3",
        "is-exact-range": true
      }
    ],
    "Powerline Extra Symbols": [
      {
        "patch-start": 57507,
        "patch-end": 57507,
        "hex-start": "E0A3",
        "hex-end": "E0A3",
        "is-exact-range": true
      },
      {
        "patch-start": 57524,
        "patch-end": 57544,
        "hex-start": "E0B4",
        "hex-end": "E0C8",
        "is-exact-range": true
      },
      {
        "patch-start": 57546,
        "patch-end": 57546,
        "hex-start": "E0CA",
        "hex-end": "E0CA",
        "is-exact-range": true
      },
      {
        "patch-start": 57548,
        "patch-end": 57556,
        "hex-start": "E0CC",
        "hex-end": "E0D4",
        "is-exact-range": true
      },
      {
        "patch-start": 9776,
        "patch-end": 9776,
        "hex-start": "2630",
        "hex-end": "2630",
        "is-exact-range": true
      }
    ],
    "Pomicons": [
      {
        "patch-start": 57344,
        "patch-end": 57354,
        "hex-start": "E000",
        "hex-end": "E00A",
        "is-exact-range": true
      }
    ],
    "Font Awesome": [
      {
        "patch-start": 61440,
        "patch-end": 62176,
        "hex-start": "F000",
        "hex-end": "F2E0",
        "is-exact-range": true
      }
    ],
    "Font Awesome Extension": [
      {
        "patch-start": 57856,
        "patch-end": 58025,
        "hex-start": "E200",
        "hex-end": "E2A9",
        "is-exact-range": false
      }
    ],
    "Power Symbols": [
      {
        "patch-start": 9211,
        "patch-end": 9214,
        "hex-start": "23FB",
        "hex-end": "23FE",
        "is-exact-range": true
      },
      {
        "patch-start": 11096,
        "patch-end": 11096,
        "hex-start": "2B58",
        "hex-end": "2B58",
        "is-exact-range": true
      }
    ],
    "Material": [
      {
        "patch-start": 983041,
        "patch-end": 989936,
        "hex-start": "F0001",
        "hex-end": "F1AF0",
        "is-exact-range": true
      }
    ],
    "Weather Icons": [
      {
        "patch-start": 58112,
        "patch-end": 58347,
        "hex-start": "E300",
        "hex-end": "E3EB",
        "is-exact-range": false
      }
    ],
    "Font Logos": [
      {
        "patch-start": 62208,
        "patch-end": 62255,
        "hex-start": "F300",
        "hex-end": "F32F",
        "is-exact-range": true
      }
    ],
    "Octicons": [
      {
        "patch-start": 62464,
        "patch-end": 62725,
        "hex-start": "F400",
        "hex-end": "F505",
        "is-exact-range": false
      },
      {
        "patch-start": 9829,
        "patch-end": 9829,
        "hex-start": "2665",
        "hex-end": "2665",
        "is-exact-range": true
      },
      {
        "patch-start": 9889,
        "patch-end": 9889,
        "hex-start": "26A1",
        "hex-end": "26A1",
        "is-exact-range": true
      },
      {
        "patch-start": 62633,
        "patch-end": 62771,
        "hex-start": "F4A9",
        "hex-end": "F533",
        "is-exact-range": false
      }
    ],
    "Codicons": [
      {
        "patch-start": 60000,
        "patch-end": 60395,
        "hex-start": "EA60",
        "hex-end": "EBEB",
        "is-exact-range": true
      }
    ]
  }
}
```

Finii:
Ok. The code looks very nice and is even documented ;-)

And the file contains not what I expected from reading the description. Maybe my comments above are all obsolete ;-)

But these questions remain:

* Do you really think self-patching un-complete is a common case and needs support?
(i.e. is one repo-global file not enough)
* This could be put into `glyphnames`'s 'metadata' field
* I believe this information is used nowhere in Nerd Fonts' own processes, well, it is needed but with higher resolution i.e. for each glyph, to create `glyphnames` and the cheat sheet

Hmm. :thinking:

Finii:
If `careful` is active, either 
* through a user supplied `--careful`
* because of a `careful` attribute
* because the target glyph is 'essential'

it will not be exempt from the ranges list. If you want to handle this font-individually, I would expect that, maybe.

What is the use-case for the ranges database? I guess I have still not understood what it is useful for.

Finii:
Closing as this seems stuck.
Moving to parallel universe to not loose the idea(s).

PR #1252, opened 2023-05-20
website: improve cheat-sheet page performance

rszyma:
#### Description

Update cheat-sheet page for better performance

#### Requirements / Checklist

##### Modifications to cheat-sheet page:
- Improved page load time
- Greatly improved search time
    - Use [MiniSearch](https://lucaong.github.io/minisearch/) lib as search engine
    - Lazy render search results. Currently up to 250 entries at a time. More entries loaded if scrolled to bottom.
- Show all icons on page load.
- Empty search field shows all icons.
- Removed "Search" and "Show All Icons" buttons, since they are would be useless anyway.

##### Other
- Remove "sidecar", because it's not working anymore.

manually

Main motivation for these changes is terrible UX/performance of https://www.nerdfonts.com/cheat-sheet

There are no tickets for neither bad cheat-sheet performance, nor sidecar not working.


rszyma:
One thing though: "removed" icons are now all over the search results, and not at the end as before. However, IMHO it's nothing compared to improvements this PR brings (it's definitely fixable, but I don't have that much time to think of the best way to do this right now).

Finii:
Thanks for the PR!

Camping right now, so I can't check it out right now. But wanted to comment that 'removed comes last' was not intended :-D

Finii:
Further note: `_posts/2017-01-04*. md` is autogenerated. I guess you replaced it by `glyphs.html`. Can not see that on the phone, but propably it needs to be regererated on glyph set changes. Or you read the data or css file.
If it happens to need updates itself, do you have a generator script?

See

* https://github.com/ryanoasis/nerd-fonts/tree/master/bin/scripts#readme
* https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/generate-css.sh

rszyma:
Yeah, I didn't know it was autogenerated. Maybe we should add a disclaimer like "This file was autogenerated by <scriptname>. Do not edit." at the beginning of the autogenerated files so it's more obvious? 

Your guess about glyphs.html was right. I moved the cheatsheet data to `glyphs.html`. However I converted it to json for smaller size and faster loading time. Also MiniSearch library I've used as search engine needs json data, so converting to json was obvious choice.

Currently `glyphs.html` looks like this
```html
<script>
const glyphs = {
  "nf-cod-account": "eb99",
  "nf-cod-activate_breakpoints": "ea97",
  "nf-cod-add": "ea60",
  "nf-cod-archive": "ea98",
  "nf-cod-arrow_both": "ea99",
  ...
}
</script>
```

rszyma:
> Further note: `_posts/2017-01-04*. md` is autogenerated. I guess you replaced it by `glyphs.html`. Can not see that on the phone, but propably it needs to be regererated on glyph set changes. Or you read the data or css file. If it happens to need updates itself, do you have a generator script?
> 
> See
> 
> * https://github.com/ryanoasis/nerd-fonts/tree/master/bin/scripts#readme
> * https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/generate-css.sh

I've updated [generate-css.sh](https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/generate-css.sh) accordingly in #1254

Finii:
Still struggling to iron out bugs in 3.0.1 and prepare 3.0.2, please do not think this is forgotten.

A casual glance left the impression that you pull in the (in)famous NPM? How deep is that dependency well?
In other places we go to some length to reduce the dependencies; can we not improve it without opening another 100 dependencies? :grimacing:

Finii:
Cheery picked already

beb808ebbdb865015ef4de7f1e0d398e08b7b8a4 `Remove gitter-sidecar` as 34cddad56b7

Finii:
Started your branch locally, the search behavior is a bit strange? Is that full word search?

## alt

image

## alti

image

## altimete

image

## altimeter

image

rszyma:
It is full word search. While it works differently than before I'd argue it's better. 
Reasons:
- a lot faster
- don't show unrelated glyphs for short terms like "a"

If you don't like the full-word search, there are some options to tweak:
https://lucaong.github.io/minisearch/#search-options
- `prefix: true` - searching for "alti" would show results for "altimeter". But the search is a lot a lot slower.
- `fuzzy` - searches for misspelled words

There is also an [Auto Suggestion](https://lucaong.github.io/minisearch/#auto-suggestions) feature, you can see it in action [here](https://lucaong.github.io/minisearch/examples/)

rszyma:
> A casual glance left the impression that you pull in the (in)famous NPM? How deep is that dependency well? In other places we go to some length to reduce the dependencies; can we not improve it without opening another 100 dependencies? 😬

If you're talking about MiniSearch, it's downloaded from CDN. See here:
https://github.com/ryanoasis/nerd-fonts/pull/1252/files#diff-022702df0bec434e85269514ea8900f6f54e698486b484ad91aede3837cfb788R21
And it's only 29 KB in size,
with no external dependencies (or at least they claim that on their site).

rszyma:
I'll comment further on what is going on in this PR to make the review easier.

- Moved the code specific to cheat sheet from `site.js` to `cheat-sheet.js`. 

- I don't know anything about the `gtag` functions that are used throughout the code, so I just ignored them. You might want to add/remove/readjust some of them.

- I'm sorry about so many changes in `_includes/css/nerd-font-tweaks.scss`.
Seems like my autoformatter acted on it's own. All I intended to change there is to remove sidecar-related css and adjust (mostly remove) css to work well will changes made to search.

Finii:
@allcontributors please add @SilverMira for bug

Finii:
First of all, I finally spend some time really looking into the code here, and it is excellent! Thank you.

The following comments are out of date. Instead I pushed some commits. Please review them.

Performance is not the main goal for me, it is still far better than before. I believe the prefix search is needed for usability and the sorting is at least a nice-to-have. I have no means to benchmark that stuff apart from gut-value ;-)

----

> One thing though: "removed" icons are now all over the search results, and not at the end as before. However, IMHO it's nothing compared to improvements this PR brings (and it's definitely fixable, but I don't have that much time to think of the best way to do this right now).

Maybe you have time for a quick look at the following patch?
Surprisingly it does not always work as expected.

```diff
From ee38d0ac6869c8a28be07724675a2b7716072da4 Mon Sep 17 00:00:00 2001
From: Fini Jastrow <ulf.fini.jastrow@desy.de>
Date: Tue, 30 May 2023 11:45:01 +0200
Subject: [PATCH] cheat_sheet: Sort removed icons last

Signed-off-by: Fini Jastrow <ulf.fini.jastrow@desy.de>
---
 cheat-sheet.js | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git a/cheat-sheet.js b/cheat-sheet.js
index 144987e23..6b9561104 100644
--- a/cheat-sheet.js
+++ b/cheat-sheet.js
@@ -9,7 +9,7 @@ document.addEventListener('DOMContentLoaded', function () {
 
     // Index all glyphs/icons
     let miniSearch = new MiniSearch({
-        fields: ['id', 'code'], // fields to index for full-text search
+        fields: ['id', 'code', 'isNew'], // fields to index for full-text search
         storeFields: ['id', 'code', 'isRemoved'], // fields to return with search results
     })
     miniSearch.addAll(Object.entries(glyphs).map(
@@ -17,6 +17,7 @@ document.addEventListener('DOMContentLoaded', function () {
             return {
                 id: key,
                 code: value,
+                isNew: key.startsWith('nfold') ? false : key,
             }
         }
     ));
@@ -136,6 +137,7 @@ document.addEventListener('DOMContentLoaded', function () {
             {
                 prefix: false,
                 combineWith: "AND",
+                boost: { isNew: 2 },
             }
         );
 
-- 
2.39.2

```


> prefix: true - searching for "alti" would show results for "altimeter". But the search is a lot a lot slower.

I think this is needed at least. I also can not really see any speed impact?

The `isRemoved` is never filled?
I added this code:

```diff
     miniSearch.addAll(Object.entries(glyphs).map(
         ([key, value]) => {
             return {
                 id: key,
                 code: value,
+                isRemoved: key.startsWith('nfold'),
             }
         }
     ));
```

But then, I guess it can be removed as it is never used.

Finii:
> * I don't know anything about the `gtag` functions that are used throughout the code, so I just ignored them. You might want to add/remove/readjust some of them.

Well, I guess Ryan set up google analytics to have some insight in usage / usage-patterns.
I have no access to the results, so 'I dont care'.
Would not remove it either, `googletagmanager` is anyhow disabled in my browser ;-)

cheat-sheet.js
```diff
@@ -0,0 +1,268 @@
...
+            searchTerm = "nf";
+        }
+
+        // TODO: search suggestions
+
+        remainingSearchResults = miniSearch.search(searchTerm,
+            {
+                prefix: true,
+                combineWith: "AND",
+                boostDocument: ((documentId, term, storedFields) => {
```
Finii, line 137: The downside of the `boost` based sorting is ... that it follows documents text searcher rules.
(`file_code` in front of `account`)

Take for example this...

image

Because `eae9` has `cod` in the name twice it is listed too early (if you expect a sort-by-name list).

rszyma: > The downside of the boost based sorting is ... that it follows documents text searcher rules.

It's not a downside to boost based sorting, but the consequence of setting `prefix: true`

rszyma:
I changed boost for `id` field from 0.001 to 0.00001, because if you search for `fa` and scroll down you can see that the removed icons come before some of the not-removed icons. Changing the boost to even lower fixes that.

> I also tried a lot variants with boost, it did never work out. 

The variant you wanted is here: 09ad62bb98e9df270f975df10f0a1bf996c6d518
It can't be done with `boost` option. `boostDocument` is the option you wanted.

> What is wrong with just sorting? 

Probably nothing. I did the changes in 09ad62bb98e9df270f975df10f0a1bf996c6d518 because I found that the sorting code didn't really work. Also there's a chance that using `boostDocument` is more optimized than just sorting.

> Adding all the strings also costs memory, so it is hard to tell if sticking to boost has any upside.

I've just checked how much memory the loaded page takes and it's about 20 MB, so I'd say it's completely acceptable. The additional "name" field is used so to fix the issue related to setting `prefix: true` (on the screenshot) https://github.com/ryanoasis/nerd-fonts/pull/1252#discussion_r1210236748

> No real sort on startup
resolved with 61a9c4dd68b56fadc22c6ca0ccc38c4e88311e91

rszyma:
> Sorting still not according to id

The results look correct to me. 

All of the below names contain prefix "cod":

coda
codrops
codiepie
codeigniter
code

rszyma:
> Of course it can be done with just boost

I see! I thought it didn't work, so you changed it. My bad. 

> I dropped that for a regular custom search with
https://github.com/ryanoasis/nerd-fonts/commit/9d09f38febbd7d59f82dcf7c233aa16bb87665fe cheat-sheet: Sort the results

I see now why you changed to manual sort. As I said before, I changed the code to use `boostDocument` because the manual sorting attempt didn't work as expected:
image

> Anyhow, resorting an almost sorted array should be really cheap. I do not know which sort is used by JS, but all usual candidates do that in linear time, right

I don't know much about sorting in JS, so I'd rather believe minisort than myself to sort it for me (by using `boost` and `boostDocument`) :smile: 

> Right, they have the same RANK, but they are not sorted for their ID within the rank.
Obviously codiepie (with an i after cod) needs to come before codrops (which has an r after cod).
TBH, sounds like a non-issue to me. Maybe that could be changed with use of some of advanced parameters in minisearch, but I think it's not worth the trouble for now.

Finii:
> because the manual sorting code didn't work as expected

I need to check that out :-D
Somewhere I have a machine with jekyll, ... \*noise of searching through a bunch of laptops\*

Finii:
Hmm, at current commit...

Why is `nf-fa-fa` first? This minisearch's sort is rubbish :-D

image

Finii:
> I see now why you changed to manual sort. As I said before, I changed the code to use `boostDocument` because the manual sorting code didn't work as expected:
> image

At 9d09f38febbd7d59f82dcf7c233aa16bb87665fe (i.e. "manual search") I get this result...

image

rszyma:
> Is this ready for merging from your perspective?

I don't see any major issues with this PR right now, so yes.

> Do you have an opinion/solution for https://github.com/ryanoasis/nerd-fonts/pull/1265?

I'd suggest working on #1265 after this PR is merged. Worth noting, that #1265 breaks lazy-loading implemented in this PR.

> The coding style is nice

Thanks!

> Hope you do not mind this discussion

Not at all

> I probably would like to add some of the comments in this PR to the commit messages if that is ok for you.
(I like to have explanations and reasons in git rather than github)

Sure, go ahead.

> Why is nf-fa-fa first? This minisearch's sort is rubbish :-D

Honestly, no clue. Maybe it has something to do with the id having double 'f'? 
There's is [this](https://github.com/lucaong/minisearch/issues/144) issue discussing possible workarounds if one wanted to "fix" it. One suggestion is to have an another dummy field that will search for all results, but I'm sure you won't like it :smile: 

> Your image seems to be not from 'my' manual search that takes isRemoved into account?

I ran 9d09f38 again just now and it still doesn't work. Idk.

Finii:
Resolving some conflicts manually, force push

Finii:
image

Manual merge via commit to get PR number into git

Finii:
@allcontributors add @rszyma for code

Finii:
Changed/added some small things.
Recreated the cheat-sheat file via script to get it 100% as generated.

image

:heavy_check_mark: Works ;-)


Thanks for the great contribution!

Finii:
> I don't know much about sorting in JS, so I'd rather believe minisort than myself to sort it for me

minisort just uses ordinary sort, just on the weights [1] with a sort lambda [2] like this:


```javascript
const byScore = ({ score: a }: Scored, { score: b }: Scored) => b - a

...

search (query: Query, searchOptions: SearchOptions = {}): SearchResult[] {

    ...

    results.sort(byScore)
    return results
  }
```

Javascript seems to use [3].
minisearch does not have any magic here, this is 100% standard stuff.
If we assign no weights the sort is kind of a noop and we can later on sort in a manner that makes sense for the cheat-sheet.


[1] https://github.com/lucaong/minisearch/blob/93bb0bc1878a48fcf2cf64e17099ae5a7042f9bd/src/MiniSearch.ts#LL1216C9-L1216C9
[2] https://github.com/lucaong/minisearch/blob/93bb0bc1878a48fcf2cf64e17099ae5a7042f9bd/src/MiniSearch.ts#LL1934C1-L1934C1
[3] https://stackoverflow.com/questions/234683/javascript-array-sort-implementation/236534#236534

PR #1254, opened 2023-05-21
modify `generate-css.sh` for changes in #1252

rszyma:
#### Description

To be merged along with #1252

### DIFF
--- a/bin/scripts/data/cheatsheet-foot.txt
+++ b/bin/scripts/data/cheatsheet-foot.txt
@@ -1,4 +1,6 @@
-</div>
+}
+</script>
+
 <h3>Removed Icons</h3>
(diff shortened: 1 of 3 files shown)


Finii:
This is very good.

Well, you do not do what you yourself suggested, but that is not so important maybe:

> Maybe we should add a disclaimer like "This file was autogenerated by . Do not edit." at the beginning of the autogenerated files so it's more obvious?

I believe people who mess with the files would need to find out how this all interacts anyhow, and have a look in the pony `release.sh` script or the workflows. Or the READMEs.

Approve, will be pulled when the other changes become ready. :+1:

Finii:
Rebase on master, force push

THIS web-merge is still broken :-(, resorting to manual

Issue #1249, opened 2023-05-19
DejaVu has no Regular with Putty


Finii:
#### 🗹 Requirements

This issue was originally raised by liaoya (Huifeng Shen) on Gitter. Thank you.

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
Putty shows only Oblique, Bold Oblique and Bold for `DejaVuSansM Nerd Font Mono`
Also Bold Oblique is depicted as Bold

Expected behavior:

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_ `DejaVuSansM Nerd Font Mono`
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - www.nerdfonts.com
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - putty latest
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Win10

#### ★ Screenshots (Optional)

![grafik](https://github.com/ryanoasis/nerd-fonts/assets/16012374/a4c392d3-3b90-40c9-807e-68b13a5f36ad)

Finii:
Hmm, did we not prevent this?

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/5d4bb669-3a8c-479e-9f21-b03225e421f0)

Finii:
Font is patched with `--has-no-italic`, but `fsSelect` incorrect:

```console
$ find . -name '*.ttf' -and -name '*Mono-*' -print -exec bash -c 'ttfdump {} | grep fsS' \;
./Regular/DejaVuSansMNerdFontMono-Regular.ttf
	 fsSelection:		 0x01c0 	 'Regular '
./Bold/DejaVuSansMNerdFontMono-Bold.ttf
	 fsSelection:		 0x01a0 	 'Bold '
./Italic/DejaVuSansMNerdFontMono-Oblique.ttf
[...]

PR #1255, opened 2023-05-22
FontnameParser: Fix fsSelection for --has-no-italic

Finii:
**[why]**
For fonts that have no Italic but an Oblique - i.e. when Oblique shall replace the Italic role in RIBBI font grouping (classic group of 4) - that grouping fails.

This affects DejaVu on Putty.

**[how]**
For RIBBI grouping only the classic bits are considered. That means that for fonts that have Oblique instead of Italic (and not additionally) we need to set the ITALIC bit and the OBLIQUE bit. This has been overlooked.

Cite from the specs:

> This bit, unlike the ITALIC bit (bit 0), is not related to style-linking
> in applications that assume a four-member font-family model comprised
> of regular, italic, bold and bold italic. It may be set or unset
> independently of the ITALIC bit. In most cases, if OBLIQUE is set, then
> ITALIC will also be set, though this is not required.

Fixes: #1249

Reported-by: Huifeng Shen <liaoya@gmail.com>

#### Requirements / Checklist

Create new set of DejaVu, install NFM on Windows, start putty.

https://learn.microsoft.com/en-us/typography/opentype/spec/os2#fsselection

### DIFF
--- a/bin/scripts/name_parser/FontnameParser.py
+++ b/bin/scripts/name_parser/FontnameParser.py
@@ -252,6 +252,9 @@ class FontnameParser:
         # Ignore Italic if we have Oblique
         if 'Oblique' in self.weight_token:
             b |= OBLIQUE
+            if not self.rename_oblique:
+                # If we have no dedicated italic, than oblique = italic
+                b |= ITALIC
         elif 'Italic' in self.style_token:
             b |= ITALIC
         # Regular is just the basic weight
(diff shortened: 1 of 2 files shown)


Finii:
Fixed:

image

Thanks again @liaoya for reporting!

Finii:
Affected:

```console
$ grep -r --include '*.cfg' '.--has-no' .
./src/unpatched-fonts/DejaVuSansMono/config.cfg:config_patch_flags="--has-no-italic"
./src/unpatched-fonts/BitstreamVeraSansMono/config.cfg:config_patch_flags="--has-no-italic"
```

Issue #1240, opened 2023-05-14
\ue0b4 cuts half of the previous character


237dmitry:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

The last update to version 3.0.1 brought many unforgettable minutes of rewriting and debugging scripts. When will you finish experimenting with users?

The essence of the problem. Character number '\ue0b4' began to eat half of the previous character. This happens with any of the installed fonts.

Expected behavior:

Fonts should not be globally changed every minor release 

Example symbols:

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - _Please give the full filename_

```
AgaveNerdFont-Bold.ttf
AgaveNerdFont-Regular.ttf
DroidSansMNerdFont-Regular.otf
FantasqueSansMNerdFont-Bold.ttf
FantasqueSansMNerdFont-BoldItalic.ttf
FantasqueSansMNerdFont-Italic.ttf
FantasqueSansMNerdFont-Regular.ttf
TerminessNerdFont-Bold.ttf
TerminessNerdFont-BoldItalic.ttf
TerminessNerdFont-Italic.ttf
TerminessNerdFont-Regular.ttf
```
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - 
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - 

Tilix, xfce4-terminal

- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - 

Linux

#### ★ Screenshots (Optional)

![font](https://github.com/ryanoasis/nerd-fonts/assets/78153320/80d5dd3e-f535-41c6-baf9-a34629bae79d)

Finii:
Sorry to hear of your problem.

Does that also happen with a normal letter before `e0b4`?
The glyph looks good

[...]

PR #1261, opened 2023-05-26
Bugfix Codicon Circles

Finii:
#### Description

Revert scaling of (most of) the Codicons back to individual maximize as it has been with `v2.3` and before.

On the way a new Scaling method is introduced and removed again as it does technically solve the problem but probably people do not want relative consistency but big icons i.e. maximizing.

#### Requirements / Checklist

Fixes: #1240

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -922,11 +922,7 @@ class font_patcher:
         #       with the same scaling and shifting. The basis for it is a 'combined
         #       bounding box' of all glyphs in that group. All glyphs are handled as
         #       if they fill that combined bounding box.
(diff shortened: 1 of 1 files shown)


Finii:
Here examples of the various tries:

Original situation:

image

This looks reasonable for `NFM` but the left bearing of `circle-filled` looks strange for `NF`.
The reason is that the small and the large circle are in a `ScaleGroup` where shifting and scaling is done with the 'combined bounding box' in mind. So the centers of both circles are on top of each other in NFM **and** NF.

PR #1274, opened 2023-06-01
CI uses ShellCheck

Finii:
#### Description

Put `ShellCheck` in the CI, so that we can monitor the code quality of the shell scripts.
For the Python scripts we use CodeClimate already.

#### Requirements / Checklist

* Create a workflow that runs ShellCheck on all `*.sh` files.
* Fix all existing warnings in the scripts.

Most important check that all changed scripts still run (with same results).

@hasecilu 's comment https://github.com/ryanoasis/nerd-fonts/pull/1270#issuecomment-1569537525

See also https://github.com/Finii/nerd-fonts/actions/workflows/shellcheck.yml


Finii:
Hmm, why did it rebuild the original-source font?
That should not have happened :thinking:

Finii:
No **visible** change

image

Still the gradle has a bug - the eye is 'closed' - but that was always the case.

Issue #1276, opened 2023-06-01
Docs: localized readmes' TOC don't work


hasecilu:
#### 🗹 Requirements

- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [X] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [X] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

It seems that the `readme` files relies on the GitHub's markdown flavor to make the links between the TOC and the headers and it works on the main `readme` but in the localized files this didn´t work.  
The reason is because the tags on the links at the TOC looks for the tags generated in the original file (English), since the headers are translated the tag is not found.

Expected behavior:

When reading the localized readmes the users should be able to travel across the file with no problems using the links on the TOC but when you click any element of the TOC nothing happens.

If instead of relying on GitHub's flavor we use html `<a name="tag-name">` tags on the headers all readme files can use links.

hasecilu:
btw, I'm working on the `readme_es.md` file, using the same tags used on the original `readme` will allow users to navigate across the localized readmes with no problem.

Finii:
Well, it does work on some but not on all translated readmes. I remember fixing some :roll_eyes:

hasecilu:
I have realized GitHub has this "new" feature that let you invoke a TOC while navigating available in the main readme.

![Screenshot from 2023-06-02 13-27-24](https://github.com/ryanoasis/nerd-fonts/assets/53124818/38f07025-bca6-451b-b8aa-eebe98104598)

And with the localized readmes there is a `Outline` button that shows all tags, very convenient.  
Since in the localized files the TOC didn't work I think we should removed it.
![Screenshot from 2023-06-02 13-28-13](https://github.com/ryanoasis/nerd-fonts/assets/53124818/f20c354f-76e3-4534-9aea-492cf30f0a0d)

PR #1277, opened 2023-06-01
Add html `<a name="tag-name">` like tags on headers.

hasecilu:
#### Description

_Make links work using html tags instead of relying on GitHub's flavor._

#### Requirements / Checklist

Add html tags on markdown headers.
I used this script:
```bash
#!/bin/bash

filename="readme.md"

[ -f "$filename".new.md ] && rm "$filename".new.md

while IFS= read -r line; do
	header=$(echo "$line" | grep ^\# | tr -cd '[:alnum:][:space:]\n' | sed 's/ /-/g' |
		tr '[:upper:]' '[:lower:]' | sed 's/^-//')
	[ ! "$header" == "" ] && echo -e "$line\n$line <a name=\"$header\"></a>" # Terminal
	[ ! "$header" == "" ] && echo "$line <a name=\"$header\"></a>" >>"$filename".new.md || 
		echo "$line" >>"$filename".new.md
done <"$filename"

```
Open the `readme.md` file and click on TOC elements, you should note that links still work.  
I checked the links and seem to work fine.
The links on the localized `readmes` don't work because the headers are translated, the tags reamin the same (in English) and the header link is broken.
#1276 

Edit: It works on most offline markdown file readers/editors and GitLab but not here on GitHub.
Edit2: Basically these changes have not effect **on GitHub**: the main readme's links still works and the **localized readme's** links still don't work. But the changes make the localized TOC links work on a local md editor (I used VNote) so it's only half progress.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -26,12 +26,12 @@ The following Sankey flow diagram shows the current glyph sets included:
 <sub><i>Diagram created using <a href="http://sankeymatic.com/" title="SankeyMATIC (BETA): A Sankey diagram builder for everyone">@SankeyMATIC</a></i></sub>
 
 
-## Important Notices
+## Important Notices <a name="important-notices"></a>
 * `master` branch file paths are **not** considered stable. [Verify your repository URI references](#unstable-file-paths)
 * cloning this repository is **not** recommended ([due to Repo size](#option-5-clone-the-repo)) unless you are going to be [contributing to development](#contributing)
 
 
-## Table of Contents
+## Table of Contents <a name="table-of-contents"></a>
 
 [**TL;DR**](#tldr)
 
@@ -66,12 +66,12 @@ The following Sankey flow diagram shows the current glyph sets included:
   * [**License**](#license)
 
 
-## TL;DR
+## TL;DR <a name="tldr"></a>
   Nerd Fonts takes popular programming fonts and adds a bunch of Glyphs.
   There is also a [font patcher](#font-patcher) available if your desired font isn't already patched.
   For more high level information see the [wiki][wiki]. If you are looking for the Vim plugin see [VimDevIcons ➶][vim-devicons].
 
-### Various Download Options for Fonts
+### Various Download Options for Fonts <a name="various-download-options-for-fonts"></a>
 
 _If you..._
 
@@ -85,7 +85,7 @@ _If you..._
   * `Option 8.` are on **Arch Linux** and want to use **Community packages** see [Arch Community Repositories](#option-8-arch-community-repository)
   * `Option 9.` want to patch your own font see the [Font Patcher](#option-9-patch-your-own-font)
 
-## Features
+## Features <a name="features"></a>
 * A [FontForge Python script](#font-patcher) to patch any font
   * Includes an option to create **Monospaced (fixed-pitch, fixed-width)** _or_ **double-width (non-monospaced)** glyphs
   * For more details see the [**Font Patcher**](#font-patcher) section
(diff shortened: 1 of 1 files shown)


hasecilu:
Speaking of links, I think the readme could be enhanced if we use a link below each header to go back to the TOC/top because the file is quite large.
An example of this is the [README file from awesome-selfhosted](https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/README.md) where you can see a `back to top` item below every header which essentially is just:
```markdown
**[`^        back to top        ^`](#awesome-selfhosted)**
```

hasecilu:
Just another thing, these changes will make the links work on most offline markdown file readers/editors.

Finii:
Thanks for the PR.

Yes, this Github md files are always a bit ... special ;-)

Is this ready or - regarding your 'Edit2' - shall we wait a bit if you find a solution?

Further notes:
* I believe the readme has to be updated/modernized completely - I just have no time
* There are _a lot_ translated readmes, but some are very very old. The problem is that any change on the original readme would need to be translated into all the languages, which I find quite impossible. Especially if we now (in the future) rewrite/modernize the readme. I have no solution for that problem, and sometimes I believe it would be better to not have translated readmes and have the people autotranslate the English readme if they need. Because the question is... If you have an outdated readme_es (for example), will you check the english one if there are un-translated changes? And if you do that, why not read the English one to begin with. It is probably nice to have translations, but only if they are current, and I see no way to _really_ keep them current :-/
Do you have any ideas/thoughts on that?

Finii:
> Speaking of links, I think the readme could be enhanced if we use a link below each header to go back to the TOC/top because the file is quite large. An example of this is the [README file from awesome-selfhosted](https://github.com/awesome-selfhosted/awesome-selfhosted/blob/master/README.md) where you can see a `back to top` item below every header

Yes that makes sense.
But the whole website stuff needs some thoughts maybe. What I find confusing is that `nerdfonts.com` is completely separate and mostly **incomplete**. All information should be there and not in the readme I guess. Now if people head over to the official website they only get part of the truth, and there is not even a link to the github readme or the wiki, or anything. It is completely disjunct. But the font representation with previews etc is only there, so people coming to the readme fist will miss that.
I have just the feeling that here something is amiss. This is not a good and thought through setup.

hasecilu:
> Is this ready or - regarding your 'Edit2' - shall we wait a bit if you find a solution?

I already tried other things even translating some tags and still don't work, I think I'll check other day if its .
For example the `font patcher` link that is in the banner `https://github.com/hasecilu/nerd-fonts/blob/Spanish_readme_file/readme_es.md#font-patcher` works fine with the tag but the links on the TOC are the ones that don't work...

Finii:
I think I got it working on some of the translated readmes, I will check.

hasecilu:
> * Do you have any ideas/thoughts on that?

Yeah, having more resources/docs at the end means more maintenance, a problem could be how to know if someone really uses it? The auto translation is always a good option.   
The case where translations make most sense is in the UI of program, in the context of NF that not applies. And being realistic most of projects (FLOSS, community based)  tend to only have docs/readmes in English.

Also, who really reads the readmes? the finals users or just developers/hackers, kind of related to :

> But the whole website stuff needs some thoughts maybe. What I find confusing is that nerdfonts.com is completely separate and mostly incomplete. 


I looks like the website was made more for the final users that want to use some specific glyph, want to see what font look fine to them, etc.

Here in GH seems to be made thinking in the maintenance of the project itself and help/doc for font-patchers users, maybe that's why it feels like there is not a mirror.

hasecilu:
With this realization: https://github.com/ryanoasis/nerd-fonts/issues/1276#issuecomment-1574220649 is evident that the need of html tags is unnecessary, even the proposal of `back to top` could be useless because having the `Outline` available all the time it's just convenient. I think we just need to decide if is worth it having still the TOC or not.

Finii:
Hmm, shall I say something to this? If it's closed I will probably never revisit it again.

hasecilu:
I think is not necessary here, just in #1276 to decide about TOC.

Issue #1284, opened 2023-06-05
zip vs tar.xz with one vs two readme files


Kissaki:
Looking at [v3.0.2](https://github.com/ryanoasis/nerd-fonts/releases/tag/v3.0.2) `CascadiaCode.zip` vs `CascadiaCode.tar.xz`

* the zip has one `readme.md` file
* the tar.xz has a `readme.md` _and_ a `README.md` file

* `README.md`: `# Cascadia Code` information / docs
* `readme.md`: Short `#Nerd Fonts` note

_I assume this is likely an issue with other font packages too, but I only saw and checked on CascadiaCode._

Kissaki:
I would expect them to at least be consistent.

Is there a reason why the packages *should* be different? Other than historic came-to-be?

---

As a release package I guess I would not include too much technical / technical source information.

Given the package is called Cascadia Code despite the patched font being called Caskaya having a generic readme - even short - seems good to at least indicate what the package is [from].

What I like about the more detailed readme is the generic font information. What is the font, where is it from, what is it designed to do / is it designed for. Different font naming reasoning is critical information.

However, `This source font has been preprocessed` is factually wrong.

The `Which font?` section seems very helpful and important.

---

As a simple consistency fix, I would add both files to the zip too.

As a slightly better low-effort fix it could be reasonable to append the extended (source) readme to the generic readme text.  
That'd solve extract file naming conflicts on Windows. And ensure no confusion between having two readme files.

I'm not too familiar with the structure or processes of this project to assess how readme infos could be split better or differently in sources and [produced in] package. But 1. generic Nerd Font info, 2. generic font info, 3. specific font distrib infos (font rename, nerd font naming) would be great.

Finii:
> However, `This source font has been preprocessed` is factually wrong.

What is wrong with that?

Finii:
I think what is missing from both is the release version?
Sure you can see that in the font itself, but maybe better to have it in the/one readme.

When we decouple font updates from patcher releases maybe the 'in parens' version needs to be included also.

Finii:
> That'd solve extract file naming conflicts on Windows.

Baaah, now I get it :woman_facepalming:

Finii:
The Cascadia readme (`README.md` in both archives, no `readme.md` anymore) looks like the following.
Any more ideas?

(The " is factually wrong" from above, I still do not get it, please explain.)

```markdown
# Nerd Fonts

[...]

PR #1285, opened 2023-06-05
archive-fonts: Unify README handling

Finii:
**[why]**
The readmes in the zip and tar.xz archives differ. The zips have only a very small and generic `readme.md`. The tar.xz have that readme as well as the `README.md` from the patched-fonts/ directory. This should be the same for both.

To have two files with names that just differ in case (`readme.md` and `README.md`) can be problematic on some platforms.

**[how]**
Combine both readmes into one file - put the generic info in the top of the readme.

Also include the RELEASE_VERSION if known into the readme. That makes it more easy to identify which Nerd Font release that archive came from.

RELEASE_VERSION is set in the release workflow.

Fixes: #1284

Reported-by: Jan Klass <kissaki@posteo.de>

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/archive-fonts.sh
+++ b/bin/scripts/archive-fonts.sh
@@ -3,6 +3,7 @@
 # Script Version: 1.2.0
 # Iterates over all patched fonts directories
 # to generate release archives of the patched font(s)
+# Set RELEASE_VERSION to attribute a specific release
 #
 # Example run with pattern matching:
 # ./archive-fonts heavydata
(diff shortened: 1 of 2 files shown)


bin/scripts/archive-fonts.sh
```diff
@@ -56,6 +58,20 @@ while read -r filename; do
     while IFS= read -d $'\0' -r descriptor; do
         path=${descriptor//|*/}
         file=${descriptor//*|/}
+        if [ "$file" = "README.md" ]; then
```
Kissaki, line 61: Does this mean it packages a README.md only if the source dir has a README.md?

Kissaki (approved):

bin/scripts/archive-fonts.sh
Finii: `src/unpatched-fonts/Agave/README.md` + some info = `patched-fonts/Agave/README.md`
`patched-fonts/Agave/README.md` + some other info = `README.md` in archives

If no `src/unpatched-fonts/Agave/README.md` exists a dummy `patched-fonts/Agave/README.md` is created:

https://github.com/ryanoasis/nerd-fonts/blob/df626d3440645a1d49d82f18709e6e4814c638bc/bin/scripts/standardize-and-complete-readmes.sh#L154-L168

Finii:
Not sure why this has not yet been pulled... :->

PR #1286, opened 2023-06-05
Update search page style

fadihania:
- Fix having multiple vertical scrollbars
 - Update icons lazy loading accordingly

#### Description
Fix having multiple vertical scrollbars while searching the icons, it's easier to use and better for User Experience. It's also update the lazy loading to work on the document scrolling instead of icons container

#### Requirements / Checklist

### DIFF
--- a/_includes/css/nerd-font-tweaks.scss
+++ b/_includes/css/nerd-font-tweaks.scss
@@ -110,9 +110,9 @@ a.nerd-font-button:before {
 }
 
 #main .nerd-font-cheat-sheet {
-	max-height: 520px;
+	// max-height: 520px;
 	overflow-x: hidden;
-	overflow-y: scroll;
+	// overflow-y: scroll;
 	margin-top: 40px;
 	max-width: none;
 	justify-content: center;
--- a/cheat-sheet.js
+++ b/cheat-sheet.js
@@ -23,9 +23,12 @@ document.addEventListener('DOMContentLoaded', function () {
         }
     ));
 
-    elementGlyphCheatSheet.onscroll = function () {
-        const threshold = 10;
-        if ((elementGlyphCheatSheet.offsetHeight + elementGlyphCheatSheet.scrollTop + threshold) >= elementGlyphCheatSheet.scrollHeight) {
+    document.onscroll = function () {
+        const elementDoc = document.documentElement;
+        const elementIconsTop = elementGlyphCheatSheet.getBoundingClientRect().top + elementDoc.scrollTop;
+
+        if (elementDoc.clientHeight + elementDoc.scrollTop >=
+            elementGlyphCheatSheet.clientHeight + elementIconsTop) {
             console.log("load more search results");
             loadMoreSearchResults();
         }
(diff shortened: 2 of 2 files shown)


Finii:
This looks excellent! Thank you very much!

Finii:
WHAT!!?

```diff
CommitDate: Mon Jun 5 22:42:31 2023 +0300

    Update search page style
     - Fix having multiple vertical scrollbars
     - Update icons lazy loading accordingly
---
 _includes/css/nerd-font-tweaks.scss | 4 ++--
 cheat-sheet.js                      | 9 ++++++---
 2 files changed, 8 insertions(+), 5 deletions(-)
```

This looks so trivial, I have expected some very complicated stuff :-D

:+1:

Finii:
What do you think about the commit I added, that...

```
    cheat-sheet: Show no icons by default
    
    [why]
    The additional hints in the bottom are hard to notice if we flood users
    with all available items first.
    Usually users want to search the icons and not browse, so showing them
    all per default seems better.
    
    [how]
    Just do not change the search term from "" to something else, so the
    search comes up empty on an empty search.
    
    If people search for a blank the previous mechanics to show all icons is
    triggered instead. A message to hint for this is added to the empty
    search result.
```

...starts the page up like this

image

rszyma:
While this is still open, for further UX improvement I'd suggest adding autofocus on the search box (just like google.com) does

rszyma:
Another suggestion: the text "Enter (part) of a word to search" is not grammatically correct usage of parantheses. The correct version would be this: "Enter (a part of) a word to search"

Finii:
I have bad experience with this autofocus stuff, but that might be a browser problem, where reloading a page steals my focus from other applications or something. But, well, why not. If you could supply an idea what needs to change I can do that together with the grammar correction (tnx, that is the result of doing multiple things in parallel I guess :grimacing: ).

fadihania:
> What do you think about the commit I added, that...
> 
> ```
>     cheat-sheet: Show no icons by default
>     
>     [why]
>     The additional hints in the bottom are hard to notice if we flood users
>     with all available items first.
>     Usually users want to search the icons and not browse, so showing them
>     all per default seems better.
>     
>     [how]
>     Just do not change the search term from "" to something else, so the
>     search comes up empty on an empty search.
>     
>     If people search for a blank the previous mechanics to show all icons is
>     triggered instead. A message to hint for this is added to the empty
>     search result.
> ```
> 
> ...starts the page up like this
> 
> image

That works for showing the comments below, but I think if the website redesigned so the upper section is smaller then we can have those comments at the top, with smaller font and maybe a catchy color or effect

Finii:
So I just merge this now, and more changes can come later.
I know that the commit I added is just a workaround, but while we have no other solution this is the cheapest (easiest) change to still get that information across.
At least for the information on including into websites, that information is maybe out of place anyhow, but we could not come up with a better solution at that time (see 46168d53d00bd43069e292e0df048175b9debb18).

If @rszyma just comments with a diff or idea regarding the focus I can just add that without PR.

PR #1287, opened 2023-06-05
Fix logging error font-patcher

lolicon0930:
#### Description
Python complains about error on line 1881 in font-patcher.
Maybe you guys should take the time to check if the function `global logger` is doing exactly the right thing.

#### Requirements / Checklist

Fix typo in font-patcher.

I'm not sure, maybe this feature isn't finished yet?

Python complains about error when I manually patch fonts.
The shell script I use is:
```
for p in ../iosevka-custom-mono/ttf/*.ttf; do
	echo -e "Patching for font \e[0;36m${p:27}\e[0m ..."
	fontforge -script font-patcher "$p" -cq --careful -out iosevka-custom-mono-nerd --makegroups 0
done
```
When I do not set the parameter `--makegroups 0`, it reports an error.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "4.4.0"
+script_version = "4.4.1"
 
 version = "3.0.2"
 projectName = "Nerd Fonts"
@@ -1878,7 +1878,7 @@ def setup_arguments():
     args = parser.parse_args()
 
     if args.makegroups > 0 and not FontnameParserOK:
-        logging.critical("FontnameParser module missing (bin/scripts/name_parser/Fontname*), specify --makegroups 0")
+        logger.critical("FontnameParser module missing (bin/scripts/name_parser/Fontname*), specify --makegroups 0")
         sys.exit(1)
 
     # if you add a new font, set it to True here inside the if condition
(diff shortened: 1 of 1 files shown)


Finii:
Thanks for finding and fixing this!

There are two more affected calls in the code. To prevent that I added an additional commit that simplifies the whole logger business and allows the programmer to use `logger` throughout and not decide on a case-to-case basis if `logger` or `logging` is the correct thing to use.

> Maybe you guys should take the time to check if the function

Yes. That is often skipped due to laziness.
On the other hand that logging message you stumbled over is just issued if some user did not follow the instructions and is kind of a saveguard for 'shall never happen'. Misuse cases are tested less rigorously, I have to admit. 

P.S. I guess you should use `--makegroups 4`, but of course I do not know your font files. At least the 'original' Iosevka needes that.

Finii:
Add skip-ci, force push

lolicon0930:
In fact in older versions I used to just use the command:
`fontforge -script font-patcher "$p" -q -c --careful -out iosevka-custom-mono-nerd`

But when I did the long-lost update yesterday, I found that there would be a warning, and it is still the same now.
2023-06-06-23:42:48-w
So why doesn't the default `--makegroups 1` work?

Thanks for the suggestion, but `--makegroups 4` doesn't work either.
2023-06-07-00:28:03-w

lolicon0930:
I use Debian sid, here is the script file I use if it helps.
[iosevka_build.sh.txt](https://github.com/ryanoasis/nerd-fonts/files/11667136/iosevka_build.sh.txt)
[func.sh.txt](https://github.com/ryanoasis/nerd-fonts/files/11667139/func.sh.txt)

Finii:
How did you obtain the `font-patcher`?

https://github.com/ryanoasis/nerd-fonts#font-patcher

image

lolicon0930:
Do I need `python3-argparse-addons`?
I think I have all other dependencies.
2023-06-07-01:56:28-r

Finii:
I believe you dont. Just try to run the patcher, it should tell you missing dependencies.

lolicon0930:
After installing `python3-argparse-addons`, the same error is reported without the parameter `--makegroups`, and `--makegroups 4` also does not work.
Ok, I'll try running the patcher directly.

Finii:
Same error, meaning FontnameParser missing?

How did you obtain the font-patcher?

lolicon0930:
I figured out what was wrong, I didn't need `python3-argparse-addons`, my `git sparse-checkout` command in shell script was wrong.
Instead of `git sparse-checkout set "src"` I should use `git sparse-checkout set bin src`.
There should be no problem now, I will look at the difference between `--makegroups 1` and `--makegroups 4`, thank you.

Finii:
As written in the readme, reccommended is using the FontPatcher.zip or the docker image, and not a (sparse) checked out repo.

But now it does not matter anymore ;) 👍

Finii:
> difference between `--makegroups 1` and `--makegroups 4`

It's not very well documented.

`1` has all name parts in full (no abbreviations). Often the resulting name are longer than some old systems can work with. This is the best for modern systems but can have problems on some.

`2` abbreviates the the style and weight parts a bit, like "ExtraBold" becomes "ExtBld".

`3` abbreviates them even more aggressive, so "ExtraBold" becomes "XBd" or something.

The options `4` to `6` are similar, but tackle additionally the problem when not the styles and weights are the problem but the plain family name. Instead of adding "Nerd Font Mono" or similar to the family name it adds the abbreviated form "NFM". This is needed where the original name is already very long.

`5` and `6` behave like `2` and `3`, but with `NFM` etc.

Option `0` produces names with a much simpler scheme, which breaks often.

Option `-1` tells the patcher to leave the names alone and not touch them. This can be used for example when patching over an already Nerd Fonts patched font (yes, people do that) or similar.

lolicon0930:
Thanks for the detailed explanation, I ended up going with option `-1` yesterday because the fonts I patched are only used on my own system.

Issue #1314, opened 2023-07-06
Error patching Input Mono Condensed font


linrongbin16:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

<img width="988" alt="image" src="https://github.com/ryanoasis/nerd-fonts/assets/6496887/95625e9e-b93f-42e5-baef-8a0f49d9565e">

There's an error:
```
ERROR: ====-< Family (ID 1)      too long (36 > 31): InputMonoCondensedIta Nerd Font Thin
   InputMonoCondensedIta Nerd Font Thin
   \===> 'InputMonoCondensedItaNerdFont-Thin.ttf'
```

Expected behavior:

patch font successful.

Example symbols:

#### 🔧 Your Setup

OS: macOS 13.4.1, apple M1 MAX chip
Input Font: https://input.djr.com/download/
brew install fontforge
latest nerdfont font-patch script.

Finii:
```
$ fontforge font-patcher ~/Downloads/input_font/Input_Fonts/InputMono/InputMonoCondensed/InputMonoCondensed-ThinItalic.ttf --debug

Nerd Fonts Patcher v3.0.2-16 (4.4.2) (ff 20230101)
DEBUG: Naming mode 1
DEBUG: Monospace check: Panose says "monospaced"; glyph-width-mono True
DEBUG: Extended glyphs wider bounding box than basic glyphs
DEBUG: Font has negative right side bearing in extended glyphs
DEBUG: Final font cell dimensions 624 w x 1320 h
ERROR: ====-< Family (ID 1)      too long (36 > 31): InputMonoCondensedIta Nerd Font Thin
DEBUG: =====> SubFamily (ID 2)   ok       ( 7 <=31): Regular
DEBUG: =====> Fullname (ID 4)    ok       (36 <=63): InputMonoCondensedIta Nerd Font Thin
DEBUG: =====> PSN (ID 6)         ok       (28 <=63): InputMonoCondensedItaNF-Thin
[...]

PR #1319, opened 2023-07-17
Feature: Select name source

Finii:
#### Description

font-patcher: Add option to select postscript name as name source

**[why]**  
In the post we used these information sources to determine the name and
weights/styles of the to-be-patched font:
* The filename
* The fullname (ID 4)
* The postscriptname (ID 6)

Usually it is best to use the Fullname of a font to determine its real
name and styles/weights:

The Postscript name has the advantage to have a hyphen between name and
styles/weight; but as it can not contain blanks the correct name can not
be determined by this. To get the styles/weights back we use a list of
all possible (?!) weights and styles and sort all the string parts.

This works reasonably well and the fullname is usually best.

Not so with Input Mono Condensed. Here are its names:
ID4: `"InputMonoCondensed LightIta"`
ID6: `"InputMonoConsensed-LightItalic"`

**[how]**
Add option to select between fullname and postscriptname as font naming
source.

For special purposes, also allow a custom (arbitrary) name input.
If that is given the specified name will be used to name the patched
font without adding any suffix - the user has full control on the name.

Fixes: #1314

#### Requirements / Checklist

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -1825,12 +1825,12 @@ def setup_arguments():
     parser.add_argument('-c', '--complete',                          dest='complete',         default=False, action='store_true', help='Add all available Glyphs')
     parser.add_argument('--careful',                                 dest='careful',          default=False, action='store_true', help='Do not overwrite existing glyphs if detected')
     parser.add_argument('--removeligs', '--removeligatures',         dest='removeligatures',  default=False, action='store_true', help='Removes ligatures specificed in JSON configuration file')
-    parser.add_argument('--postprocess',                             dest='postprocess',      default=False, type=str, nargs='?', help='Specify a Script for Post Processing')
-    parser.add_argument('--configfile',                              dest='configfile',       default=False, type=str, nargs='?', help='Specify a file path for JSON configuration file (see sample: src/config.sample.json)')
-    parser.add_argument('--custom',                                  dest='custom',           default=False, type=str, nargs='?', help='Specify a custom symbol font, all glyphs will be copied; absolute path suggested')
-    parser.add_argument('-ext', '--extension',                       dest='extension',        default="",    type=str, nargs='?', help='Change font file type to create (e.g., ttf, otf)')
-    parser.add_argument('-out', '--outputdir',                       dest='outputdir',        default=".",   type=str, nargs='?', help='The directory to output the patched font file to')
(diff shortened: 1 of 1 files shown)


HUMORCE:
Screenshot_230719035208

When specifying name with arg `--name 'Example NF'`, space was ignored. How to specify a font name contains spaces?

Finii:
Oh. Well, I guess you can't because blanks are always removed from the name (e.g. `NetBrains Mono` -> `NetBrainsMono`).

Can you describe your use-case?

Finii:
```console
$ fontforge font-patcher --dry --debug 2 --name 'Example Mono' src/unpatched-fonts/Agave/Agave-Regular.ttf
[...]
DEBUG: =====> Family (ID 1)      ok       (11 <=31): ExampleMono
DEBUG: =====> SubFamily (ID 2)   ok       ( 7 <=31): Regular
DEBUG: =====> Fullname (ID 4)    ok       (11 <=63): ExampleMono
DEBUG: =====> PSN (ID 6)         ok       (11 <=63): ExampleMono
DEBUG: =====> Filename 'ExampleMono-Regular.ttf'
```

If you can edit the `font-patcher` script you can enable the blanks by changing the circled value from `True` to `False`:

image

Finii:
Maybe that should be the default if `--name` is given? :thinking: 

Use case would be helpful to decide on this. I in fact did not think anybody would ever use that feature :-D

HUMORCE:
> Can you describe your use-case?

I just want to get a font name that is as short as possiable.

e.g.: Input Mono #1314

```
fontforge -script font-patcher --mono Input/Input-Regular.ttf --name 'Input NF Bold'
# actual: InputNF
# expect: Input NF
```

> enable the blanks by changing the circled value from True to False

Yeah, It works like I expected. Thanks!

> Maybe that should be the default if --name is given? 🤔
> 
> Use case would be helpful to decide on this. I in fact did not think anybody would ever use that feature :-D

It should be default, IMO. Becuase user explicitly specified a name contains space, but space disappeared after patching.

Finii:
Makes sense, thank you. Will change accordingly.

Finii:
Implemented via 12b51f51e257 in `font patcher 4.5.1`

PR #1362, opened 2023-09-29
Describe patcher family-name generation in readme

frantisekhanzlikbl:
#### Description

updated the project's readme to mention that the font patcher CamelCases and removing whitespace from the resulting font's family names.

#### Requirements / Checklist

the change contains GitHub's alert syntax, which might not render in the preview, but should render in the actual readme. (ref: https://github.com/orgs/community/discussions/16925#discussioncomment-6879962)

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -362,8 +362,8 @@ Patching the font of your own choosing:
   docker run --rm -v /path/to/fonts:/in:Z -v /path/for/output:/out:Z nerdfonts/patcher [OPTIONS]
   ```
 
-[!NOTE]
-The resulting font's family will be set to the original family after CamelCasing, removing whitespace and appending ` Nerd Font`. For example, `iosevka term` would become `IosevkaTerm Nerd Font`.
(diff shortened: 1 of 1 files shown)


Finii:
Thanks for the PR.

I'm not sure with the term "Family"... altough technically correct, maybe we should say just 'name' instead?
For the not so expertly users?
Or at least "Family name"?

Finii:
@allcontributors pleas add @frantisekhanzlikbl for doc

frantisekhanzlikbl:
> Thanks for the PR.
> 
> I'm not sure with the term "Family"... altough technically correct, maybe we should say just 'name' instead? For the not so expertly users? Or at least "Family name"?

That makes sense. I'd lean towards mentioning both, since at least for web devs, the term family is more widely used, I believe. What about the current patch, does that look OK?

Finii:
> What about the current patch, does that look OK?

Perfect. Thanks again.

PR #1365, opened 2023-10-01
Update nerd-fonts-generated.css

asteriskayush007:
i update yours font weights, line heights, styles,etc

#### Description

#### Requirements / Checklist

### DIFF
--- a/css/nerd-fonts-generated.css
+++ b/css/nerd-fonts-generated.css
@@ -9,18 +9,18 @@
   font-family: 'NerdFontsSymbols Nerd Font';
   /* @TODO you need to adjust the path to the font: */
   src: url("../fonts/Symbols-2048-em Nerd Font Complete.woff2") format("woff2");
-  font-weight: normal;
-  font-style: normal;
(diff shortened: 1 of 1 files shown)


Finii:
Thanks for the PR.

Maybe you can explain the reasons for the change?
I'm intrigued to know why the weight is first 100 and then 300?


The file you change is a generated files. Any changes you introduce will be undone the next time that file is generated - see comment on line 5:

Screenshot 2023-10-01 at 11 39 42

So instead the generator must be changed if you want to change anything in this file.

And then there is the `nerd-fonts-generated.min.css` which is in principle the same file, so it should be changed in tandem. Which is easy if the gerenator is changed.

The generator is https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/generate-css.sh
And the place to edit to introduce your change would be the 3 files starting with `css-` (which are inputs to the script):

Screenshot 2023-10-01 at 11 57 52

Finii:
Closing for now, we can reopen if you explain.

PR #1385, opened 2023-10-16
doc: Correct name for AnonymicePro Nerd Font

kingarrrt:
#### Description

The AnonymicePro Nerd Font is incorrectly named Anonymice Nerd Font.

#### Requirements / Checklist

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -114,7 +114,7 @@ See [Wiki: Icon names in shell][wiki-icon-names-in-shell]
 |:--------------------------------------------------|:----------------------------------|:------|:-----|:------------------|
 | [3270 Nerd Font][p-3270]                          | [3270][f-3270]                    | 3.0.1 | NO   | ![w] ![m2] ![l]   |
 | [Agave][p-agave]                                  | [Agave][f-agave]                  | 37    | NO   | ![w] ![m2] ![l]   |
-| [Anonymice Nerd Font][p-anonymous-pro]            | [Anonymous Pro][f-a-pro]          | 1.002 | YES  | ![w] ![m2] ![l]   |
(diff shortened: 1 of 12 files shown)


Finii:
Good catch!

Thanks for the PR!

Hmm, I see more strangenesses...:

image

Some have `Nerd Font` in the listed name, and some dont. :thinking: 

Also checked `fonts.json`, but there AnonymicePro is correct.

Finii:
@allcontributors please add @kingarrrt for doc

Finii:
Maybe you like to comment on this...:

image

I would drop the Win-Linux-Mac indicators, the table would get vertically condensed - and the indicators are defunct anyhow.

Then we should unify the naming, add - or remove - the `Nerd Font` part from the patched font names (column 1)?

The column names (column 1 and 2) are also rather ... similar `Font Name` and `Font Name and Repo`...
Maybe change first column header to `Patched Font Name`?

kingarrrt:
Yes re defunct Win-Linux-Mac indicators.

It looks like the Patched Fonts table could be generated from `fonts.json`  and the fonts' `README.md` files to ensure correctness.

Finii:
> could be generated from `fonts.json`

**Excellent** idea! :+1:

(Except at least ProFont which has more than one entry in the table because of different versions. Hmm, I seem to remember that there are other fonts also with a variety of versions. But that is just a detail.)

I would even put the version information into `fonts.json`. We could reuse that information for nerdfonts.com then.

Are you interested in implementing something? Otherwise I would code something up.

kingarrrt:
Very sorry, I don't have bandwidth to implement.

Issue #1383, opened 2023-10-13
Icon for Neovim in custom?


Finii:
### Discussed in https://github.com/ryanoasis/nerd-fonts/discussions/1380

<div type='discussions-op-text'>

<sup>Originally posted by **georgeguimaraes** October 13, 2023</sup>
Hi,

Would you consider adding Neovim icon as custom directly into nerd-fonts?

Reasons:

1. There's no Neovim icon in upstream repos. However, there are some PRs lingering in some upstream repos that appear discontinued:

    - https://github.com/lukas-w/font-logos/pull/93 (font-logos appear to be unmaintained)
    - https://github.com/vorillaz/devicons/issues/151 (vorillaz/devicons appears to be unmainted, last commit 6y ago)

2. Even the "popular, maintained devicons/devicons" (as described in https://github.com/ryanoasis/nerd-fonts/issues/615) does not have a Neovim icon yet. And their last release was 17 months ago. So even if a PR was accepted upstream, there's gonna be a potentially huge delay until nerd-fonts can use that.

3. There's already `custom-vim` in Nerd Fonts (which I'm using as replacement for a neovim logo).</div>

georgeguimaraes:
Tks for creating the issue!

PR #1391, opened 2023-10-20
Add Neovim icon (solves #1383)

georgeguimaraes:
#### Description

I decided to take shot into implementing #1383. Not sure if this is it, or if I should run `bin/scripts/gotta-patch-em-all-font-patcher!.sh` and commit all the changes. Let me know if this is ok.

#### Requirements / Checklist

Adds the neovim icon as discussed in #1383 

The generated `original-source.otf` now has the Neovim icon:

CleanShot 2023-10-20 at 15 04 30

#1383


Finii:
Thanks for the PR!

To add a icon one just needs to throw the `svg` into the correct directory and add a line to `icons.tsv`. The workflow than automagically updates the `i_seti.sh` and the `original-source.otf` ([workflow PackSVGs](https://github.com/ryanoasis/nerd-fonts/blob/master/.github/workflows/packsvgs.yml)). Maybe we should add that to the comments in `icons.tsv`?

Manually one needs to
* expand the patching range in `font-patcher`
* correct the [wiki page](https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points)

In the original issue you mentioned font-logos. There is a PR pending to add the neovim icon there. You correctly pointed out that the repo seems dead.
But in fact, after #1391 has been raised I investigated and it seems we can revive the repo (see [Issue 95 over there](https://github.com/lukas-w/font-logos/issues/95)).
That would of course be preferred.

So maybe we should wait a bit and see if we can get that going again. The messages seemed rather promising.

Finii:
I believe the font-logos neovim icon has a solid lift-side-bar thing, while here it is hollow (to reflect that it is blue instead of green, a guess).
This is just an observation.

Screenshot 2023-10-21 at 12 44 55

_Taken from [font-logos fork release](https://github.com/hasecilu/font-logos/releases/tag/v1.1.0)_

Finii:
The original black-and-white logo published by neovim themselves is like THIS logo, i.e. with a hollow left leg:

https://github.com/neovim/neovim.github.io/blob/master/logos/neovim-logo-1color.svg

georgeguimaraes:
> In the original issue you mentioned font-logos. There is a PR pending to add the neovim icon there. You correctly pointed out that the repo seems dead.
But in fact, after https://github.com/ryanoasis/nerd-fonts/pull/1391 has been raised I investigated and it seems we can revive the repo (see https://github.com/lukas-w/font-logos/issues/95).
That would of course be preferred.

That's interesting! Let's wait then if font-logos comes back to activity. It seems that the maintainer has not given access to other people.

However, I think there may be an underlying question here: Should Nerd Fonts rely on upstream repos that (as you mentioned in that issue) appear to have no standalone use? As oppose to, I'd say, Material Design. Of course, that's just me thinking out loud, it's a questions for nerd fonts' maintainers to discuss. :)

Just hoping to get Neovim icon soon :)

georgeguimaraes:
> The original black-and-white logo published by neovim themselves is like THIS logo, i.e. with a hollow left leg:
> 
> https://github.com/neovim/neovim.github.io/blob/master/logos/neovim-logo-1color.svg

Yes, that's the one I used (hollow left leg).

georgeguimaraes:
> To add a icon one just needs to throw the `svg` into the correct directory and add a line to `icons.tsv`. The workflow than automagically updates the `i_seti.sh` and the `original-source.otf` ([workflow PackSVGs](https://github.com/ryanoasis/nerd-fonts/blob/master/.github/workflows/packsvgs.yml)). Maybe we should add that to the comments in `icons.tsv`?

I guess this paragraph in the contributing file, in a entry regarding new icons, should be enough.

georgeguimaraes:
btw @Finii, tks for the interactions here! I really appreciate the care you have with nerd fonts and all the work

georgeguimaraes:
> Manually one needs to
> 
> * expand the patching range in `font-patcher`

You mean this line https://github.com/ryanoasis/nerd-fonts/blob/18ff974fbab4eaf56d0c02d1a3e49492a44197bd/font-patcher#L1054?

It seems that U+E5AE is already included in that range. Is that so?

Finii:
> It seems that U+E5AE is already included in that range. Is that so?

Oh, I anticipated that already in https://github.com/ryanoasis/nerd-fonts/commit/869d6f93513ddca7a0bddfd999002daaf2d00589, so there is nothing to do in font-patcher :-)

Screenshot 2023-10-22 at 23 51 08

Finii:
Thanks again for the PR.

I hope the changes of the documentation is readable and helps a bit.

Here you can see the workflow in action that crates the new original font:
https://github.com/ryanoasis/nerd-fonts/actions/runs/6612722696/job/17959104734

And here the automatic commits added after your PR:

image

So the changes are now in the zip and the docker patcher.
All the prepatched fonts will only follow on the next release - to have it already now you need to self patch (or tell me which font(s) you need, I can prepare them)

georgeguimaraes:
Nice changes to `contributing.md`. Much clearer to understand that all the heavy lifting of patching the fonts and committing them to the repo is done by the workflows.

I'll try to patch Victor Mono by myself. If I don't succeed I'll ask for help. Thanks so much!

Finii:
Nearly forgot...

I always struggle with the type; code seems not fitting, but all other are also not really good. I believe we used code before for this kind of contribution.

berlinx2104:
Screenshot from 2023-10-27 20-57-47
Can you guys reconsider that the left part should be overpainted so it's actually easier to see at the terminal? That edge is really small to see at the terminal. Anyway, thank you so much for taking the time to make it and I respect all your decisions.

Finii:
@berlinx2104 Thanks for the feedback!

I will try to improve the visibility.

The completely filled symbol is already in Font Logos (see links above) which I expect will pull the PR that contains it soon and afterwards we update that here, so both variants will end up in the patched fonts.
(But unfortunately Lukas did not make one of us Maintainer, yet)

georgeguimaraes:
CleanShot 2023-10-27 at 13 51 07

I kinda got used to see it as a modified "V", but yes, the left leg is too subtle. Maybe it was a design decision by the Neovim team?

Finii:
Do you think this is better, or not enough?

image

Finii:
Corrected the aspect ratio:

image

Finii:
> Maybe it was a design decision by the Neovim team?

Well, it was designed to be displayed at a much bigger size. That is why icons come in different sizes, because rescaling usually is not good enough. Here I increased the gap and increased the 'stroke width' of the left leg. In that process I changed the angle of the middle-bar. If I do not change that the ascpect ratio changes, see image above.

georgeguimaraes:
That looks way better (the 2nd version)

Finii:
Ok, I will push that, so that you (both) can try it out in real...

Finii:
Changes pushed via 27f6ea142d40

image

ZIP and docker ready.

georgeguimaraes:
CleanShot 2023-10-27 at 15 16 10

It is certainly more visible. I can clearly see the hollow leg on my terminal now. Thanks, @Finii

berlinx2104:
Thanks, @Finii
Screenshot from 2023-10-29 00-29-42

Issue #1242, opened 2023-05-15
Caskaydia Code and Visual Studio


Finii:
I have removed Caskaydia Cove, verified no Caskaydia font files in system nor user folder, restarted the system, installed all the Caskaydia ttf for all users.

They work fine in various programs (VS Code, KDiff3, Double Commander).
But not in Visual Studio. *(It falls back to a different font, seemingly Courier.)*

Windows 10. VS 2022 17.5.5. Nerd Fonts 3.0.1.

_Originally posted by @ite-klass in https://github.com/ryanoasis/nerd-fonts/issues/785#issuecomment-1547453922_

ite-klass:
I tested in the text/code editor.

![image](https://github.com/ryanoasis/nerd-fonts/assets/8045680/ae1a98d8-2a48-442f-814b-0ebd7553e442)

As you can see in the settings dialog the font preview looks fine, but as you can see in the background text editor (applied settings) it falls back to Courier.

ite-klass:
Cascadia Code works fine.

![image](https://github.com/ryanoasis/nerd-fonts/assets/8045680/2e27b7b9-9b09-48f2-9bd7-5e6beed33f43)

Finii:
The 'massaged' source font (`CascadiaCode-Regular-vtt.ttf`, i.e. after VTT and `ttx`) seems to work fine in VS2022.

Both `CaskaydiaCove NF` and `CaskaydiaCove NFM` are not working and will be replaced by `Consolas` :thinking:

Finii:
:heavy_check_mark: `CascadiaCode-Regular-vtt.ttf` -> open in `fontforge` and immediately generate
:red_circle: `CascadiaCode-Regular-vtt.ttf` -> `font-patcher --mono`
:red_circle: `CascadiaCode-Regular-vtt.ttf` -> `font-patcher`
:red_circle: `CascadiaCode-Regular-vtt.ttf` -> `font-patcher` without tweaking
:heavy_check_mark: `CascadiaCode-Regular-vtt.ttf` -> `font-patcher` without `FontnameParser.rename_font()`
:red_circle:  `CascadiaCode-Regular-vtt.ttf` -> `font-patcher` with `FontnameParser.rename_font()` without flags
:red_circle:  `CascadiaCode-Regular-vtt.ttf` -> `font-patcher` with `FontnameParser.rename_font()` without PS
[...]

PR #1406, opened 2023-11-11
Bugfix for Visual Studio 2022

Finii:
#### Description

**name-parser: Unify abbreviations in family names**

**[why]**
Because the newer Preferred/Typographic names ID 16 and ID 17 have not
a very demanding length limit we add the long form of the name
addendum (i.e. Nerd Font, Nerd Font Mono, Nerd Font Propo).

In the more restricted old names ID 1 and ID 2 we use the short forms
(i.e. NF, NFM, NFP).

This seems to be problematic with Visual Studio (Community) 2022 and the
fonts can be selected but are not really used.

The Postscript family name is never shortened which seems to be of no
consequence, but still is different than the other.

**[how]**
When creating the Preferred/Typographic Family (ID 16) we check the
shortening mode first and abbreviate the parts as needed and alike ID 1.

This will also change the filenames, because they base on the SFNT
table. We can not change that without changing the whole mechanism.

**[note]**
Also add new tool that lists all names of fonts, including the
Postscript ones.

Fixes: #1242

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/name_parser/FontnameParser.py
+++ b/bin/scripts/name_parser/FontnameParser.py
@@ -180,6 +180,15 @@ class FontnameParser:
         sub = FontnameTools.postscript_char_filter(sub)
         return self._make_ps_name(fam + sub, False)
 
+    def long_family(self):
+        """Get unabbreviated Familyname"""
+        (name, rest) = self._shortened_name()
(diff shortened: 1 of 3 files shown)


Finii:
Forgot about the `keep regular` rule -> force push

PR #1416, opened 2023-11-14
Update nerd-fonts-generated.css

dnoice:
#### Description

The css 'speak' property is not widely supported and is currently not a standard property. If you're using it to prevent screen readers from reading certain elements, a more standard approach is to use the aria-hidden="true" attribute in your HTML.

This will hide the span with the class nf from screen readers. This is a more widely supported and standard way of controlling how screen readers interact with your elements.

it's really not much to get excited about, I just noticed it, and wanted to bring it up to your attention.
the line has been wrapped in a comment.

/* speak: none; */


Finii:
Thanks for the PR!

I'm no web programmer and such details are beyond me, so we must rely on expert like you :-)

dnoice:
oh, no problem what so ever, in fact github CoPilot caught it, I was just
being a good steward of information flow

PR #1419, opened 2023-11-16
Increase Powerline glyph overlap

### LINKED ISSUE #1245: nf-hack behaves differently than the original hack-font
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
On `Arch` (x86_64, Kernel 6.3.2) with `Bash` on `Konsole` (23.04.1) and `oh-my-posh` the hack-font provided by the arch-repos (`extra/ttf-hack 3.003-3`) behaves differently on rendering e.g. powerline-symbols. The original hack-font already contains some powerline-symbols (which do render correctly in `konsole`), but the nf-hack font slightly squishes the symbols together (depending on font-size. however, the usable range is broken anyways, i usually use fonsize 10.0). On the example of the `right|left-hard-devider` icons, it looks like, as if the triangle is misplaced by 1px or so, see screenshots for details. While it doesn't look super-annoying, the original hack-font renders differently. So far i tested some more fonts (but without direct comparison to their original counterpart), namely dejavu-sans, 3270, anonymice-pro, meslo and sourcecode pro, but all of them seem to render slightly off, so i thought this was a konsole-issue, however, og-hack renders fine... (tested basically all other konsole-options which might have an influence but with no result)

Expected behavior:
Both fonts should behave identical

Example symbols:
- nf-pl-left_hard_divider
- nf-pl-right_hard_divider
- ...and many more (at least basically all useable dividers)
#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - `HackNerdFont-Regular.ttf` (but applies to mono-variants as well as other nf-fonts as well)
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   -  https://github.com/ryanoasis/nerd-fonts/releases/latest/download/hack.zip (by included downloader within `oh-my-posh`)
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - konsole 23.04.1
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Arch Linux

#### ★ Screenshots (Optional)
fonsize=10:
original hack-font:  
og-hack  

nf-hack:  
nf-hack  
another example with more colors:  
nf-hack-root

Finii:
#### Description

**[why]**
We still fight with the faint lines on the big side of the powerline
glyphs. They come from the LCD antialiasing mode that has problems with
the borders.

Other fonts use far more overlap. We use only a modest amount of overlap
(2% of the width).

**[how]**
As the other fonts do, increase the overlap (to 7% now). We could not do
that before, because we had no full-height 'landing platforms' on the
outsides and so the triangle would look cut off at not 100% height.

Now with the landing platforms we can increase the overlap and still
have nice looking triangles that visibly reach the top and bottom of the
line.

Implement this only for the filled triangles to see how it goes.

#### Requirements / Checklist

Fixes #1245 (maybe, at least partially)

image
_left: before, right: after_


Finii:
I wrote 7% in the commit message and in the comment above, but in reality it's only 6% :grimacing:

Issue #1454, opened 2023-12-09
Wrong character for U+2588 (Full Block) in D2Coding


hdd1013:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:
U+2588 (Full Block), commonly used in progress bars, was incorrectly patched into a "Biohazard" symbol

Expected behavior:
U+2588 (Full Block) in D2 Coding Ligature Nerd Font should show the correct Block element symbol

Example symbols:
█, U+2588, Block Elements

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - D2CodingLigatureNerdFontPropo-Bold.ttf
   - D2CodingLigatureNerdFontPropo-Regular.ttf
   - D2CodingLigatureNerdFontMono-Regular.ttf
   - D2CodingLigatureNerdFontMono-Bold.ttf
   - D2CodingLigatureNerdFont-Regular.ttf
   - D2CodingLigatureNerdFont-Bold.ttf
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - The affected font files are currently being distributed on the main nerd-fonts repository, oh-my-posh website, and oh-my-posh's font install command.
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - Windows Terminal, Lxterminal, Konsole
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Applies to all major operating systems (Windows, OS X, Linux, Raspberry Pi OS-Also a linux)

#### ★ Screenshots (Optional)

![d2codingu2588](https://github.com/ryanoasis/nerd-fonts/assets/7456108/bbf3dd01-2d51-40b2-b9ff-995f667f676d)
Normal Full Block symbol in the original D2Coding Font

![d2codingnerdu2588](https://github.com/ryanoasis/nerd-fonts/assets/7456108/567bff58-7b55-40d3-9d51-604e5e44315f)
Incorrect symbol in the Nerd Font

[...]

PR #1457, opened 2023-12-10
font-patcher: Break alternative unicode links before patching

Finii:
**[why]**
The D2Coding font has a lot of glyphs shared between multiple unicode codepoints.
Usually we just want to change the glyph on one particular codepoint and not on all linked ones; which will affect codepoints not in our patching ranges.

For example the glyph FULL BLOCK is used by codepoint `2588` and `E286`. In the `Exxx` range they have a lot of icons, and also a full block.

**[how]**
Remove all links on glyphs that we are about to patch over, so that we really just patch the codepoint we want to patch (and have no hidden 'links').

**[note]**
This could in principle break alts or ligs if they rely on another codepoint that we removed. But breaking ligs is always a possibility when patching over existing glyphs.

[note]
I checked several (not all) fonts in our repertoire, and only D2coding is affected.

Fixes: #1454

Reported-by: @hdd1013

#### Requirements / Checklist

Once check one font file of all source-fonts via ... there is a test script for that?


font-patcher
```diff
@@ -1408,6 +1408,20 @@ class font_patcher:
...
                 glyph_scale_data = self.get_glyph_scale(sym_glyph.encoding, scaleRules, stretch, self.sourceFont, currentSourceFontGlyph) if scaleRules is not None else None
             else:
+                # Break apart multiple unicodes linking to one glyph
+                if currentSourceFontGlyph in self.sourceFont:
+                    altuni = self.sourceFont[currentSourceFontGlyph].altuni
+                    if altuni:
+                        codes = { self.sourceFont[currentSourceFontGlyph].unicode }
+                        for v, s, r in altuni:
+                            if v >= 0:
+                                codes.add(v)
```
Finii, line 1418: Maybe less verbose

```python
codes = { v for v, s, r in altuni if v >= 0 }
codes.add(self.sourceFont[currentSourceFontGlyph].unicode)
```

PR #497, opened 2020-06-28
Update Fira Code to latest version

gitfool:
#### Description

Updates Fira Code to latest version (v5.2). Fixes #464.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Read or at least glanced at the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting)
- [x] Read or at least glanced at the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki)
- [x] Scripts execute without error (if necessary):
```bash
fontforge -script font-patcher src/unpatched-fonts/FiraCode/Bold/FiraCode-Bold.ttf
rm 'Fira Code Bold Nerd Font.ttf'

pushd bin/scripts
./standardize-and-complete-readmes.sh FiraCode
rm ../../patched-fonts/FiraCode/extras/font-info.md
./gotta-patch-em-all-font-patcher\!.sh FiraCode

popd
zip -9 FiraCode -rj patched-fonts/FiraCode -i '*.[o,t]tf'
```
- [x] Extended the README and documentation if necessary, e.g. You added a new font please update the table
  - replaced Fira Code `showcases` with `extras` to align with the latest upstream

#### What does this Pull Request (PR) do?

Updates Fira Code to latest version (v5.2).

#### How should this be manually tested?

Install the font and try it in your favorite editor. I'm already using it in VS Code and Windows Terminal.

[FiraCode.zip](https://github.com/ryanoasis/nerd-fonts/files/4843043/FiraCode.zip)

#### What are the relevant tickets (if any)?

#464

PR #1476, opened 2024-01-09
fix: Latest configuration for `devcontainer.json` and Dockerfile

mdsanima:
#### Description

#### Requirements / Checklist

Updating the `.devcontainer` configuration to latest version.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,5 @@
-ARG VARIANT="3"
-FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+ARG VARIANT=3
+FROM mcr.microsoft.com/vscode/devcontainers/python:${VARIANT}
 
 RUN apt-get update \
     && export DEBIAN_FRONTEND=noninteractive \
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,32 +1,36 @@
 // For format details, see https://aka.ms/vscode-remote/devcontainer.json or this file's README at:
-// https://github.com/microsoft/vscode-dev-containers/tree/v0.122.1/containers/python-3
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/python-3
 {
     "name": "Python 3",
     "build": {
         "dockerfile": "Dockerfile",
         "context": "..",
-        "args": { "VARIANT": "3" }
+        "args": {
+            "VARIANT": "3",
+            "NODE_VERSION": "none"
+        }
     },
-
-    "settings": {
-        "terminal.integrated.shell.linux": "/bin/bash",
-        "python.pythonPath": "/usr/local/bin/python",
-        "python.linting.enabled": true,
-        "python.linting.pylintEnabled": true,
-        "python.formatting.autopep8Path": "/usr/local/py-utils/bin/autopep8",
-        "python.formatting.blackPath": "/usr/local/py-utils/bin/black",
-        "python.formatting.yapfPath": "/usr/local/py-utils/bin/yapf",
-        "python.linting.banditPath": "/usr/local/py-utils/bin/bandit",
-        "python.linting.flake8Path": "/usr/local/py-utils/bin/flake8",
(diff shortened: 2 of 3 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,7 +1,6 @@
-ARG VARIANT="3"
-FROM mcr.microsoft.com/vscode/devcontainers/python:0-${VARIANT}
+ARG VARIANT=3
+FROM mcr.microsoft.com/vscode/devcontainers/python:${VARIANT}
```
Finii, line 2: Just for documentation:

See https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3/README.md#configuration

.devcontainer/devcontainer.json
```diff
@@ -1,32 +1,35 @@
...
-// https://github.com/microsoft/vscode-dev-containers/tree/v0.122.1/containers/python-3
+// https://github.com/microsoft/vscode-dev-containers/tree/v0.245.2/containers/python-3
 {
     "name": "Python 3",
     "build": {
         "dockerfile": "Dockerfile",
         "context": "..",
-        "args": { "VARIANT": "3" }
+        "args": {
+            "VARIANT": "3"
```
Finii, line 9: I believe we do not need Node?

```suggestion
            "VARIANT": "3"
            "NODE_VERSION": "none"
```

Finii (commented):
Thanks for the PR!

The devcontainer has been introduced silently with
* #497 

and no explanation given why we possibly want that.

Maybe you can explain what the advantages are or the use-cases, maybe adding documentation. To be honest I was on the brink of removing it some months ago for uselessness and non-documentation.

.devcontainer/Dockerfile
```diff
@@ -1,7 +1,6 @@
...
 
-RUN apt-get update \
-    && export DEBIAN_FRONTEND=noninteractive \
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
```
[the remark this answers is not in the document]
mdsanima: This change is from [official documentation](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3/.devcontainer/Dockerfile#L15).

.devcontainer/Dockerfile
mdsanima: This changes also is from [official documentation](https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3/.devcontainer/Dockerfile#L2C1-L2C1) but we put the latest version of Python.

.devcontainer/devcontainer.json
mdsanima: Now I added this option too.

Finii:
> As for documentation, I can write it, can you tell me where? Should this be in the READMD.md file or maybe in some other file?

That would be great! Just a `./devcontainer/README.md` maybe is enough?

.devcontainer/Dockerfile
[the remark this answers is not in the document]
Finii: I'd rather keep it one command per line as before. In principle it is the same, but how they write it it is misleading - the DEBIAN_FRONTEND envvar is used by the next command (apt-get install) and has nothing to do with the update.

Finii:
@allcontributors please add @mdsanima  for code

.devcontainer/Dockerfile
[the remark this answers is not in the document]
mdsanima: Ok, I rebase this!

Finii:
Personally I believe translations age not gracefully, i.e. get out of sync more or less quickly.

Thanks for adding the explanation/readme!

I will pull this in the next minutes.

Just out of interest... Why would one want to use THIS instead of just the docker container that we always keep up to date [1] directly :thinking: 

I mean, if you have docker installed, you just need to
```
docker run --rm -v /path/to/fonts:/in:Z -v /path/for/output:/out:Z nerdfonts/patcher [OPTIONS]
```
which will download the most up to date font-patcher etc etc.
Maybe this is a leftover from the time where we did not automagically rebuild the container on every push?

[1] https://hub.docker.com/r/nerdfonts/patcher

PR #1480, opened 2024-01-11
fix: Added missing `NODE_VERSION` options in Dockerfile

mdsanima:
Now this is a properly configured file for Dev Container.

Only installing Python on the container, the Node is not installed, as was the case before.

#### Description

#### Requirements / Checklist

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,9 @@
 ARG VARIANT=3
 FROM mcr.microsoft.com/vscode/devcontainers/python:${VARIANT}
 
+ARG NODE_VERSION="none"
+RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
+
 RUN apt-get update \
     && export DEBIAN_FRONTEND=noninteractive \
     && apt-get install --no-install-recommends -y bc fontforge \


.devcontainer/Dockerfile
```diff
@@ -1,6 +1,9 @@
 ARG VARIANT=3
 FROM mcr.microsoft.com/vscode/devcontainers/python:${VARIANT}
 
+ARG NODE_VERSION="none"
+RUN if [ "${NODE_VERSION}" != "none" ]; then su vscode -c "umask 0002 && . /usr/local/share/nvm/nvm.sh && nvm install ${NODE_VERSION} 2>&1"; fi
```
Finii, line 5: You set `NODE_VERSION` explicitly to `none` and then you do some code when it is not `none` (which can never be the case). The `if` is always false.

Finii (commented):
This PR does not change the resulting docker image?
I do not understand.

.devcontainer/Dockerfile
mdsanima: This is form official documentation on [Dockerfile](https://github.com/microsoft/vscode-dev-containers/blob/v0.245.2/containers/python-3/.devcontainer/Dockerfile#L6C1-L7C144).

Finii: Well that is just a template. As the commended out lines further down this is intended as a starting point for users that want to have `node` and can easily set a version there - the code is use-ready.

But here, we have a concrete Dockerfile, and we never ever want `node`, so why have code that does nothing?

mdsanima: You're actually correct, but the overall idea of having this file is to build a development environment for the application. The `Dockerfile` is placed in a dedicated folder `.devcontainer` solely for building the container on Dev Container extension in Visual Studio Code. Then, you can open it inside the Visual Studio Code environment, where you continue working in an isolated setup. But I understand if you don't need this code.

PR #1485, opened 2024-01-18
nerdish

marttie:
#### Description

#### Requirements / Checklist


.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
```
Finii, line 29: Is installation via `pip` not ... deprecated?

I seem to remember that it raises a 'use virtual environments' and other warnings and one should instead use system packages (which is not a problem, just use `apt install -y python3-pytest-flake8`)?

Finii (commented):
Thanks for the PR!

Unfortunately you did not fill any field in the PR template (Description, etc).

Can you explain what it does, esp in comparison to Codeclimate :grimacing: ;-)

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v3
```
Finii, line 23: `v5` is current :grimacing:

.github/workflows/python-app.yml
Finii: Hmm, from the action it seems that they still endorse `pip`?
But on lots of CI at work I know I removed all the pip stuff.

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+      with:
+        python-version: "3.10"
```
Finii, line 25: 3.11?

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+    - name: Lint with flake8
+      run: |
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
Finii, line 36: Why complexity 10?

Is that a different complexity than Codeclimate's?
Here we set 9: https://github.com/ryanoasis/nerd-fonts/blob/master/.codeclimate.yml

.github/workflows/python-app.yml
```diff
@@ -0,0 +1,39 @@
...
+    - uses: actions/checkout@v3
+    - name: Set up Python 3.10
+      uses: actions/setup-python@v3
+      with:
+        python-version: "3.10"
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install flake8 pytest
+        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
```
Finii, line 30: We don't have a requirements file, and this PR does not add one.
But of course there are other dependencies :grimacing: Probably they are not needed for linting, though. :woman_shrugging:

Finii:
Closed due to no response.

Issue #1451, opened 2023-12-05
When do we get our inverse U+E0B0 Triangle?


a-usr:
**Requirements**

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my request and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

**Is your feature request related to a problem? Please describe.**
Some Terminals dont apply effects like transpaency to foreground colors.

**Describe the solution you'd like**
Inverted version of U+E0B0 and other powerline symbols without already existing inverse counterparts

**Describe alternatives you've considered**
U+E0D4: Looks bad

**Additional context**
This has been requested on multiple occasions, for example [Here](https://github.com/ryanoasis/powerline-extra-symbols/issues/7)
Also, its not like the svg's have to be made from scrap, some already exist in the Powerline Extra Repo ([here](https://github.com/ryanoasis/powerline-extra-symbols/blob/master/src/uniE0D7_Powerline_normal-left-inverse-cutout.svg) for example)

Finii:
Note to self:
Check if this is still a problem with Win11 (i.e. Windows Terminal)

Finii:
> U+E0D4: Looks bad

Can you explain why? Looks bad for your use case, or is there an issue with the icon itself?

Finii:
Are you willing to test the new glyphs?
I can create patched fonts for your preferred sourcefont and you can try.
Please specify which font you'd like to test @a-usr.

a-usr:
> > U+E0D4: Looks bad
> 
> Can you explain why? Looks bad for your use case, or is there an issue with the icon itself?

It's not the icon itself, but it just don't match with what I need in my use case (specifically the empty line in the middle and the fact that the angle of the triangle when the symbol has the background colour as foreground and vice versa)

Finii:
> I would love to test CascaydaCove

You can find that in the PR: https://github.com/ryanoasis/nerd-fonts/pull/1490#issuecomment-1902035907

Please report back in the PR thread.

PR #1490, opened 2024-01-19
Add inverse powerline arrow heads

Finii:
image

Fixes: #1451

See also (merged but not released):
https://github.com/ryanoasis/powerline-extra-symbols/commit/eb6b9721a893bf97268dfed507680467ef91c189

#### Requirements / Checklist

Add the inverted arrow heads that PowerlineExtra inteded to be at `E0D6` and `E0D7` but never released.

The glyphs are new/hand drawn.


Finii:
Anyone interested to test the glyphs out?

I can create patched fonts for your preferred sourcefont and you can try.
Please specify which font you'd like to test.

Finii:
Example:

image

Basic font-set to try out:

[CaskaydiaCoveInverse.zip (10MB)](https://github.com/ryanoasis/nerd-fonts/files/13997450/CaskaydiaCoveInverse.zip)

```
Archive:  CaskaydiaCoveInverse.zip
  Length      Date    Time    Name
---------  ---------- -----   ----
  2198512  2024-01-20 10:01   CaskaydiaCoveNerdFont-BoldItalic.ttf
  2349184  2024-01-20 10:01   CaskaydiaCoveNerdFont-Bold.ttf
  2197532  2024-01-20 10:01   CaskaydiaCoveNerdFont-Italic.ttf
  2122672  2024-01-20 10:01   CaskaydiaCoveNerdFontMono-BoldItalic.ttf
  2270856  2024-01-20 10:02   CaskaydiaCoveNerdFontMono-Bold.ttf
  2121692  2024-01-20 10:01   CaskaydiaCoveNerdFontMono-Italic.ttf
  2269344  2024-01-20 10:02   CaskaydiaCoveNerdFontMono-Regular.ttf
  2347672  2024-01-20 10:01   CaskaydiaCoveNerdFont-Regular.ttf
---------                     -------
 17877464                     8 files
```

trashner:
I tested the provided patched fonts in Windows Terminal/PowerShell.

The new glyphs are available at the given code points, they look seamlessly alongside other glyphs and the background transparency works as well!

Finii:
Thanks for reporting!

Finii:
Now I get what you meant with transparency 🤦‍♀️ And I have something similar myself in the terminal I work with every day (it is semi-transparent so that I can see windows underneath it) but I failed to make the connection.

Thanks for testing!

That the corner gets rounded is indeed ... interresting :-D
But maybe the used point type is wrong (I need to check, I drew the glyphs with Fontforge, I should have used Glyphs3 😬).

And isnt there an option in Windows Terminal to change the renderer?

Finii:
Ah, you see the individual glyphs do have sharp points.
I believe the reason is that the next glyph has overlap (to avoid the ugly gaps between the glyphs, the dreaded thin vertical lines that result in color subpixel rendering...

Screenshot 2024-01-22 at 07 51 19

Here the "landing platform" left of the 'cell', that will cut off the arrow tip of the glyphs to the left of it...

Screenshot 2024-01-22 at 07 54 52

a-usr:
> And isnt there an option in Windows Terminal to change the renderer?

Yes, but that doesnt really change anything. Also I would add some more information with a link to a comment by one of the Windows Terminal Devs explaining why, but i dont have the link rn, so maybe I'll add that later. 
Furthermore, there might actually be an issue with the glyph. I did a quick test with cmder 

With E0B0 and inverted Colors:
image


With E0D7: 
image

While writing this comment @Finii sent a comment, wich is why I'll just reply here.
> Ah, you see the individual glyphs do have sharp points.
> I believe the reason is that the next glyph has overlap (to avoid the ugly gaps between the glyphs, the dreaded thin vertical lines that result in color subpixel rendering...

That might be the reason for the overlap. Windows Terminals Renderer, and maybe also ConEmu (The underlying program used by CMDer) dont crop glyphs to the correct size (as far as I remember).

Finii:
Cascadia Cove has even bigger landling platforms:

image

Maybe we should decrease ours a bit - but if they are too small other people will complain about the faint vertical subpixel lines...
* #1419 
* #780 
* #1245

Finii:
Maybe for your use-case we should also add those glyphs without landing platforms.
That could be also automated as CALTs; but maybe just to have the glyphs with different codepoints would be enough.

That is a classical target conflict :thinking:

a-usr:
> Maybe for your use-case we should also add those glyphs without landing platforms. That could be also automated as CALTs; but maybe just to have the glyphs with different codepoints would be enough.
 I don't have that much knowledge regarding fonts and glyphs, could you please ellaborate on the term CALT ?
 
> That is a classical target conflict 🤔

Sadly yes..

However I might have a solution, although it might be a rather dirty and tedious one.

You see, as far as I know, (please do correct me if I'm wrong) Windows Terminal is having the most issues with the overlap on powerline glyphs provided by nerd font. 

The reason for that is, that WT uses the so called "glyph advane width" to determine the width that a glyph should be.
This is better explained by [this comment](https://github.com/microsoft/terminal/issues/8993#issuecomment-1593840202)

As per my understanding, this glyph advance width is determined using the width of the character "0".

So, I propose to either manually, or automatically change the widtht of the 0 character in all nerd fonts. As I said, this feels wrong.

But I believe that with the correct width this problem would be kind of solved (?) 

However youre the expert on this, so I would love to hear your opinion on this

Finii:
> We could also move the points a bit inward

Well, that would break the current glyph scaling algorithm; it has no possibility to have a positive overlap on one side and a negative on the other. :-(
And the scaling algo is already very complex with a lot rules for special cases that can be selected
```
0xe0cc: {'align': 'l', 'valign': 'c', 'stretch': 'xy2', 'params': {'overlap': 0.02, 'xy-ratio': 0.85}},
0xe0ce: {'align': 'l', 'valign': 'c', 'stretch': 'pa', 'params': {}},
'trig': {'align': 'c', 'valign': 'c', 'stretch': 'pa1!', 'params': {'overlap': -0.10, 'careful': True}}
```
_Example rules_

a-usr:
> Well, that would break the current glyph scaling algorithm; it has no possibility to have a positive overlap on one side and a negative on the other. :-( And the scaling algo is already very complex with a lot rules for special cases that can be selected

but does the Inverse Triangle even need a negative overlap?

Finii:
Ah sorry I edited your comment, trying to undo.

Finii:
> The reason for that,, is that WT uses the so called "glyph advane width" to determine the width that a glyph should be.

Thanks for linking the Windows Terminal Issue.

Well, all terminals base their cell width on the advance width, that is the "width of the cell", there is no other measure in the font for that purpose.

Differing from the claims in the linked Issue the powerline glyphs in Nerd Fonts are "perfectly scaled". They are bigger than the cell specifically to avoid the thin vertical lines that are due to LCD subpixel rendering. On todays displays everyone should be able to use grayscale subpixel rendering and THAT would work flawlessly. That is the reason why Apple ditched LCD rendering some years ago and now only uses greyscale - it just avoids so many problems.

> change the widths of the 0 character in all nerd fonts

Well, a lot people will complain that the patched font does not look exactly as the unpatched font - as the cell width and with that the line length will differ :grimacing: 
Anyhow we can scale the powerline glyphs as we want. We could scale them to exactly fit one cell (without overlap), but that will create thin color lines (on all platforms btw) when LCD subpixel rendering is active.

Windows is especially bad in this regard because (at least to my knowledge) there are no clear settings, but one has to walk through this obscure ClearType settings sequence where no clear indication is given which subpixel rendering will be the result. Apart from the fact that ClearType itself is broken even when that settings dialogue has been mastered.

But that is in principle no Windows problem but a general LCD subpixel rendering problem.

a-usr:
> Ah sorry I edited your comment, trying to undo.

I noticed, you got me really confused there 🤣

Finii:
> but does the Inverse Triangle even need a negative overlap?

That is a good question. The overlaps are needed if you switch background and foreground color to continue a colored line that starts with a Triangle.. like...:

image

Imagine you do not want the coffee icon and all the grey stuff on the left side.
The line will start with some (transparent) blanks, than a inverse triangle in blue, then ... a blue full block or a space with background blue, then the tilde with background blue ...
I have no clue how this transparent background color works ;)

But if you have something you might get a thin colored line between the inverse triangle and the full-block / blank, if there is no overlap.

Is that even possible with transparency?

And then. Assuming the Inverse Triangle does not need an overlap because it is used solely in transparent contexts where never ever a overlap is needed.
We still have the problem with the ordinary Triangle overlap hanging into the Inverse Triangle cell and making the point blunt. (see my hand drawn images above).

Finii:
image

That PR aimed at 4% overlap (from 2%), and was moved to ... parallel universe.

And then later, we have this:
* #1419 

where the overlap is increased to even 6% !!!

Maybe we should unroll that all and use 2% as decided before, people with vertical lines need to change to greyscale :woman_shrugging:

Finii:
I guess if we go back to 2% the issue you see is still there but by far less pronounced. Maybe then you can life with that, as all solutions are compromises. What do you think, is that worth a try? (I would need to edit all the glyphs to reduce the landing and then change the percentages in the patcher itself - that takes some time so I want to avoid it if that has not even a change to be fruitful....

Finii:
BTW, _personally_ I really like the rounded tips :-D

Finii:
I would revert/change #1419 anyhow, to 4%. It is just that a lot other fonts with their own powerline glyphs have a very big landing patch, Hack, Cascadia Code PL, ... up to 12% iirc; and I thought if they can successfully live with that ... but now I see again the downsides of such a big overlap.
Maybe 3% is good, it covers a bit more LCD subpixel rendering problems (i.e. vertical lines) and still be less obvious in your setting. The 3% was almost invisible lines with most colors iirc, so the vertical line situation would be improved nonetheless.

a-usr:
> > but does the Inverse Triangle even need a negative overlap?
> 
> That is a good question. The overlaps are needed if you switch background and foreground color to continue a colored line that starts with a Triangle.. like...:
> 
> image
> 
> Imagine you do not want the coffee icon and all the grey stuff on the left side. The line will start with some (transparent) blanks, than a inverse triangle in blue, then ... a blue full block or a space with background blue, then the tilde with background blue ... I have no clue how this transparent background color works ;)
> 
> But if you have something you might get a thin colored line between the inverse triangle and the full-block / blank, if there is no overlap.


Could we clarify real quick in wich directions positive and negative overlap go respectively? I was imagining the following:


`Negative Overlap | Glyph | Positive Overlap`

> but does the Inverse Triangle even need a negative overlap?

What I meant with that is wether u/E0D7 needed an overlap like this:
`Overlap | u/E0D7 | Overlap`

If something like this would be sufficcient:
`No Overlap | u/E0D7 | Overlap`


> I have no clue how this transparent background color works ;)

Its not that difficult really, and I have even tested this behaviour just now. If a glyph that needs to be rendered has no background color associated with it, the background is made transparent. If the glyph has any background color associated, the background has that color, regardless of wether it matches the background or not. At least Windows Terminal behaves like this at the time of writing this comment. And the transparency Is set within the Windows Terminal Settings, either for all profiles (that dont specify their own), or per profile.

Finii:
> Could we clarify real quick in wich directions positive and negative overlap go respectively?

Positive Overlap: Real overlap, resulting glyph overlaps other glyphs cell
Negative Overlap: Glyph has an empty border to the other cell, the opposite of "being bigger than the cell we should fit in": "be smaller than the cell".

The Overlap works in all 4 directions up/down/left/right. Well, with exceptions and specials, but that is the general concept.

Have 1% of overlap into all (top buttom left right) cells to cover up all these round-error-LCD-rendering problems.

a-usr:
~Also, I just noticed, there might be an issue with the u/e0b0 Triangle~ 
image

Ignore that, the Inverse Triangle doesnt have negative padding

Finii:
Proposal:
I will reduce the overlap to 3 or 4 %, and make a new test font set.
I believe now the 6% overlap should be reverted anyhow.

But that will take some hours, need to simulate some work for my real life job now ;-D

Finii:
image

* It seems to be transparent for fore- and background, I can see the texture everywhere
* The landing platform is ... missing?

a-usr:
> image
> 
> * It seems to be transparent for fore- and background, I can see the texture everywhere
> * The landing platform is ... missing?

I edited the comment this screenshot was from, sorry for the Confusion I may have caused.

By toying around I noticed that CMDer only crops unneccesary padding on the side of a glyph if the foreground of the glyph next to it doesnt match with the glyph whose padding is to be cropped

Proof (In CMDer):
image

image

a-usr:
Upon further Inspection, The Inverse Triangle should definetly be moved to the Left (in CMDer):
image

a-usr:
> Proposal: I will reduce the overlap to 3 or 4 %, and make a new test font set. _Edit: Maybe even only 2?_ I believe now the 6% overlap should be reverted anyhow.

We could try that too.

> But that will take some hours, need to simulate some work for my real life job now ;-D

Thats a good Idea, and I should probably do the same.

a-usr:
> What is `CMDer`? Is that the old terminal one started with `cmd`? Is that not removed on Windows11? At least I had the impression Windows11 always has `Windows Terminal` instead of `cmd`?

CMDer is a (more or less) standalone and portable Console Emulator indipendent from Microsoft (I think)

[Website](https://cmder.app)
[GitHub](https://github.com/cmderdev/cmder)

Official Image:
image

a-usr:
> What is `CMDer`? Is that the old terminal one started with `cmd`? Is that not removed on Windows11? At least I had the impression Windows11 always has `Windows Terminal` instead of `cmd`?

## (unnecessary info, if you dont care just skip this)

The terminal (or Console Emulator, Terminal Emulator to be precise, if you want I can try to explain why it is a terminal EMULATOR and not a terminal / what terminal actually means) that you're talking about is conhost.exe. cmd is a so called Shell (like powershell). A Shell needs a Terminal Emulator (or an actual terminal) to interact with the user (because it usually doesnt have any GUI). 
When starting cmd.exe by either double clicking on the file in the C:\system32 folder or using the windows run dialog Windows first starts the default terminal Emulator and then connects the cmd.exe process to it. And as soon as the process is connected to the terminal emulator it can send characters to it.

a-usr:
Also, hurmit nf is another font that I really like, so would it be possible for you to patch that one instead when youre done figuring out wich way to solve this little issue?

a-usr:
Hey, any updates @Finii ?

Finii:
> Hey, any updates @Finii ?

With the upcoming release it would be good to have this in.
I need to pencil this down on paper to find some solution that does not break the 'ordinary' / normal Powerline glyphs and still solves the overlapping problem with the new ones. :thinking:

a-usr:
> With the upcoming release it would be good to have this in. I need to pencil this down on paper to find some solution that does not break the 'ordinary' / normal Powerline glyphs and still solves the overlapping problem with the new ones. 🤔

I have been using the new glyphs for the past 2 months now, and if I'm being honest, I quite like the Horizontal overlap now. As you said earlier, the round-ish look has something unique. Sadly not everyone likes that look. But I think it would be nice to keep the PL glyphs with the overlapping problem somewhere in the repo so that people could use them instead of the fixed PL glyphs to patch their fonts if they want

Finii:
Rebase on master, force push

Finii:
@allcontributors
please add @a-usr for review and idea
please add @trashner  for review

Finii:
@allcontributor please add @trashner for review

Finii:
@allcontributors please add @trashner for review

Finii:
How it looks like without transparence (`tilix`)

image
_no hinting, LCD antialiasing_

image
_full hinting, LCD antialiasing_

Finii:
I think we should use this PR for now, and can develop further over time. If this is in the release maybe more people will feedback.

CEbbinghaus:
Thank you so much for adding this Icon. Absolute life savior (wish there was an easier way to find out about it than this PR). I have noticed a minor problem which is that it seems that the icon is 1px too tall:
image
image

a-usr:
Does that also happen with the non-inverse arrow head? I dont think the inverse version gives you any benefits in this use case, but feel free to correct me

CEbbinghaus:
@a-usr it does not happen with the regular arrow head (as far as I can tell). The inverse arrow does give me a deciding benefit since I do use the transparency (just outside these screenshots). It also keeps all of my separators consistent

a-usr:
> @a-usr it does not happen with the regular arrow head (as far as I can tell). The inverse arrow does give me a deciding benefit since I do use the transparency (just outside these screenshots). It also keeps all of my separators consistent 

I see. In that case you'll have to wait until @Finii  finds time to check on his GitHub account. But since that might take some time, if the overlap really annoys you I recommend that you use the non-inverted arrow as much as possible for the time being

Finii:
Hmm, what is your OS and terminal emulator?
And which font do you use (if possible where did you download it, exact version number would also be great).

The problem is that the 'top' of the glyph is defined by two different means, the normal arrow just is as high as one line; while the inverted arrow ends in a outline on top. The coordinates are handled differently by font renderers usually - they need to be rounded and maybe subpixel rendered.

If you are on a system where you can select the subpixel rendering it might be an idea to change that.

But lets check for some random current font...

image

The top outline is at vertical position 1912 and the bottom (not shown) at -492.

And the line coordinates are:

image

### Ah yes!

There is a 12 unit difference which is about 0.5%, and we see about 1% (i.e. 1 pixel) in your screenshot.

Vertical overlap is designed as 1% in the code:

image

Are you willing/available to try a variant here, with 0% overlap? I am not sure that will introduce a subpixel problem :thinking:

CEbbinghaus:
Happy to try out a variant with 0% overlap. I wonder if that will help fix my problem. 
I also noticed that the inverse + regular don't line up correctly such that they form a perfect arrow:
image

As for my font I use the **FiraCode** variant of NerdFont. I believe I have the latest version installed (I use Scoop and it claims 3.2.1). As for my Terminal Emulator I use **Windows Terminal V1.20.11781.0** using **Powershell** as my shell with **Starship** for the Prompt

Finii:
> also noticed that the inverse + regular don't line up correctly such that they form a perfect arrow

That problem is already mentioned right in the top of this thread/PR, with a possible fix given in https://github.com/ryanoasis/nerd-fonts/pull/1490#issuecomment-1903418143 above (I sketched exactly that image on paper there ;)

The culprit are the 'landing platforms' on the left resp right that help avoid the vertical colored lines problem.
Are these adjacent triangular things a common setup?

I'll add the new patched font into this comment in a minute...

CEbbinghaus:
Just tried the patched font. It is a big improvement over the previous in that it has very little offset. However there seems to still be 1 subpixel that is wrong. It feels offset downwards by .5 of a pixel:
image

This is 1 physical pixel too low vertically on a 4k monitor so its far better and barely noticable. But its existence is messing with me.

Overall a success though.

This does make the Powerline arrow stand out with its offset though and clearly shows the box you were talking about:
image

Finii:
Hmm. The overlap is there to prevent the subpixel-rendering issues.
I must admit that I never saw any problem with horizontal lines that show up. As a result I decreased the vertical overlap to 1% some time back. Maybe it is even not needed at all. The overlap has been there since ... ancient times. Buit changing that would require a broad survey of terminal emulators and how they would handle the font without vertical overlaps.
:thinking: 
Windows Terminal is one of the easier terminals to generate fonts for; kitty for example has much stronger opinions how fonts shall look like and squashes the glyphs into submission if one of their assumptions is violated :grimacing: 

Well, if I look at my current text size here, with the release font, I spot some small problems but they seem to have too little overlap

image
_Tilix on X Windows_

With a bigger font size the effects change

image

image


Possibly this is because the line 'boundaries' are always on a pixel boundary (i.e. it's either yellow or white, no subpixel rendering), while the glyphs outline is subpixel rendered to fall exactly on a specified coordinate and rendered lighter if it does not match.

But you can clearly see, `tilix` seems to render the inverted triangle rather too small albeit the 1% overlap is still in place in the font I have installed. So without it it would look even more ugly - on `tilix`. So this is a compromise how the powerline glyphs that have a filled border to the line above/below behave and can not be solved perfectly for all.

CEbbinghaus:
Shouldn't the glyph simply be the height if 1em (or whatever term there is for the bounding box for a char in a font). The same amount of vertical space that a space takes up.

I have never created a font before so I have no idea how it works but I am surprised with how much difference there is between font renderers in the various terminal applications

Finii:
The 'problem' with the powerline glyphs is that they are supposed to fill exactly "one line".

In the usual fonts here we have a 'advance width' for each glyph, that indicates how far the cursor moves to the right to draw the next glyph. There is no such easy thing for the baseline to baseline distance, as the line height is called.

As I showed above here: https://github.com/ryanoasis/nerd-fonts/pull/1490#issuecomment-2296382746

> The top outline is at vertical position 1912 and the bottom (not shown) at -492.

In the font file there are three 'systems' to specify the baseline to baseline distance, and upon patching Nerd Fonts removes any `gap` (as that complicates everything beyond any reasonable reason;) and makes the 3 systems all alike. Usually they are NOT in the unpatched fonts.

```
=== patched-fonts/CascadiaCode/Regular/CaskaydiaCoveNerdFontMono-Regular.ttf ===
Version 2111.001; VTT 6.35;Nerd Fonts 3.2.1
SHA1: 380a8c3bd92efda8e7ac47959244a996bcf5578d

::::::::::::::::::::::::::::::::::::::::::::::::::
  Metrics
::::::::::::::::::::::::::::::::::::::::::::::::::
[head] Units per Em:   2048
[head] yMax:           2304
[head] yMin:          -989
[OS/2] CapHeight:      1420
[OS/2] xHeight:        1060
[OS/2] TypoAscender:   1900
[OS/2] TypoDescender: -480
[OS/2] WinAscent:      1900
[OS/2] WinDescent:     480
[hhea] Ascent:         1900
[hhea] Descent:       -480

[hhea] LineGap:        0
[OS/2] TypoLineGap:    0

::::::::::::::::::::::::::::::::::::::::::::::::::
  Ascent to Descent Calculations
::::::::::::::::::::::::::::::::::::::::::::::::::
[hhea] Ascent to Descent:              2380
[OS/2] TypoAscender to TypoDescender:  2380
[OS/2] WinAscent to WinDescent:        2380

::::::::::::::::::::::::::::::::::::::::::::::::::
  Delta Values
::::::::::::::::::::::::::::::::::::::::::::::::::
[hhea] Ascent to [OS/2] TypoAscender:       0
[hhea] Descent to [OS/2] TypoDescender:     0
[OS/2] WinAscent to [OS/2] TypoAscender:    0
[OS/2] WinDescent to [OS/2] TypoDescender:  0

::::::::::::::::::::::::::::::::::::::::::::::::::
  Baseline to Baseline Distances
::::::::::::::::::::::::::::::::::::::::::::::::::
hhea metrics: 2380
typo metrics: 2380
win metrics:  2380
```

But note, this is the font I use in `tilix`.
* The b2b is for all 3 systems 2380
* The bounding box height of `E0D7` is 2404 (see linked comment)
* And still it is rendered smaller than the colored line distance

I believe the problem is that the actual font renderer does not care about lines as such - it does work also with glyphs that span multiple 'lines' with swashes, like this:

image

All parts of the glyph, the outline, this is drawn, and it is drawn with a resolution higher than one pixel. Using only the pixel resolution make the font look very jagged so noone nowadays does not use subpixel rendering. It makes pixel more or less black/grey to create a visually pleasing round form perceptionally.

But then terminal emulators have the concept of 'cells' (that we also use here, but that is nonexisting in fonts). And they for example want to fill one line with a yellow and one line with a white background. The emulator can calculate the exact baseline boundary, but that will be on some partial pixel, like 10.213 pixels down. What to do now? I have never seen subpixel rendering for background coloring, so it just fills all background pixels in pixel line 0 to 10 with yellow and from 11 to 20 with white. Obviously that does not coincide with the glyph subpixel exact boundaries. The glyph says it shall fill from 0 to 10.213 with black, and it renders this as 0-10 in black and 11 in some light grey, 21% grey presumably. And now you have the effect you see, the glyph seems to be too tall.

I would guess the terminal emulators push the line distances to be an exact integer number of pixels, to get very sharp boundaries if the background color changes. It then places the glyphs into THAT lines, more or less ignoring what the font's baseline to baseline distance is, because it can not possibly match.
Tilix seems to round that up typically (to be never too narrow), while Windows Terminal maybe rounds exactly or always down, or whatever.

If one things about fonts one has to forget 'cells'. This is not how fonts work. And font rendering is always designed to work in a 'proportional' environment like for example `Writer` or `Word`. With terminal emulators the "lines" and such is just bolted on top and is implemented _by the terminal emulator_ differently everywhere.
Some for example do not allow the glyphs to extrude outside the 'cell' and they just scale the glyph down until it fits. This breaks any alignment of course and looks more or less ugly if glyphs extrude into the neighboring 'cells' like the powerline glyph landing platforms (that are needed to prevent the vertical colored lines that come from the subpixel rendering :grimacing: )

Btw I have (almost) no clue how terminal emulators really do their job; all what I know is by observing their behavior to different font details. Occasionally I look into code but I try to avoid it. I come from a publishing background so I know rather well how the font renderer works - on a blank canvas like a sheet of paper ;-)

PR #1497, opened 2024-01-26
Update all Iosevka variants to version 29.0.4

iruoy:
#### Description
Bring all Iosevka versions up to date

#### Requirements / Checklist

- Update Iosevka to version 28
- Update the filenames since they're now in CamelCase


Finii:
Thanks for the nice PR!

I would put this on hold until directly before the Nerd Fonts v3.2 release, as there might be another Iosevka release in between.

And then one question:
You put all the weights into individual directories for `IosevkaTermSlab`, where it was one big flat directory before.
Is there a reason for that, except because the other `Iosevka`s have it and you wanted to unify?

Do you download the fonts from the website for installation?

The flat setup in the newer variant - do you see advantages in the 'deep' setup?

My reasoning was that grouping by weight is not very useful, as people usually want all weights but only one variant (Mono, Propo). Did not finish that redesign, but that is where the flat source (and patched) directories come from.

What I mean:

```
.../patched_fonts/
                  Iosevka/
                          Nerd Font/         (all the weights, flat)
                          Nerd Font Mono/    (all weights, flat)
                          Nerd Font Propo/   (you get the idea)
```

In the release packages they are always flat anyhow.

iruoy:
Hey, thanks for your extensive comment.

First off, I think it's a good idea to hold off on merging this until right before a new version is tagged, because Iosevka releases so often. There's even been a new release since I made this PR.

The reason for the changes to `IosevkaTermSlab` was indeed to synchronize the workflow across the three Iosevka folders. I still think that's a good thing to do. Apparently, I settled on the wrong concept though. Tomorrow, I'll try to wrangle these configs into generating the folder structure you're looking for.

Finii:
> I still think that's a good thing to do.

What do you mean by 'it'? Unifying or having a deep-weights-directory-structure?
If the second, can you explain? Just because I once thought it would be a good idea does not make it the best solution.
Every discussion welcome.

Finii:
Updated this PR to use 29.0.4 which is current.
Also drop patched fonts, these are created by the release process.
Also some small other changes.

Finii:
@allcontributors please add @iruoy for code

Issue #1507, opened 2024-02-04
10x times faster docker patch


nobk:
I am using docker run nerdfonts/patcher for patch my custom iosevka fonts.
But I find that the fontforge command line is single process, only one CPU core are using.
So I changed the docker script, make it use all CPU cores.
And get 10x times faster when run docker patch on an i7-12700K.
Only need 2 lines modification.

Finii:
I have only a Ryzen7 3700U so I would expect a speedup of 3.5, but that still is significant!

:tada:

Well, that's the reason why I use `gotta-patch-em` with `-j`. That also has the advantage to read the config file.
I guess we should someday unify the config files and incorporate them into the patcher itself. ... Sigh

PR #1508, opened 2024-02-04
Parallel execution fontforge in docker

nobk:
#### Description

Improve the docker execution method, optimize from single process to parallel execution, and increase the speed of font patching.
see [issue 1507](https://github.com/ryanoasis/nerd-fonts/issues/1507)

#### Requirements / Checklist

Speed up docker patcher

### DIFF
--- a/bin/scripts/docker-entrypoint.sh
+++ b/bin/scripts/docker-entrypoint.sh
@@ -23,6 +23,6 @@ done
 printf "Running with options:\n%s\n" "$args"
 
 # shellcheck disable=SC2086
-for f in /in/*.otf /in/*.ttf /in/*.woff /in/*.eot /in/*.ttc; do [ -f "$f" ] && fontforge -script /nerd/font-patcher -out /out $args "$f"; done
+find /in -type f \( -name "*.otf" -o -name "*.ttf" -o -name "*.woff" -o -name "*.eot" -o -name "*.ttc" \) | parallel fontforge -script /nerd/font-patcher -out /out $args {}
 
 exit 0
--- a/bin/scripts/docker-entrypoint.sh
+++ b/bin/scripts/docker-entrypoint.sh
@@ -23,6 +23,16 @@ done
 printf "Running with options:\n%s\n" "$args"
 
 # shellcheck disable=SC2086
-find /in -type f \( -iname "*.otf" -o -iname "*.ttf" -o -iname "*.woff" -o -iname "*.eot" -o -iname "*.ttc" \) | parallel fontforge -script /nerd/font-patcher -out /out $args {}
+if [ "$PN" -eq 1 ]; then
+	find /in -type f \
+	  \( -iname '*.otf' -o -iname '*.ttf' -o -iname '*.woff' -o -iname '*.eot' -o -iname '*.ttc' \) \
+	  -exec fontforge -script /nerd/font-patcher -out /out $args {} \;
+else
+	njob=""
+	[ "$PN" -gt 1 ] && njob="-j $PN"
+	find /in -type f \
+	  \( -iname '*.otf' -o -iname '*.ttf' -o -iname '*.woff' -o -iname '*.eot' -o -iname '*.ttc' \) \
+	  | parallel $njob fontforge -script /nerd/font-patcher -out /out $args {}
+fi
 
 exit 0
--- a/bin/scripts/docker-entrypoint.sh
+++ b/bin/scripts/docker-entrypoint.sh
@@ -23,6 +23,6 @@ done
 printf "Running with options:\n%s\n" "$args"
 
 # shellcheck disable=SC2086
-find /in -type f \( -name "*.otf" -o -name "*.ttf" -o -name "*.woff" -o -name "*.eot" -o -name "*.ttc" \) | parallel fontforge -script /nerd/font-patcher -out /out $args {}
+find /in -type f \( -iname "*.otf" -o -iname "*.ttf" -o -iname "*.woff" -o -iname "*.eot" -o -iname "*.ttc" \) | parallel fontforge -script /nerd/font-patcher -out /out $args {}
 
 exit 0
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ LABEL org.opencontainers.image.title="Nerd Fonts Patcher" \
       org.opencontainers.image.source="https://github.com/ryanoasis/nerd-fonts" \
       org.opencontainers.image.licenses="MIT"
 
-RUN apk update && apk upgrade && apk add --no-cache fontforge --repository=https://dl-cdn.alpinelinux.org/alpine/latest-stable/community
+RUN apk update && apk upgrade && apk add --no-cache fontforge parallel --repository=https://dl-cdn.alpinelinux.org/alpine/latest-stable/community
 
 ENV PYTHONIOENCODING=utf-8
 
--- a/readme.md
+++ b/readme.md
@@ -365,9 +365,19 @@ Patching the font of your own choosing:
   ```
 
 * Use docker
+  * Default parallel tasks
   ```
   docker run --rm -v /path/to/fonts:/in:Z -v /path/for/output:/out:Z nerdfonts/patcher [OPTIONS]
   ```
+  * Single process (slow)
+  ```
+  docker run --rm -v /path/to/fonts:/in:Z -v /path/for/output:/out:Z -e "PN=1" nerdfonts/patcher [OPTIONS]
+  ```
+  * Specify the parallel tasks number to 10
+  ```
+  docker run --rm -v /path/to/fonts:/in:Z -v /path/for/output:/out:Z -e "PN=10" nerdfonts/patcher [OPTIONS]
+  ```
+
 
 > [!NOTE]
 > The resulting font's family (aka font name) will be set to the original family after CamelCasing, removing whitespace and appending ` Nerd Font`. For example, `iosevka term` would become `IosevkaTerm Nerd Font`.


bin/scripts/docker-entrypoint.sh
```diff
@@ -23,6 +23,6 @@ done
 printf "Running with options:\n%s\n" "$args"
 
 # shellcheck disable=SC2086
-for f in /in/*.otf /in/*.ttf /in/*.woff /in/*.eot /in/*.ttc; do [ -f "$f" ] && fontforge -script /nerd/font-patcher -out /out $args "$f"; done
+find /in -type f \( -name "*.otf" -o -name "*.ttf" -o -name "*.woff" -o -name "*.eot" -o -name "*.ttc" \) | parallel fontforge -script /nerd/font-patcher -out /out $args {}
```
Finii, line 26: I like `find` much more than the loop. We could use `-iname` here, as some fonts out there have their name and extension in caps. (Which would be an additional commit of course ;)

Finii (approved):
Thanks for the nice PR!

I did never put much effort in the docker patcher ;-) because I do not use it at all.
But than means that suggestions like yours are so valuable!

I think this is a very good improvement. The `gotta-patch-em` also can parallel-patch (but with a hand-rolled parallelism and a hard wired number of jobs).

I would allow people to specify the `-j` option, as not all want parallel patching.
Maybe also helpful would be some output if find turned up with nothing.

Tell me if you want to expand this or if you prefer me merging as-is.

bin/scripts/docker-entrypoint.sh
Finii: Probably I would use `-iregex` but that is because I'm a regex person and hate the `-o` style in `find` ;-)

Finii:
```suggestion
find /in -type f -iregex ".*\.\(otf|ttf|woff|eot|ttc\)" | parallel fontforge -script /nerd/font-patcher -out /out $args {}
```
_untested_

Finii:
@allcontributors please add @nobk for code

Finii:
As a side note, I did never check if the debug log will be even avaiable when docker is used.
Note to self: Check that and maybe add moving the logfile to `/out`.

nobk:
> I would allow people to specify the `-j` option, as not all want parallel patching. Maybe also helpful would be some output if find turned up with nothing.
> 
> Tell me if you want to expand this or if you prefer me merging as-is.

I will add an option for  `-j PN`, with docker option `-e "PN=1"`, to disable parallel execute.
```
docker run --rm -v /path/to/fonts:/in:Z -v /path/for/output:/out:Z -e "PN=1" nerdfonts/patcher [OPTIONS]
```

bin/scripts/docker-entrypoint.sh
nobk: Dockerfile's `find` is a BusyBox multi-call, which is not support `-iregex`, I will change it with `-iname`.

nobk:
I think docker patcher is for users, not for developers, so I am not test logfile output.
When I set `-e "PN=20"` , all hyper threads of i7-12700K CPU is used, up to 4.9GB RAM used, speed up max.

Finii:
Thank you! Appreciate your work :green_heart:

Finii:
image

Fixing

Finii:
```console
$ shellcheck docker-entrypoint.sh

In docker-entrypoint.sh line 28:
	  -exec fontforge -script /nerd/font-patcher -out /out $args {} \;
                                                               ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
	  -exec fontforge -script /nerd/font-patcher -out /out "$args" {} \;


In docker-entrypoint.sh line 34:
	  | parallel $njob fontforge -script /nerd/font-patcher -out /out $args {}
                     ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.
                                                                          ^---^ SC2086 (info): Double quote to prevent globbing and word splitting.

Did you mean: 
	  | parallel "$njob" fontforge -script /nerd/font-patcher -out /out "$args" {}
```

nobk:
no, that warnning caused by BusyBox multi-call `sh`. If the environment variable PN is not defined, this warning will appear.
```
docker run -it --rm alpine:latest sh
/ # [ "$PN" -eq 1 ]
sh: out of range
/ # which sh
/bin/sh
/ # /bin/sh --version
/ # ls -l /bin/sh
lrwxrwxrwx    1 root     root            12 Jan 26 17:53 /bin/sh -> /bin/busybox
/ #
```

nobk:
Adding one more line of code eliminates those two warnings.
`[[ -z "$PN" ]] && PN=0`
need me submit a PR for this?

Finii:
no, thanks :-)

nobk:
My fix is like this, and not submit that then, it just warning.
```
git diff
diff --git a/bin/scripts/docker-entrypoint.sh b/bin/scripts/docker-entrypoint.sh
index acd931a..bb852d3 100644
--- a/bin/scripts/docker-entrypoint.sh
+++ b/bin/scripts/docker-entrypoint.sh
@@ -23,6 +23,7 @@ done
 printf "Running with options:\n%s\n" "$args"

 # shellcheck disable=SC2086
+[[ -z "$PN" ]] && PN=0
 if [ "$PN" -eq 1 ]; then
        find /in -type f \
```

Finii:
I added two more commits, hope you find them ok.

1fccd8a63edb256d7920e738bc190ccb55c36187 `docker: Allow blancs in filenames`
7ebbc4e1733663307face4174d785a138371bdeb `docker: Include logfile in output`

That also fixes the shellcheck warning and the runtime warning.

This is a very good addition! Thanks again for the idea and implementation/PR!

Finii:
> I think docker patcher is for users, not for developers, so I am not test logfile output.

Well, that output helps (me) when those users have problems with some particular font.
They can report with the debug messages and I can possible see what goes wrong.
This is important for fonts that are not free, i.e. I can not download and test myself.

For example
* "After patching the gap is too wide - I can not share the font file"
* "Please run with `--debug 1`"
* In the debug log: `Gap detected ...`

nobk:
As you changed the default value of PN to 1,  you need modifiy readme.md docker usage section, default singale process, `-e "PN=0"` let `parallel` decide the number of tasks.
But I fell default value of PN as 0 will be better because all docker user can benefit by speed up from this upgrade.
Unless you need debugging or there is insufficient memory, you need to set PN to 1.

Finii:
I didnt intend to change your default, though :grimacing: 
Rewrote that several times, maybe got confused ;-)
Fixing.

Thanks for reporting. I believe `-j` is the better default.

Finii:
96497b4fe `docker: Run parallel by default`

nobk:
> Thanks for reporting. I believe `-j0` is the better default.

I am not using -j0 in my PR, I just omitted the -j parameter in shell script by default or NP=0, `parallel` generate 8 tasks for me, I can see them in `htop`.
Your latest commit with parallel -j0 I just tested, `parallel` generate 32 tasks for me, because my custom fonts have totally 32 .ttf files, it use 7.8GB memory.
If uses have more font files, I think -j0 will cause out of memory.

nobk:
omitted the -j parameter means `-j 100%`
`man parallel`
```
       --jobs num
       -j num
       --max-procs num
       -P num
           Number of jobslots on each machine.

           Run up to num jobs in parallel. Default is 100%.

           num    Run up to num jobs in parallel.

           0      Run as many as possible (this can take a while to determine).

                  Due to a bug -j 0 will also evaluate replacement strings twice up  to  the  number  of
                  joblots:

                    # This will not count from 1 but from number-of-jobslots
                    seq 10000 | parallel -j0   echo '{= $_ = $foo++; =}' | head
                    # This will count from 1
                    seq 10000 | parallel -j100 echo '{= $_ = $foo++; =}' | head

           num%   Multiply  the  number  of  CPU threads by num percent. E.g. 100% means one job per CPU
                  thread on each machine.

           +num   Add num to the number of CPU threads.

           -num   Subtract num from the number of CPU threads.

           expr   Evaluate expr. E.g. '12/2' to get 6, '+25%' gives  the  same  as  '125%',  or  complex
                  expressions  like  '+3*log(55)%'  which means: multiply 3 by log(55), multiply that by
                  the number of CPU threads and divide by 100, add this to the number of CPU threads.

                  An expression that evalutates to less that 1 is replaced with 1.

           procfile
                  Read parameter from file.

                  Use the content of procfile as parameter for  -j.  E.g.  procfile  could  contain  the
                  string 100% or +2 or 10.

                  If procfile is changed when a job completes, procfile is read again and the new number
                  of  jobs is computed. If the number is lower than before, running jobs will be allowed
                  to finish but new jobs will not be started until the wanted number of  jobs  has  been
                  reached.   This  makes  it  possible to change the number of simultaneous running jobs
                  while GNU parallel is running.

           If the evaluated number is less than 1 then 1 will be used.

           If --semaphore is set, the default is 1 thus making a mutex.

           See also: --use-cores-instead-of-threads --use-sockets-instead-of-threads
```

Finii:
> omitted the -j parameter means `-j 100%` `man parallel`

But that would require an (additional) if and/or code duplication.

In principle we could

```diff
-parallel --verbose --null "--jobs=${PN}" fontforge -script ../../font-patcher -out ../../out_fonts $args {}
+[-n "$PN"] && jobs=-j${PN}
+parallel --verbose --null "${jobs}" fontforge -script ../../font-patcher -out ../../out_fonts $args {}
```

The problem is that we can not use double `jobs` then, because that passes an empty option (instead of no option, which would be passed without quotes).

image

Your original PR had the `if` and code duplication (`find` twice and `font-patcher` twice), which is not good in the long run.

Not having the params double quoted raises`schellcheck` warnings and should be avoided as much as possible (imho).

That's the reason I used an explicit 100% (i.e. `-j0`) to get it - branchless programming ;-)

Finii:
Debugging is not about debugging the code but debugging the font.

Having code that works (supposedly) but is hard to read does not help anyone in the long run. So looking well IS important.

Sorry that you do not like the changes.

nobk:
The -j0 make max task as many as total font files  can not as default, need avoid.

nobk:
If you want keep the  double qoutes, you can write like this
```sh
jobs="-j 100%"
[-n "$PN"] && jobs="-j${PN}"
```

nobk:
> Sorry that you do not like the changes.

I like correct code refactoring, not the introduction of bugs.

Finii:
image

That still bugs with an obscure error if someone does `-e "PN=test"`. That is not an improvement but making the error feedback worse.

> I like ... not the introduction of bugs

But still you suggest the above code. That will end up as

`parallel --verbose "-j 0" ...`

and that will raise an error because the allowed syntax is only `"-j0"` or `"-j=0"` (for single arguments, and keeping the quotes exactly means that: 1 argument) (and the two argument version `"-j"` `"0"`.)

PR #1531, opened 2024-03-05
Bugfix font preview(s)

Finii:
#### Description
**[why]**
0xProto (and possibly others) use a default font for the previews.
For some reason some are not correct at all. Possibly my own lazyness.

**[how]**
We need to put the font family names in quotes...
Correct the other entries.

#### Requirements / Checklist

Reported-by: Lucy Phipps <landfillbaby69@gmail.com>

See https://github.com/ryanoasis/nerd-fonts/commit/407898321e42394ae5924ea805c9d1b50411c312


Finii:
@allcontributors please add @landfillbaby for bug

Issue #1547, opened 2024-03-18
Straight vertical subpixel rendering e0b4 & e0b6 ./. e0b0 & e0b2


luebking:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

**Experienced behavior:**
Subpixel rendering e0b4 & e0b6 (round powerline caps) is broken in comparism to e0b0 & e0b2

There's a red vertical bar on right end of e0b6 and a blue vertical bar on left end of e0b4
Happens with at least SauceCode and Monofur

The behavior can *somewhat* be mitigated by fumbling with the font sizes (heavily depends on TE and DPI, using urxvt and pixelsize I can precisely control this, but not fix it) and the LCD filter (lcdnone and lcdlegacy generally look "bad", but I can find a font size where it's just about fixed - but then there's an even amount of pixels and the glyph vertically off by one. lcddefault and lcdlight will always render a discolored bar - at least I could not fumble it away but for unreasonably -on 96 DPI- large fonts)

There's no such issue with the angular powerline caps, maybe fixed by https://github.com/ryanoasis/nerd-fonts/issues/156#issuecomment-1278487200 ?

**Expected behavior:**
Subpixel rendering e0b4 & e0b6 to be more like that of e0b0 & e0b2

**Example symbols:**
- e0b4
- e0b6

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - SauceCodeProNerdFont-Regular.ttf
   - MonofurNerdFont-Regular.ttf
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - https://archlinux.org/packages/?q=nerd
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - urxvt
   - alacritty
   - xterm
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - archlinux

Finii:
Thanks for reporting!

[...]

PR #1551, opened 2024-03-20
Increase Powerline rounded overlap

Finii:
#### Description

Similar to
* #1419

this increases the overlap of the rounded Powerline glyphs to 6%

image
_Patch result with the PR applied, note the increased overlap into next/previous 'cell'_
_Green outline is new glyph_

Fixes: #1547 

#### Requirements / Checklist


Finii:
@luebking Do you want to try this out? I can patch you a font for that...

luebking:
I was already looking which branch holds the change and will happily be your guinea-pig, out of pure altruism and totally not to get the rounded caps to work for at least myself ;)

I have fontforge installed and can probably run the patcher locally, if that helps you anything.

Finii:
Running, will upload the result here.

image

Finii:
image

Some subset?

luebking:
```cat fonts.zip | curl -F 'file=@-' 0x0.st```

**Comment by Finii** [comment]
I guess I don't get it...?

**Comment by luebking** [comment]
This would upload "fonts.zip" to https://0x0.st and get you a tiny URL in return.
The quota is 512 MB and the uptime size dependent (bigger files are deleted earlier)

It's the universal weapon of the archlinux forum :)

**Comment by Finii** [comment]
> It's the universal weapon of the archlinux forum :)

Nice thing, good to know.

Anyhow, here just the basic RIBBI set(s)

[SauceCodeProNF-Rounded.tar.gz](https://github.com/ryanoasis/nerd-fonts/files/14670461/SauceCodeProNF-Rounded.tar.gz)

**Comment by luebking** [comment]
Good news ahead, it's much better.

#### urxvt (main testcase because of the more precise size control and main TE)
Even w/ _lcddefault_ there's no subpixel line in even pixel sizes.
Odd pixelsizes retain them until 15px
(this can be a problem because the font is "flatter" for even pixels sizes, at least on small ones)

_lcdlight_ shows the same behavior, but more faint
(esp. on dark backgrounds tolerable, but not perfect. Still what one might opt for for smaller fonts because the filter is, subjectively, superior to lcdlegacy and for small sizes the odd height looks better)

I was unable to get any subpixel line on _lcdlegacy_ or _lcdnone_

The artifact btw. seems a bit more pronounced on e0b6 "(]" than e0b4 "[)" (but that could be color psychology because they'll expose differently colored subpixels)

urxvt at 11px and lcddefault (on the bottom you can just about spot the subpixel line in the blue prompt)
scrot_1710957756


#### A bigger problem on xterm and alacritty
… turns out to be centereing the glyph - for many sizes it'll be offset by one px or not be "round"
(this is almost not a problem w/ urxvt except for some very small sizes)
On 8.5pt alacritty has the glyph 2px too small
scrot_1710962226

=> I'd have to restore the previous fonts to see whether that was a problem there as well, but it _also exists for the triangular caps_.
This problem vanishes at 19pt what at 96dpi should be ~25px what means ~11pt at 160dpi
Otoh I don't get subpixel lines there.

#### In conclusion
HiDPI users will probably not face any problems w/ this variant, on lower resolutions it takes some fiddling but it's possible to edge out solutions.

**Comment by luebking** [comment]
The vertical centering of the glyph is actually an antialiasing problem, only.
These are on alacritty for 8.5pt and 9pt at 96 DPI:
scrot_1711010008
scrot_1711010046

The indention/offset is because of the faint/asymmetric corner pixels.
Removing the antialising etc. it looks like the glpyh is rendered too small (8.5pt):
scrot_1711010521

I'll revert to the release version and report on the difference in a couple of hours.

**Comment by luebking** [comment]
The release version exhibits the exact same problem in alacritty (plus the subpixel line) - so the situation hasn't gotten any worse wrt the antialiasing but better wrt subpixel rendering in that case.

Looking close at the triangular glpyh
scrot_1711043465
it probably has the exact same "problem" but going from the square into the 45° angle just doesn't look off from a distance (the left-outmost pixels should probably be interpolated but are in fact the background of the adjacent block.
Blue and pink for better visualization of what actually happens in alacritty (MR font):
scrot_1711044081

**Comment by Finii** [comment]
I guess we now look into specific `alacritty` problems.
This looks like it assumes the 'cell' height to be different from what the font wants.

Some problems are not solvable with some clients. For example we have a bit of 'overhang' into previous/next cell to paint over the faint line. But some clients shrink glyphs that are bigger than the expected cell size to exactly fit, and then we end up with a not-tall-enough glyph that still has a faint line :grimacing: 

These clients that handle the font rendering all on their own, to be faster or better than the system... Sigh.
I would test this with `Writer` and some dumb (font rendering wise) client like `tilix`. If it works there the problem is the idiosyncrasies of that particular client.

A good way to check for cell size problems are Powerline glyphs `E0B8` and `E0BA`:

image

**Comment by Finii** [comment]
> Blue and pink for better visualization

I like that!

**Comment by luebking** [comment]
I mostly wanted to make sure that there're no regressions w/ other (popular) TEs (where afaict there are not)

I also was gonna say that xterm/st have the same problem, but actually they don't - the moment you configure their fonts using the pixelsize freetype syntax, they behave exactly like urxvt.
It's mostly about giving the dpi calculations free reign that the results become hard to control but usually "wrong".

----
Wrt alacritty, I could get better results when running "alacritty -v" and fiddling w/ the size variable until I had odd font and cell heights (which however doesn't seem to be always possible, at least not with single-digit precision)

**Comment by Finii** [comment]
Testing this now (playing around), with this output:

`test.sh`

<details>

```sh
#!/usr/bin/env bash

COL_A='\033[48;5;51m\033[38;5;200m'
COL_B='\033[48;5;200m\033[38;5;51m'

COL_A='\033[48;5;238m\033[38;5;253m'
COL_B='\033[48;5;253m\033[38;5;238m'

printf "\
${COL_A}                                                         \n\
Under \ue0b2${COL_B} lap ${COL_A}\ue0b0 and \ue0b6${COL_B} top ${COL_A}\ue0b4 mop\n\
${COL_A}Over \ue0b2${COL_B} lap ${COL_A}\ue0b0 and \ue0b6${COL_B} top ${COL_A}\ue0b4 mop trop lop klop\n\
                         \n"
```

</details>

which looks like this

image

image

image

image

with this settings (for the top image, afterwards is greyscale and none)

image

And the font is Delugia v2111.01.2, sorry ;-)
Terminal `tilix`.

---

Now taking the colors from this
* https://github.com/ryanoasis/nerd-fonts/issues/156

> The vertical centering of the glyph is actually an antialiasing problem, only.

Hmm, that is not the case for me :thinking: 
here see the output for antialiasing = LCD and varying hinting:

image

_antialiasing LCD, hinting none_

image

_antialiasing LCD, hinting slight_

image

_antialiasing LCD, hinting full_

Hinting moves (especially) the horizontal 'borders' of letters to pixel borders, to make them more sharp.
Obviously it moves the border of the powerline glyph down instead of up to be exactly on a pixel border (in my example).
If it moves up or down depends on font size and so on. Sometimes it is the lower edge that is moved.

Kooha-2024-03-29-08-37-23

Anyhow, I do not see the vertical antialiasing line; I know `tilix` is much ~~less picky~~ more forgiving than these super-terminal-emulators that do the font rendering on their own :roll_eyes: , but I know I can see the line on my other machine with the same setup, investigating.

---

AHHH, I know why I prefer Delugia over Caskaydia Cove :-D
That already has the vertical line fixed since ancient times

image

* https://github.com/adam7/delugia-code/pull/34

_**tbc**_

Finii:
This is `Caskaydia Cove Nerd Font Mono v2.1.0` in `tilix`, no hinting, RGB antialiasing

image

This is `Caskaydia Cove Nerd Font Mono v3.1.1` in `tilix`, no hinting, RGB antialiasing

image

This is `Caskaydia Cove Nerd Font Mono v3.2.0RC` in `tilix`, no hinting, RGB antialiasing

image

This looks clean. I always imagine faint lines, but that seems an optical illusion, even more magnification (click to enlarge)

image

I believe this PR successfully does what it sets out to do.

Finii:
@allcontributors please add @luebking for bug and review

Finii:
Thanks for all the input! I am very grateful for this / the discussions! :green_heart:

luebking:
Many thanks for fixing this, wrt.
> The vertical centering of the glyph is actually an antialiasing problem, only.

I meant to say that the visual perception of the shift or indention is caused by the antialiasing effect, ie. it's not really that the glyph is misplaced but the illusion is caused by the uneven density applied by the antialiasing.
But as mentioned before, that's entirely not related to this change and strongly related to rasterizer and font settings and only became more apparent once the subpixel line was gone.

tilix is btw. just a wrapper around VTE3, so all terminal emulators wrapping that should get the same behavior, see eg. https://archlinux.org/packages/extra/x86_64/vte3/

PR #1557, opened 2024-03-25
cheat-sheet: Add autofocus on load to search input text

guillaumeboehm:
#### Description

Autofocusing the cheat sheet input text on page load would improve some workflows and I doubt would break other workflows.
> I'm not 100% sure I understood the framework, but hopefully this works.

#### Requirements / Checklist

Adds the `autofocus` attribute to the search input field.

See if the search input field is focused automatically on page load

### DIFF
--- a/bin/scripts/data/cheatsheet-head.txt
+++ b/bin/scripts/data/cheatsheet-head.txt
@@ -10,7 +10,7 @@ style: container
 
 <h1 class="center">Cheat Sheet</h1>
 
-<input type="text" id="glyphSearch" placeholder="Search for glyphs/icons here..." title="Type in a glyph name or hex codepoint" class="nerd-font-cheat-sheet-search nerd-font-input tertiary sanity-test" />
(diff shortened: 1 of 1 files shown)


Finii:
Thanks for the PR.

Usually I hate focus stealing stuff, but here it is ok and helpful I guess.
Also considered concerns as stated here: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

I played around with it a bit and could not find and false trigger or problem.

Great! Thanks again :-)

Finii:
@allcontributors please add @guillaumeboehm for code

guillaumeboehm:
> Thanks for the PR.
> 
> Usually I hate focus stealing stuff, but here it is ok and helpful I guess. Also considered concerns as stated here: [developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus)
> 
> I played around with it a bit and could not find and false trigger or problem.
> 
> Great! Thanks again :-)

Oh yeah thanks for the link, good little read. I gotta say I didn't think about accessibility ^^ Anyway thanks for the approval !

PR #1560, opened 2024-03-26
Update readme_fr.md

cricrio:
Fix error: Error: caskroom/fonts was moved. Tap homebrew/cask-fonts instead.

#### Description

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/generate-casks.sh
+++ b/bin/scripts/generate-casks.sh
@@ -1,9 +1,9 @@
 #!/usr/bin/env bash
 # Nerd Fonts Version: 3.1.1
-# Script Version: 2.2.2
+# Script Version: 2.2.3
 #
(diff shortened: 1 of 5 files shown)


Finii:
Thanks for finding and correcting this!

Finii:
There is the general problem of outdated translated readmes.

I can not translate the readme into all these languages. Maybe it is better to not translate the readme _at all_ but let people use an online translator that became widespread in recent years. Firefox pesters me always with some 'do you want to translate this page'. And/or at least we should maybe have a very prominent statement that translated readmes can/will be out of sync.

As you obviously use the translated readme, what do you think about the situation here?
I do not even have a clue how/when the translated readmes will be shown, is there some automatism at work?
Do you have any experience how other projects do this?
Any input welcome :-)

Merci beaucoup!

PR #1564, opened 2024-03-28
font-patcher: Prevent excessively tall icons in mono fonts

Finii:
**[why]**
Very slender and tall icons will be scaled huge compared to their fatter neighbors. This happens for example to the chess pieces (`E25F` ff) or text style icons like italic (`EB0D`).

**[how]**
When a monospaced font is created use a slightly less tall cell height for all icons (except icons that have to match the actual line height).

This can be prevented by the new scale control character `^` that symbolizes maximized vertical size (height) within the real cell.

The smaller cell (now called icon cell) is less tall than the line height, but higher than the cap height of the font; to prevent people complaining that the icons shrunk too much.
It is about 1/3 higher than the cap height compared to the line height.

This is NOT applied to the other Nerd Fonts fonts variants (i.e. `Regular` and `Propo`) because all icons are generally a bit bigger than the cap height (no scaling down to fit into one advance width) and as the scaling is far less pronounced the slender-and-tall icons scale comparably and do not look out off place there.

Suggested-by: Aaron Bell @aaronbell

#### Requirements / Checklist

Some examples, prior left, with this PR right; glyphs marked blue by Fontforge:

image

image

image


Finii:
Rebase on master, force push

PR #1580, opened 2024-04-01
Added another good font to patch (SF Mono)

Kirito139:
#### Description

Updated README to add the font "SF Mono" to the "Other good fonts to patch" list.

#### Requirements / Checklist


Finii:
Thanks for the suggestion.

Issue #1435, opened 2023-11-22
Update Ubuntu Nerd Font to use refreshed 2023 Ubuntu Font


olivercalder:
**Requirements**

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my request and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

**Is your feature request related to a problem? Please describe.**
The Ubuntu fonts, originally designed in 2010-2011, received a redesign and re-release in 2023, making them variable and including various other improvements and extensions to the old design. The current Ubuntu Nerd Fonts are still based on the old 2010 version.

**Describe the solution you'd like**
I would like the Ubuntu Nerd Fonts to be rebased on the new 2023 Ubuntu Fonts. This would only need to apply to the latin/cyrillic/greek character sets, not to the symbols which make up the bulk of the Nerd Font symbol set.

**Describe alternatives you've considered**
An alternative would be to continue to use the old 2010 Ubuntu Fonts as they exist now, but I believe this is undesirable, as I think it is preferable for the Ubuntu Nerd Font to be a drop-in replacement for the stock Ubuntu font, which is now substantially different. Ubuntu preserves the original Ubuntu Font as the `ubuntu-fonts-classic` package, with the new fonts in the `ubuntu-fonts` package --- a similar approach could be taken with the Ubuntu Nerd Fonts, for those who wish to continue using the old Ubuntu Fonts.

**Additional context**
The newest release of the Ubuntu font can be found at: https://github.com/canonical/Ubuntu-fonts/releases/tag/v.1001

Finii:
Thanks for suggesting.

We had this some month ago already, and the problem was that Nerd Fonts can only patch static fonts, and the new Ubuntu font is only available as variable font. This just does not work.
Maybe they in the meanwhile added static fonts to their set, I will check out, and also try to link the past discussion.

Finii:
<img width="1004" alt="Screenshot 2023-11-22 at 18 20 48" src="https://github.com/ryanoasis/nerd-fonts/assets/16012374/57d200d2-c0d0-4136-b2c4-5b96750e5bdf">

Ah, ok. Will definitely do that in the next release. Thanks again for reminding.

Finii:
[...]

PR #1583, opened 2024-04-02
Add Ubuntu Sans

Finii:
Prevent the "SansMono" becomes "SansM" rule for UbuntuSansMono.

Fixes: #1435

#### Requirements / Checklist


olivercalder:
Does this entail that Ubuntu Sans Mono will be available as part of the Ubuntu Sans nerd font?

Finii:
I thought it would be ok to have one release archive with `UbuntuSans` and `UbuntuSansMono` together. Is that not good?

Finii:
We can still change that, I just realized I need to create a new Font-Logos release first to pull that into Nerd Font :grimacing:

olivercalder:
That's great, I just wanted to make sure that it was the case that mono would be included as well (I'm not very familiar with font packaging). Thanks!

Issue #1578, opened 2024-03-31
Provide machine readable metadata for fonts


jcgruenhage:
**Requirements**

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my request and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

**Is your feature request related to a problem? Please describe.**

I'm currently working on packaging nerd-fonts for a Linux distribution. Right now, packaging nerd-fonts means I have to maintain a list of the fonts myself, check the licenses for the fonts from different places in your docs. Other distros are duplicating a lot of work here as well, and most of them could probably benefit from metadata being provided as well.

Examples of other Distros maintaining lists:
- https://github.com/NixOS/nixpkgs/blob/master/pkgs/data/fonts/nerdfonts/shas.nix
- https://github.com/void-linux/void-packages/blob/master/srcpkgs/nerd-fonts/template#L12-L21
- https://git.alpinelinux.org/aports/tree/community/nerd-fonts/APKBUILD

**Describe the solution you'd like**

It would be immensely helpful for me, and probably other distros as well, if there was a json/csv/whatever file attached as a release artifact, that contained few fields for each font. The fields I've gathered from looking at the other distros packaging nerd-fonts are:
- original name for the font
- nerd font name of the font
- archive name
- sha256sum of the release archive
- license, as an [SPDX License Expression](https://spdx.github.io/spdx-spec/v2.3/SPDX-license-expressions/)

**Describe alternatives you've considered**

A lot of manual labor downstream for each distro that packages nerd-fonts. That's what's happening at the moment.

**Additional context**

None

Finii:
Thanks for the suggestion. This sounds useful and not too hard :+1:

Finii:
Do you have an idea how to express a SIL with RFN as SPDX? Just `OFL-1.1-RFN` or has the reserved name to be put somewhere? :thinking:

Finii:
SO. Needed to finish the release first ;-)

I believe this is a very good idea. We here also 'pack' the Nerd Font release, as Cask, and for that
* all the archives are fetched (there is a helper script [here](https://github.com/ryanoasis/nerd-fonts/blob/master/bin/scripts/fetch-archives.sh))
[...]

PR #1599, opened 2024-04-09
Improve licensing info

Finii:
#### Description

Add SPDX license info
    
[why]
It can be hard to find out which license a particular font is using.
Therefore we add a new field to the fonts.json with an SPDX license
identifier (if possible).
    
[how]
For fonts with a license that has no SPDX identifier we use a free
`LicenseRef-*` identifier, but that reference is not solved in the
fonts.json file.
    
[note]
Also correct some licensing info given in the license audit.
    
Related: #1578
    
Suggested-by:  Jan Christian Gruenhage <jan.christian@gruenhage.xyz>

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/lib/fonts.json
+++ b/bin/scripts/lib/fonts.json
@@ -2,6 +2,7 @@
   "fonts": [
     {
       "unpatchedName": "0xProto",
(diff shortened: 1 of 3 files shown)


Finii:
Fix typo, force push.

PR #1604, opened 2024-04-10
Custom: Added meson icon

sp1ritCS:
#### Description

This patch adds the icon for the [meson build system](https://mesonbuild.com/) to the "custom" icon set.

#### Requirements / Checklist

- [ ] This is going to need "written permission" from @jpakkane, as per [Meson logo licensing](https://mesonbuild.com/legal.html#meson-logo-licensing):
> Meson's logo is (C) Jussi Pakkanen and used by the Meson project with
> specific permission. It is not licensed under the same terms as the
> rest of the project.
> If you are a third party and want to use the Meson logo, you must
> first obtain written permission from Jussi Pakkanen.

The icon file is extracted (and slightly resized) from [mesonbuild/meson:graphics/meson_logo.svg](https://github.com/mesonbuild/meson/blob/master/graphics/meson_logo.svg)


jpakkane:
According to the docs, the fonts are licensed under MIT et al. That is fundamentally incompatible with Meson's logo and if I were to give permission it would be equivalent to relicensing the logo under MIT. This is not acceptable. The logo is explicitly not freely licensed because we want and need to maintain control of it. So unfortunately I won't be able to give permission for this.

Finii:
But maybe (I did not check) we can add this log :clown_face: 

image

(Which I never use; but some colleagues seem to prefer it - albeit it often is not able to do the setup with the rules I use.)

Finii:
Hmm, iconify lists the meson icon as MIT, in error:

image

Which originates here https://github.com/vscode-icons/vscode-icons
They lapidarily say the brand icons retain their original license, without mentioning it individually, if I looked right.

Issue #1321, opened 2023-07-22
Can you add https://github.com/uiwjs/file-icons to nerd-fonts


cathaysia:
https://github.com/uiwjs/file-icons contains fonticons from vscode-meterial-icon, it's useful for terminal icon, and it's license is MIT. Can you consider add it to nerd-fonts?

mafzaldev:
Hi! Can I contribute to this issue under HacktoberFest?

Finii:
I am a bit underwhelmed by file-icons font file. They originally have color icons (`svg`) and just stuff them into a font file.
But fonts are just black-and-white, which does require some manual postprocessing sometimes:

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/150b2e48-f974-4782-8e63-c53a437e7546)

This mouse-head icon is mostly a filled surface (white) which of course will be filled with the font-color which is usually black. The eyes, the ears, all becomes black. If that all-black blob is intended, then they should remove all the inner contours? And if not one needs to adjust the paths such that the white area is **cut out** instead of just filled with font-color.

This is a typical example if a svg collection for mainly web use, and for good measure they just throw it into a font file (because 'we can'), but that borders on un-usable and noone cares.

Sorry for the rant, but I would have to fix all the icons manually myself.

Finii:
@mafzaldev I'm not sure we want this and if we do there are several not-so-obvious things to consider when adding glyph sets. So while it would be easy to 'just add' the set the future maintenance is a problem here. I guess it does not have any codepoint stability guarantees, which means we need to add that afterwards via build scripts or something.
Then there are problems with the icons themselves (see my comment above).
And additionally, at least some of these icons are in Nerd Fonts already; what will we do with the n-th instance of a GO logo?

So thanks for the offer, but this is maybe beyond Hacktoberfest.

cathaysia:
Ok, so I close it. :(

PR #1610, opened 2024-04-15
Improve bazel icon

zaucy:
#### Description

Improved bazel icon. The old icon didn't have the separation lines which makes it look bad when used in applications such as neovim or other TUI applications.

Bazel icon was modified by myself

| old | new |
| --- | --- |
| image | TODO  |


Finii:
Thanks for the improvement!

codepoint `e63a` aka `nf-seti-bazel`

Finii:
image

Turns out good, just one nitpick, the gap seems to be non-constant in width? A bit narrower at A than at B?

image

Finii:
I wonder if these two have the same 'problem'?

image

_Kotlin and Rubi_

image

image
_But then, 'rubi' is a bad search term_

Finii:
And firebase?

image


And we have some alternative kotlin icon in the repo, which would turn out the same because fonts just have color or transparent, so one needs to crate the gaps.

image

---

As usual, each bug is a rabbit hole.

zaucy:
> Turns out good, just one nitpick, the gap seems to be non-constant in width? A bit narrower at A than at B?

That's fair! It was eyeballed. I could clean it up a bit.

> I wonder if these two have the same 'problem'?

Yeah I think a few icons could benefit from some transparent gaps. They show up so much better that way IMO.

Finii:
Codepoint `E552`, related:

* https://github.com/ryanoasis/nerd-fonts/issues/1321#issuecomment-1764174337

Also remember having seen `E511` with eyes :thinking:

Finii:
### Kotlin

The striped logo is obsolete, with the one rouchgly K shaped block we have already the latest and correct one.
See assets in https://kotlinlang.org/docs/kotlin-brand-assets.html

### Firebase

They have a monochrome icon, see https://firebase.google.com/brand-guidelines/
Will change.

### Ruby

Does not have a canonical monochrome icon, https://www.ruby-lang.org/en/about/logo/

### Bazel

There is no monochrome icon, but from the design ideas presented here https://blog.bazel.build/2017/07/05/new-logo-and-homepage.html it seems that the proposed change by zaucy is 'correct', at least if one does not want just the outside lines of the cubes.

image

Finii:
Ruby monochrome ...

image

The left one would be my favorite, but the angles are all wrong :unamused: 
The one in the middle looks too coarse for my eyes. Hmm.

image

This looks nice ;-)

Finii:
image

Finii:
@allcontributors please add @zaucy for code

Issue #1609, opened 2024-04-13
CaskaydiaCove Nerd Font -- "Arrow" Ligature


jesusvelezx:
Hello! I want to report a problem with the **"Caskaydiacove Nerd Font"** ligatures, specifically in _regular_ and _semibold_ type. When creating an arrow with `-` and `>` = `->` signs it looks strange, as if they are not of the same thickness.

I attach some screenshots.

**IDE:** PyCharm Professional
**Font:** Caskaydiacove Nerd Font Mono -- Regular & SemiBold
**Reserved Font Name:** Cascadia Code
**Version:** 2111.01

**Psdt:** 
-- The first image is Regular type.
-- Second image is SemiBold type

![Issue2](https://github.com/ryanoasis/nerd-fonts/assets/79605135/7532a524-209f-4e73-ba19-f885804c4181)
![Issue1](https://github.com/ryanoasis/nerd-fonts/assets/79605135/3235c828-9340-48db-8b3a-e890cd19226e)
![image](https://github.com/ryanoasis/nerd-fonts/assets/79605135/7b7529dc-c29c-4837-a1cf-32ec4f0f1cd4)

**Thanks!!**

Finii:
Lig looks ok in `writer`

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/05a88758-2acb-4cf5-9ada-51cdc07c189d)

The ligature uses these glyphs:

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/1aef19ae-3caa-4743-800a-0efeb2522cea)

The point coordinates are exactly the same, but the hinting might be different.

Ah, yes, here

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/867f5662-124f-4d8f-9cb6-463ecbfb0b27)

Which OS are you on?

jesusvelezx:
> Lig se ve bien en`writer`
> 
[...]

Issue #1291, opened 2023-06-06
Caskaydia Cove introduces ligature issue not in Cascadia Code


Kissaki:
Testing in Windows Terminal - font size 12, Windows 11, NF v3.0.2

I would expect the ligature for `=>` to be the same and without visual artifacts.

Cascadia Code:

![image](https://github.com/ryanoasis/nerd-fonts/assets/93181/e6bd7aac-408f-4eae-90d6-230e0b44b136)

CaskaydiaCove Nerd Font Mono:

![image](https://github.com/ryanoasis/nerd-fonts/assets/93181/08f0c635-ff35-4b68-a9a0-ab77ad0ace68)
 

#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

Finii:
Oh, sorry to hear that. Lets find the reason.

Works in `Writer`

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/330ef965-1adb-4495-874d-e5b745511941)

phjordan:
I see the same, in Windows Terminal as well as in VS Code, Nerd Fonts 3.0.2

![image](https://github.com/ryanoasis/nerd-fonts/assets/27995479/e80e1ae6-1ad1-4ed6-8c3c-41de1c886627)

Finii:
Can not reproduce on Win10, VS Code 1.71.0, NF 3.0.2:

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/98fbcd87-23c0-4326-97ee-1d2ec72bb871)

My settings:

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/d4bf0148-8fa1-4c6d-b2e3-3ce1751f1eb5)

Finii:
The problem with `Windows Terminal` is that they bundle Cascadia Cove (VF) with it and one can not uninstall it, which I hate.

Installing that thing now, hold tight.

Finii:
@phjordan  Btw, please share a picture of your `config.json` and maybe a view of the font settings dialog with 'cas' as search term:

[...]

PR #1613, opened 2024-04-18
font-patcher: Allow to rehint some Cascadia glyphs

Finii:
**[why]**
Some Caskaydia Cove glyphs that are used in ligatures (to create endless arrows for example) show uneven line thickness on some platforms.

The reason is the not-matching hinting of glyphs that are places next to each other and so minuscule differences are quite visible.

Note that the 'original' hinting is generated by VTT on the static Cascadia Code instances, which upstream just have ttfautohint hints that are different from the hints in the variable fonts and people complained that the look is different.

**[how]**
Add a new field to the config.json file that holds regexes of glyph names for glyphs that should be re-hinted by fontforge on patching time.

In principle we could also implement that as an additional pre-step that needs to be manually done after running VTT on the static font files.

I'm not sure which is better.

Note that fontforge generates a lot of debug output because the hinting is not ideal - the global tables are kept and probably less compatible to fontforge's own hinting... But the results are good.

Fixes: #1291
Fixes: #1609

#### Requirements / Checklist


Finii:
image

Finii:
Compared previous and current passed options via this change

```diff
index c2535a1dc..4c71ad6e7 100755
--- a/font-patcher
+++ b/font-patcher
@@ -2082,6 +2082,11 @@ def main():
         logger.info("Can not write logfile, disabling")
     logger.debug("Naming mode %d", args.makegroups)
 
+    qqq = list(vars(args))
+    qqq.sort()
+    print("|||", [[k,vars(args)[k]] for k in qqq])
+    sys.exit(0)
+
     patcher = font_patcher(args)
 
     sourceFonts = []
```

Run via `gotta-patch-em` on all fonts and logging into one file each before and after the PR.

After some `sort` and `uniq -u` and so on it could be seen that the delivered options were exactly the same (well, except that some had a `configfile` set that previously hadn't. These are the fonts which got the parameters before via shell script and now via INI.

Issue #1611, opened 2024-04-17
U+276F prompt character (❯) got smaller between 3.1.1 and 3.2.1


gibfahn:
### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### Experienced behavior:

Prompt character is now small.

#### Expected behavior:

Prompt character is large and easy to spot in the terminal.

#### Example symbols:

- U+276F (❯) 

### 🔧 Your Setup

- Terminal: [kitty.app](https://sw.kovidgoyal.net/kitty/) 0.34.0, but it also reproduces with Terminal.app
   - Configuration (in case useful): [kitty.conf](https://github.com/gibfahn/dot/blob/16a690144d8c9457204441ec74e3142bf75ced86/dotfiles/.config/kitty/kitty.conf#L28)
- Font: `Hasklug Nerd Font Mono`, installed by brew from: [font-hasklug-nerd-font](https://github.com/Homebrew/homebrew-cask-fonts/blob/HEAD/Casks/font-hasklug-nerd-font.rb)
   - Brew [downloads from](https://github.com/Homebrew/homebrew-cask-fonts/blob/b415b1b34f996ced02785721a2fd30e5c5ea21e4/Casks/font-hasklug-nerd-font.rb#L5) https://github.com/ryanoasis/nerd-fonts/releases/download/v3.2.1/Hasklig.zip
- OS: latest macOS

#### ★ Screenshots

Screenshots taken at the same zoom in kitty.app on old and new font versions:

| 3.1.1 | 3.2.1 |
| --- | --- |
| ![3.1.1](https://github.com/ryanoasis/nerd-fonts/assets/15943089/3747a6df-1d62-40df-ac16-80388312c8e0) |  ![3.2.1](https://github.com/ryanoasis/nerd-fonts/assets/15943089/de1fd3c3-fc56-406b-99f6-6576f0c09985) |

Finii:
Sorry to hear of your size problem.

I can see it directly in the font :thinking: 

[...]

PR #1617, opened 2024-04-22
font-patcher: don't double-shrink heavy angle brackets

gibfahn:
Fixes: https://github.com/ryanoasis/nerd-fonts/issues/1611

Directly copied from https://github.com/ryanoasis/nerd-fonts/issues/1611#issuecomment-2061294859

#### Description

> Ah, the heavy brackets are now two times shrunken, there was an old rule that already sized them down relative to the full cell:
> 
> image
> 
> The rule needs to be changed to `^pa1!` ....

#### Requirements / Checklist

Restores heavy angle brackets to the size they were in 3.1.1 

Check size of heavy angle brackets before and after this change.

https://github.com/ryanoasis/nerd-fonts/issues/1611

 | 3.1.1 | 3.2.1 |
| --- | --- |
 | 	|  |

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -944,7 +944,7 @@ class font_patcher:
             0xf0de: {'align': 'c', 'valign': '', 'stretch': 'pa', 'params': {}}
         }
         SYM_ATTR_HEAVYBRACKETS = {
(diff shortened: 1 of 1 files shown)


Finii:
@allcontributors please add @gibfahn for code

Issue #1635, opened 2024-05-14
Homebrew/cask-fonts will soon be deprecated


p-linnane:
Hello 👋 . I'm a maintainer for the [Homebrew](https://brew.sh) project. I'm opening this issue to inform the nerd-fonts maintainers of the upcoming deprecation of Homebrew/cask-fonts. All fonts will be migrated to the main Homebrew/cask repo.

We will migrate all nerd-fonts to the main Homebrew/cask repo and introduce a redirect so existing users will seem seamlessly transitioned. I would like to work together to find a time the nerd-fonts release workflow can be updated, and migrate the fonts at the same time so there is no disconnect. Please let me know if there are any questions.

cc @Finii who I've worked with before on this.

Finii:
Thank you for the information! :+1:

I believe the change in our workflow will be easy. Was busy today and will be tomorrow, but I will come back to this on the weekend.

p-linnane:
Sounds good. I will keep an eye on this thread. Thank you!

p-linnane:
Closing as this migration has been finished for some time. Shout if you need anything @Finii!

PR #1637, opened 2024-05-16
CI: Adapt Homebrew Cask generation

Finii:
**[why]**
The Cask-fonts repo has been incorporated into the main Cask repo. On releases we would use an obsolete repository to create the Cask into.

The casks are now in a dedicated subdirectory (`Casks/font`) and under that we have one subdirectory for each font name starting character (to have a smaller number of files per directory I assume).

**[how]**
Change the repository path.

Put part of the new (longer) path, the fixed part, just in the CI process, while the generator script takes care about the one-directory-per-startchar substructure.

Fixes: #1635

#### Requirements / Checklist

Update the Cask generation  and workflow to the upstream (or is that downstream) architectural changes.

Let the workflow run and examine the resulting PR, then close the PR :-D
Well, we need some change in our repo that causes the casks to differ or the PR will be empty.


Finii:
Running the scripts locally looks good:

image

Finii:
Rebase on master, force push

Finii:
PR created: https://github.com/Homebrew/homebrew-cask/pull/192468

Unfortunately some glitches, so did some hand-adjustments. This needs to be addressed.

Issue #1647, opened 2024-06-01
Docker nerdfonts-patcher errors out on fontforge start


kjkent:
#### 🗹 Requirements

- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [X] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [X] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior:

Docker container nerdfonts/patcher errors out when run. I've attempted both compose and run syntax:
```yaml
 services:
   nerdfonts-patcher:
    command: |
      --mono
      --adjust-line-height
      --progressbars
      --complete
    container_name: nerdfonts-patcher
    image: nerdfonts/patcher
    #user: ${UID}:${GID}
    volumes:
    - ./nf-in:/in
    - ./nf-out:/out
```
or `docker compose run --rm -v ./nf-in:/in -v ./nf-out/out nerdfonts/patcher --mono --adjust-line-height --progressbars --complete`.

Both result in:
```shell
Running with options:
 --mono --adjust-line-height --progressbars --complete
Parallelism 0
fontforge -script /nerd/font-patcher -out /out --mono --adjust-line-height --progressbars --complete /in/berkeley-mono/OTF/BerkeleyMono-BoldItalic.otf
#### above line repeated for each font

Nerd Fonts Patcher v3.2.1 (4.14.3) (ff 20230101)
Done with Patch Sets, generating font...
Copyright (c) 2000-2023. See AUTHORS for Contributors.
 License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
 with many parts BSD <http://fontforge.org/license.html>. Please read LICENSE.
 Version: 20230101
 Based on sources from 2023-10-26 13:09 UTC-D.
Core python package 'pkg_resources' not found: Cannot discover plugins
Traceback (most recent call last):
  File "/nerd/font-patcher", line 2155, in <module>
    main()
  File "/nerd/font-patcher", line 2147, in main
    patcher.generate(sourceFonts)
  File "/nerd/font-patcher", line 415, in generate
    sourceFont = sourceFonts[0]
                 ~~~~~~~~~~~^^^
IndexError: list index out of range
##### above block repeated multiple times
```
[...]

PR #1648, opened 2024-06-02
font-patcher: Fix behavior when trying to patch non-font files

Finii:
**[why]**
When the file specified to be patched is not a font file the patcher run errors out with an out of index runtime error:

```
Traceback (most recent call last):
  File "/home/fini/extra/git/nerd-fonts/font-patcher", line 2155, in <module>
    main()
  File "/home/fini/extra/git/nerd-fonts/font-patcher", line 2147, in main
    patcher.generate(sourceFonts)
  File "/home/fini/extra/git/nerd-fonts/font-patcher", line 415, in generate
    sourceFont = sourceFonts[0]
                 ~~~~~~~~~~~^^^
IndexError: list index out of range
```

**[how]**
Do not assume that the specified file will be a font file but rather check if fontforge detects a font in the file and error out if there is no font found.

Fixes: #1647

Reported-by: Kristopher James Kent <kris@kjkent.dev>

#### Requirements / Checklist

Prevent a runtime error in certain situations and instead prints an understandable explanation why processing can not continue.

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "4.14.3"
+script_version = "4.14.4"
 
 version = "3.2.1"
 projectName = "Nerd Fonts"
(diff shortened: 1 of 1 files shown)


Finii:
Screenshot from 2024-06-02 08-21-38

Issue #1652, opened 2024-06-11
GohuFont question mark is upside-down


dsvoid:
#### 🗹 Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related and/or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

#### 🎯 Subject of the issue

Experienced behavior: GohuFont 14 Nerd Font Propo Regular has an upside-down question mark. This may be the case for the variants.

Expected behavior: Right-side up question mark.

Example symbols: "?"

#### 🔧 Your Setup

- _Which font are you using (e.g. `Anonymice Powerline Nerd Font Complete.ttf`)?_
   - GohuFont14NerdFontPropo-Regular.ttf
- _Where did you get the file from (download link, self patched, source downloaded from link...)_
   - https://github.com/ryanoasis/nerd-fonts/releases/download/v3.2.1/Gohu.zip
- _Which terminal emulator are you using (e.g. `iterm2`, `urxvt`, `gnome`, `konsole`)?_
   - Alacritty. Also replicated in dmenu and other tools.
- _Are you using OS X, Linux or Windows? And which specific version or distribution?_
   - Asahi Fedora Remix 40. Appears to be in the ttf itself, however, when opened with fontforge.

#### ★ Screenshots (Optional)

![Untitled](https://github.com/ryanoasis/nerd-fonts/assets/7375973/84363f7b-af88-4223-a714-c13ebefbafe0)

Finii:
![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/31ba7867-7cda-46d6-9d9f-934d2b91d6c9)

Finii:
@allcontributors please add @dsvoid for bug

kierun:
[⸮ is the irony mark, isn't it?](https://en.wikipedia.org/wiki/Irony_punctuation)

Finii:
Yes, it is the spanish start-a-question-symbol ... like _**¿Hablas español?**_

In the 11 px variant all is well:

![image](https://github.com/ryanoasis/nerd-fonts/assets/16012374/6175045a-cadc-43e6-a2a2-22489360d48f)

[...]

PR #1654, opened 2024-06-11
Fix upstream Gohufont bug

Finii:
#### Description

Separate the question mark and the inverted question mark in Gohufont 14.
They have the same glyph name in that font file and the wrong one is usually used.
The other becomes inaccessible.

See report in https://github.com/koemaeda/gohufont-ttf/issues/3

Here we fix the source font file directly.

#### Requirements / Checklist

Fixes: https://github.com/ryanoasis/nerd-fonts/issues/1652


Finii:
Need to check the bitmaps before merging.

Finii:
Bitmaps are ok after PR. 

image

In `fontforge`, View, enable bitmap at bottom

image

PR #1664, opened 2024-06-23
style: Sort variable name ascending in shell script library

mdsanima:
Optimized the definition of icon variables in `i_cod.sh` by sorting the name ascending for better responsiveness. This improves code maintainability for better visual representation and allowing for easy localized names for a given icon.

This change only sorts names alphabetically, variable names were not changed, they remained as they were at the beginning.

#### Description

#### Requirements / Checklist


Finii:
Thanks for this (still in progress) PR!

I'm not sure I understand the reason for the alphabetic sorting ("more responsiveness")?

At the moment they are sorted by codepoint.
And they are usually autogenerated, so changing the current files does not change anything if the generation is re-triggered.

Usually they are used via the `i_all.sh` script, that does just set a lot of shell variables (where the order does not matter), and from there the shell is used to access the variables (i.e. glyph names) and values (i.e. codepoints).

These files are not for manual ... interaction?

:thinking: 

Fini

mdsanima:
Oh no :( I sorted the variable names because it's easier for me to find specific icons just by their name and it makes my work much easier if I need a given icon.

Finii:
See git log ...

```console
$ git log i_cod.sh
commit 2fa45981ee4b46eded3ceee3e4a5100fee442786
Author: Fini Jastrow <ulf.fini.jastrow@desy.de>
Date:   Sun Mar 17 14:21:15 2024 +0100

    generate-glyph-info: Fix double entries
    
    [why]
    Two glyphs may not have the same name (normally).
    Our glyphnames.json will break.
    
    [how]
    Do not add two entries with the same name but rather report the
    codepoints in the bottom.
    
commit 6932ff2d8fe5652301330de76cf260bfd52f2452
Author: Fini Jastrow <ulf.fini.jastrow@desy.de>
Date:   Sun Mar 17 13:47:24 2024 +0100

    Codicons: Update i_cod.sh
    
    With tool, call:
    
    fontforge generate-glyph-info-from-set.py --font ../../src/glyphs/codicons/codicon.ttf --start ea60 --end ec1e --offset 0 --prefix cod > i_cod.sh_part
    
```

> it's easier for me to find specific icons just by their name

Maybe you can use the https://github.com/ryanoasis/nerd-fonts/blob/master/glyphnames.json file, which allows very easy access to the names via `jq` sorted in any order you might need?
And/or use the cheat-sheet?

Just suggestions to solve your (unknown) problem; because I guess what you want to do is not what the `i_*` files are intended for.

mdsanima:
I've sorted almost all the variables in the `lib` folder, where the files are located.

If the variables are autogenerated, I can modify the script to generate files sorted as I did. I think this might be useful. For me, alphabetically sorted names are very helpful.

mdsanima:
> 
> Maybe you can use the https://github.com/ryanoasis/nerd-fonts/blob/master/glyphnames.json file, which allows very easy access to the names via `jq` sorted in any order you might need? And/or use the cheat-sheet?
> 

Yes, that's actually what I need. So I guess my sorting is unnecessary. Too bad I didn't find this file earlier.

mdsanima:
But if it was useful, I can change the main file that processes them so that they are sorted.

Finii:
I'm not sure if it is useful. They **are** already sorted right now, but by codepoint instead of name. Which is a more technical aspect of the glyphs, but then, the `i_*.sh` 'database' is kind of the technical side of keeping all the glyphs organized.

I'm not against sorting it, but the question is ... what is better.

Usually if a new icon is added to some font, it is added in the end and the codepoint-sorted files make it very easy for maintainers to just add them in the bottom of the `i_*` file. What is also lost in your versions is the number of aliases as they are not very apparent anymore and one would need to search for them explicitly with some not-simple code.

I guess the aliases and keeping them organized is one fact that is against alpha sorting :thinking:

mdsanima:
Yes, that's a mistake. The option you suggested @Finii with `jq` is the best for now.

PR #1667, opened 2024-06-25
feat: overhaul issue templates

gu5e:
#### Description

This PR overhauled the issue templates. It changed them to issue template forms and fixed mistakes and consistency.

#### Requirements / Checklist

This PR changes the format of the issue templates and fixes mistakes.

It can be manually tested by clicking the New Issue button in the Issues tab and selecting one of the categories

bug_report_template
feature_request_template


Finii (commented):
Thanks for the nice PR!

This looks very good! I have two comments, though:

1. In the bug report terminal emulator field the text gets truncated (in your screenshots as well as in my tests); of course this depends on the actually used fonts etc, but maybe we should make it a bit shorter.
   I would suggest using `kitty` instead of `gnome-terminal`, which probably fits and according to the issue frequency is more used.
2. The checkboxes are now all mandatory, which means you can not press the submit button until all checkboxes are ticked.
  I guess when people are honest and did not do all the stuff, and then want to submit, they just tick the other ones, because, well, it is required. Maybe it is better to let these non-required to get honest answers to the questions. Of course they should all be checked but just forcing people to check them can have diametrical effects I fear?
    What do you think?

Finii:
@allcontributors please add @gu5e for doc

Finii:
I would also add these comments somehow, because frequently people do not add screenshots at all (which makes reproduction very hard) or snip just a very small fraction of the screen...

```
<!-- Provide screenshots where appropriate, they help tremendously -->
<!-- Try to include the title of terminal windows and possibly settings dialogues -->
```

Previously this section has been called 'Screenshots' explicitely, for the same reasons:

image

Finii:
Example of people ticking all tickboxes, although it is not true...

image

(There was no previous discussion, it is ticked nonetheless :grimacing: )

gu5e:
I have tried to address most of the things you pointed out @Finii (sorry for the accidental ticking btw, ill be sure to do it next time)

Finii:
> sorry for the accidental ticking btw, ill be sure to do it next time

:-D no problem at all, ppl do it all the time

Thanks for the PR, I will fix one small thing an then merge.
:+1:

Finii:
Thank you! :green_heart:

PR #782, opened 2022-02-08
Update issue templates

Finii:
A lot of issues do not have screenshots and which glyphs are the problem has to be asked and or guessed.

Also change to newer github issue template method.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/master/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?

Change the Issue Template

#### How should this be manually tested?

Can be tested interactively in my fork, installed it there in master [`Finii:master`](https://github.com/Finii/nerd-fonts), so it is in effect.

#### Any background context you can provide?

Frustrated by people that think I know all the icons by heard ;-)

#### What are the relevant tickets (if any)?

Replacement for #763

#### Screenshots (if appropriate or helpful)

After **_Create New Issue_**:

image

image

PR #1671, opened 2024-06-27
fix: rename issue form files for consistency

gu5e:
#### Description

This PR renames the issue form files to use `-` instead of `_`. Sorry for not remembering this in #1667

#### Requirements / Checklist

It renames the issue form files to match GitHub **issue form yaml file** naming conventions.

If you have functional eyes, you can look at the filenames of the issue forms.

Nope.


Finii:
> to match GitHub naming conventions

The only 'convention' I could find was this:

image

And `ISSUE_TEMPLATE/FORM-NAME.yml` (with underscore in the path and dash in the file name) is far from consistent in my eyes?

image

Maybe you have a link to the naming convention?

gu5e:
Unfortunately i don't, but there's not documentation referencing that you should name the file with underscores.

gu5e:
And you're also naming your github actions with dashes

Finii:
Here the history of how we came to the current names (via #782)

image

and they still say underscore is the thing ;-)

image

gu5e:
But you are already using dashes for the workflow file names

Finii:
Some examples for repos I work with

image

image

And this

image

I just checked these 3, so I guess underscore is ok ;)

Finii:
> But you are already using dashes for the workflow file names

I have no clue why, but that tends to be what people do.
The same 3 repos:

image


image
_Azure of course for Microsoft_

image

gu5e:
Okay, i actually didn't know that, gonna close it. Thanks!

Finii:
Thank you for caring for the details like filenames :+1:

PR #1694, opened 2024-08-27
Add CLI cheatsheet script

hasecilu:
#### Description

_CLI cheatsheet._

#### Requirements / Checklist

Adds a script that let users to search for an icon using its name using the NF library scripts.
```shell
$ ./cheatsheet divider
       e0b0    i_pl_left_hard_divider
       e0b1    i_pl_left_soft_divider
       e0b2    i_pl_right_hard_divider
       e0b3    i_pl_right_soft_divider
       e0d7    i_ple_left_hard_divider_inverse
       e0d6    i_ple_right_hard_divider_inverse
```
There are times when I want to look for a glyph but using the online cheatsheet tool takes "a lot of time". The tool is not slow I mean the whole workflow.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,3 @@
+[*.sh]
+indent_style = space
+indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,0 @@
-[*.sh]
-indent_style = space
-indent_size = 2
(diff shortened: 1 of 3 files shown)


hasecilu:
To fix the shellcheck fail we can update the hook to use `-x` flag to let it access external files `shellcheck -x cheatsheet.sh` or add a config file. [docs](https://www.shellcheck.net/wiki/Directive#external-sources)

Finii:
Thanks! Excellent!

Finii:
Ah, added also these commits:

66ae03de2 `shellcheck.sh: Unify variable style`
d294607b1 `cheatsheet.sh: Make CWD independent`

Finii:
This (again) reminds me that `glyphnames` needs a mechanism for aliases, so that in some far future we can drop the `i_` files and instead work with the json 'database'.

Finii:
What is `redundant` ?! :thinking: 

image

Issue #1695, opened 2024-08-28
Inverse triangle position


Finii:
This is a bug :fearful: 

![image](https://github.com/user-attachments/assets/85a2cc35-7ff0-4f21-a337-8843748eb125)

_Originally posted by @Finii in https://github.com/ryanoasis/nerd-fonts/issues/1694#issuecomment-2314576050_

Finii:
It's ok in the 'normal' patched fonts, all 3 variants :sweat_smile: 

![image](https://github.com/user-attachments/assets/66d81a43-3cdd-49ef-9013-6adcf3545b82)

It happens only in the `SymbolsNerdFont-Regular` font (not the `Mono`), but seems to be with all right aligned glyphs?

![image](https://github.com/user-attachments/assets/2e28e915-a813-4f7d-9175-fd9859d2ad0d)

PR #1727, opened 2024-10-30
Fix Powerline glyphs position and size in some cases

Finii:
#### Description

Correct right aligned glyphs in proportional fonts (i.e. `--variable`).

Only very few icons are right aligned, all Powerline glyphs.

#### Requirements / Checklist

* #1695

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "4.14.7"
+script_version = "4.14.9"
 
 version = "3.2.1"
 projectName = "Nerd Fonts"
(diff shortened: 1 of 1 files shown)


Finii:
Comparing `SymbolsOnly` font for `4.14.7` (before this PR) and `4.14.8` (with this PR):

### Nerd Font Mono

NO CHANGE

image

### Nerd Font (which is secretly Propo)

image

Examples (left is 4.14.7, right 4.14.8)

image
image
image
image
image
image

Not exactly what I wanted I guess.

Finii:
Now only the right aligned icons are different, and the original / proportional width is kept.

image

_Left: pre PR, right after PR, changed glyphs marged with blue tag_

I'm much more comfortable with this change,which is minimal. Making the Powerline glyphs monospaced, albeight **_I_** believe that is what they ought to be, that might break other people's setup, and noone ever complained.

Finii:
Checked with SymbolsOnly and `CaskaydiaCoveNerdFontPropo-Regular.ttf`

Issue #1345, opened 2023-09-03
Add Fira Code's progress bars and spinner. 


oshaboy:
**Requirements**

- [*] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my request and found nothing related and/or helpful
- [*] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [*] I have searched the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

**Is your feature request related to a problem? Please describe.**
It's not a problem it's a request for more symbols from Fira Code.

**Describe the solution you'd like**
I would like the Fira Code progress bar and spinner to be added to the font. 

**Describe alternatives you've considered**
The only fonts I know that support Fira Code's progress bar and spinner are Fira Code and Iosevka. 

**Additional context**
![](https://raw.githubusercontent.com/tonsky/FiraCode/master/extras/progress.png)

Fira Code added glyphs for a progress bar and spinner in the PUA. Not much software support for them is out there. It feels like they should be glyphs in nerd fonts but they aren't.

oshaboy:
Shouldn't this be added to #1095. I didn't know about that thread when I made this issue

mikesmithgh:
Hey 👋, this is really nice. I am using Fira Code Nerd Font `v3.1.0` and can see the spinners.

I modified `test-fonts.sh` locally for expirementing.

```sh
	echo "Nerd Fonts - Fira"
	print-unicode-ranges ee00 ee0b ee10 ee15
	echo
	echo
```
![fira-nerd-fonts](https://github.com/ryanoasis/nerd-fonts/assets/10135646/b9cae492-4462-40f0-9be8-11c4e82a935e)

I have a question around what this ticket actually means for implementation.

1. Since I can see the symbols, does this mean that this is already implemented in `v3.1.0` ?
or
2. Will this add these symbols be patched to all Nerd Fonts and maybe use another private-use unicode area?

Finii:
Type `fc-list :charst=ee00` and see where the icons come from. Nerd Font does not patch in anything at `EE00`, see https://github.com/ryanoasis/nerd-fonts/wiki/Glyph-Sets-and-Code-Points#overview

The spinners are originally in Fira (so if you use Fira or a patched version of it) they are there.
This Issue aims to steam them from Fira and put them in all fonts.

oshaboy:
[...]

PR #1733, opened 2024-11-14
font-patcher: Add progress indicators

Finii:
Fixes: #1345

#### Description

Add these 12 glyphs that can be used as progress indicators and originate from Fira Code:

We specifically left these codepoints un-occupied for these glyphs in previous PRs.

It turned out that the scaling and shifting algorithm has some ... peculiareties, and these glyphs need to be properly aligned and have overlap for gapless display (the first 6 glyphs), so that needs to be fixed first :grimacing: 

That in turn showed even more bugs ;-)
It is always an endless pit. Lets see if we can straiten out all the problems.

#### Requirements / Checklist

image


Finii:
Something is wrong with `--variable` when the glyph is `align = c` :thinking: 

Compare handling of `EE01` and `2630`

Finii:
:grimacing: 

image

Finii:
There is obviously a problem with center aligned glyphs... But that is already there before this MR :-(

Two notabke commits:

acd49cc33 `font-patcher: Fix overlap for align c and r`
5b36e8ec1 `font-patcher: Keep overlap in proportional font`

```
    font-patcher: Fix overlap for align c and r
    
    [why]
    The box drawing glyphs (center aligned) and some Powerline glyphs (that
    are right aligned and have a limited xy-ratio) are positioned wrong.
    
    Affected are all box drawing glyphs (e.g. 2548) and some Powerline
    glyphs (i.e. E0B2, E0B6, E0C5, E0C7, and E0D4).
    
    [how]
    The box drawing glyphs are center aligned and have overlap. The code
    does not correct the overlap (left bearing) but uses the default case of
    'make the left bearing zero'. The code does just check left aligned
    glyphs and not center aligned ones.
    Add the correct overlap for center aligned glyphs (i.e. half the
    overlap).
    
    The Powerline glyphs are right aligned. Usually that works, because the
    glyphs are created with the right size, so that no additional
    manipulation is needed.
    But if the glyph has a ratio limit the resulting size will be different.
    We could in fact fix the size code, somehow, but that is rather
    complicated, formula-wise. Instead we just scale these glyphs (which are
    the 5 listed above) and shift them to the right position such that the
    correct overlap results.
```
```
    font-patcher: Keep overlap in proportional font
    
    [why]
    The previous commit is somewhat incomplete in some cases and plain wrong
    in others (with proportional fonts). Examples:
    
    The Heard 0x2665 gets a positive right side bearing, which is
    unexpected. The commits prevents negative right side bearings but not
    positive ones.
    
    The glyphs with overlap (which are the Powerline ones) like 0xE0B0 and
    0xE0B2 end up in wrong sizes.
    
    This can especially be seen with the Symbols-Only (non-Mono) font,
    because that is (secretly) a 'Nerd Font Propo' (--variable-width-glyphs)
    font.
    
    [how]
    This is kind of a design choice: As with the other patched font variants
    we ignore existing borders (positive bearings) around the glyph. The
    previous commit tried to keep them, which seems to be impossible and
    is inconsistent). Also negative bearings would be ignored (but there are
    none).
    
    The only place where bearings come into play is now when we have
    overlap. All non-overlap glyphs render without any bearing.
    
    If we have overlap we need to
    a) reduce the width by the overlap
    b) introduce a negative bearing on the appropriate side
    
    First we remove any left side bearing by transforming the glyph to the
    side, such that the bearing becomes zero.
    
    For left-side overlap we additionally transform the glyph by the overlap
    amount to the left (as usual). This creates the neg. left bearing.
    
    For right-side overlap we keep the left bearing to be zero.
    
    After correcting the left-side bearing (by transforming) we set the
    corrected width. That is the width subtracted by the overlap.
    In the left-aligned case this makes the right-side bearing zero.
    In the right-aligned case this results in a negative right-side bearing.
    
    Note how fontforge handles size and bearing changes:
      Fontforge handles the width change like this:
      - Keep existing left_side_bearing
      - Set width
      - Calculate and set new right_side_bearing
```

Finii:
image

Not centered in SymbolsOnlyNF

Also here in Propo

image

Finii:
```
DEBUG: Advance of 0xF0A6 in BBDIMS set 0xF0A4 - 0xF0A7 is 1417, expecting 1890
DEBUG: Advance of 0xF155 in BBDIMS set 0xF153 - 0xF15A is 1175, expecting 1181
DEBUG: Advance of 0xF177 in BBDIMS set 0xF175 - 0xF178 is 1890, expecting 1417
DEBUG: Advance of 0xF05C in BBDIMS set 0xF059 - 0xF0B1 is 1537, expecting 1533
```

Finii:
So lets crate a list of fonts that need to be tested
* CascadiaCode NF NFM NFP
* SymbolsOnly NF(P) NFM
* BigBlueTermPlus NF NFM NFP

Preparing fonts prior to this PR...

Finii:
## Cascadia Code NF

Minimal differences, some cell height problems fixed.

image

Finii:
Checked all the 8 check fonts mentioned above.

Last issue is this :thinking: 

image

Finii:
## Seems we are ready

image

Finii:
fixed via hotfix e6426bbc602ff2d47e1109623a1c1b039ab531e8

PR #1753, opened 2024-11-21
Overhaul README

realguse:
### Description

This PR will hopefully do a good job of cleaning up the enormous and kind of confusing README. I have plans to short it down and make it look prettier. This will most likely take some time to complete and the README may be broken until completed.

### Improvements for README

- [x] Introduction
- [x] Description
- [x] Diagram (may be removed)
- [x] Notice
- [ ] Table of Contents (will be TEMPORARILY removed)
- [x] Tl;DR (removed)
- [ ] Features (may be removed)
- [ ] Glyph Sets (may be removed)
- [ ] Patched Fonts
- [ ] Installation (will be heavily shortened or half moved)
- [ ] Font Patcher
- [ ] Gotta Patch 'em All Font Patcher! (will be merged into above)
- [ ] Contributing (may be removed, already in contributing.md)
- [ ] Common Problems (dont know)
- [ ] Unstable File Paths (may be removed)
- [ ] Other Good Fonts to Patch (dont know)
- [ ] Project Motivation (only in wiki)
- [x] Changelog (may be removed, already in changelog.md)
- [x] License (may be removed, already in LICENSE)

### Requirements


Finii:
Thanks for starting work on the readme. That is indeed a mess and needs cleanup, but doing it carefully and without breaking expectations by long time "old" users is not so easy. I also believe it should be much much smaller and rather link to subtopics - but then searchability can be a problem. Anyhow, you are right, this needs to be worked on.
I started it ... twice iirc, but the final result was also not too nice or it was just too different 😬 Some information in there is too prominent imho and the impression it transports can be wrong. Yes, difficult.

This is still a draft, but let me quickly add two thoughts.

1. Why are the links replaced by raw url strings. This does not improve the readability but degrades it. We had all links in the bottom of the file before. And then, while the symbolic links are replaced by concrete urls, the link definition is not removed? 🤔 
2. I would not promote the translates readmes at all. They are a maintenance nightmare, and I would rather remove them because they are so utterly outdated.

realguse:
Hi, thanks for the interest. As I said, this will probably take around a month to complete, I'm currently in school so not much free time here, but I swear to work on it consistently.

Regarding your first issue, I unfortunately don't quite understand. Is it the relative links that I should add back?

On the translations, do you want me to remove them, most developers use english anyways.

Thanks again

realguse:
The readme will probably not look great or be broken until its finished

Finii:
I fear this is too big or radical; it also changes things that are not on the agenda as given above (i.e. out of scope).

For example take commit 360dd8882c9a7c496733f221a763ee8e9acb1528

```diff
-| [CaskaydiaCove Nerd Font](patched-fonts/)              | [Cascadia Code][f-cascadia]            | 2111.01    | Yes      |
+| [CascadiaCode Nerd Font](patched-fonts/CascadiaCode%20NF)                   | [Cascadia Code][f-cascadia]            | 2111.01    |
```

What?!? Why is the name changed from `Caskaydia` to `Cascadia`? This makes no sense.

Or 9d7c43fca02c8036c9cacc558fa308ea573d7f37 `feat!: prepend NF to every font dir`, this is a no-go. Why; I guess something not understood?

I also think this is too big in its completeness, even IF you would not expand the changes to unrelated parts (i.e. not `/readme.md`).

Finii:
Or take this plan item

Screenshot 2024-11-26 at 22 26 13

It seems you did no research on why the items are there, this specifically has been JUST added (git blame shown):

Screenshot 2024-11-26 at 22 26 30

realguse:
Ive just always thought that this felt so "budget" if I could say so; the long readme etc. The licenses are resolved when prepending NF, and it still links to the original project

PR #1765, opened 2024-12-21
docs: improve contributors badge

realguse:
#### Description

Made the all-contributors badge (at the top here https://github.com/ryanoasis/nerd-fonts/blob/master/CONTRIBUTORS.md) update live and changed the color (`orange` → `#ee8449`).

#### Requirements / Checklist

Improves the all-contributors badge in CONTRIBUTORS.md

All in accordance with https://allcontributors.org/docs/en/bot/faq and https://shields.io/badges/git-hub-contributors-from-allcontributors-org

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -1,5 +1,6 @@
 
-[![All Contributors](https://img.shields.io/badge/all_contributors-118-orange.svg?style=flat-square)](#contributors)
+[![All Contributors](https://img.shields.io/github/all-contributors/ryanoasis/nerd-fonts?color=ee8449&style=flat-square)](#contributors)
+
 ## Contributors
 
 Thanks goes to these wonderful people ([emoji key](https://github.com/kentcdodds/all-contributors#emoji-key)):


realguse:
Another possibility is to use this https://shields.io/badges/git-hub-contributors, either with `contributors` or `contributors-anon`.

all-contributors badge: GitHub contributors from allcontributors.org

github contributors badge: GitHub contributors

github contributors badge (with `anon`): GitHub contributors

Finii (approved):
That is a hard-coded number? 🤦‍♀️ 

Thanks for noticing and creating this PR! 💚 

I manually counted the contributors in that file and it was `28 * 7 + 5 = 201` so I guess to have the number related to the contributors listed below what you chose is exactly the right count.

But seeing that contributors-anon is so many (which is just based on commits I assume, while the allcontributors number includes a lot of contributors without commit) I fear we have a lot people not on the allcontributors list. Well, probably from the first years.

Maybe I should just grep the git repo like

```console
$ git log | grep '^Author:' | sort | uniq | wc -l
     229
```

and add the missing contributors. But then, there are a lot duplicates on that list, people that have different emails in commits like Ryan or myself.

Whatever, what you did is a very good improvement!

Finii:
Ah, rate limits... That's the reason why the stars and fork count sometimes just do not work

Screenshot 2024-12-22 at 10 28 28

We could replace that with a workflow that updates hard coded count in the `CONTRIBUTORS.md` file 🤔

realguse:
Thanks, the github repo shows 204 so I don't know the actual number. Is the CONTRIBUTORS.md linked to the website in any way, because then i could make a seperate pr just checking all of the info etc.

Finii:
I guess it is good as-is.
It's the list of allcontributors and the shield just shows the number of people listed below, which is would I would expect.

PR #1768, opened 2024-12-22
fix(font-patcher): update `SymEnd` codepoint for devicons font

hasecilu:
#### Description

Updated `SymEnd` codepoint for devicons font in order to have the new icons available on Nerd Fonts.

#### Requirements / Checklist

Include glyphs that were left outside because a lack of update on the font-patcher script.

As you see in the devicons lib file there are some glyphs at the end that were not included on the last release.

https://github.com/ryanoasis/nerd-fonts/blob/76179ef368c98851353b90bfdb0a9da06d6c7017/bin/scripts/lib/i_dev.sh#L504-L515

Run font-patcher.
I was trying to use the YAML icon but despite appearing on lib and cheat-sheet there was no icon.
None
### online cheat-sheet

image

### E8 on NerdFonts Symbols Only from 3.3.0 release
2024-12-22-141747_hyprshot

### E7 on src/glyphs/devicons.ttf
2024-12-22-141007_hyprshot


Finii:
😱 

In the commit ab08bd9da46d9d4697a483b256d8125b1ce117bb comment I clearly talk about 496 glyphs, but change the End only to `E7E3` (from `E6C5`).
But `E600` + `1F0` (496) - `1` = `E7EF` as you said.

Thank you for finding this!!

Screenshot 2024-12-22 at 23 05 35

Finii (approved):

Finii:
Thanks again!!

I guess there might be an ordinary release in January anyhow (Monaspace for example seems tp get ready for a new release), so we can tackle this and the missing font updates, and ... sigh ;)

hasecilu:
> I guess there might be an ordinary release in January anyhow (Monaspace for example seems to get ready for a new release), so we can tackle this and the missing font updates, and ... sigh ;)

Great, meanwhile <> `e6a8` could be used.

Issue #1772, opened 2024-12-30
Patching Berkeley Mono v2 leads to incorrectly spaced chars


clementpoiret:
### Requirements

- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related or helpful
- [X] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [X] I have checked the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

### Experienced Behavior

The arrow ligature is offcentered to the left, and the `r` in the `import` word is to far to the right. It happens for many characters.
Even if the font is behind a paywall, the issue should also happen on the free trial version of Berkeley Mono v2. Command used to patch fonts:

```
sudo docker run --rm \
  -v /home/clementpoiret/.local/share/fonts/TX-02:/in \
  -v /home/clementpoiret/.local/share/fonts/TX02-NerdFont:/out \
  nerdfonts/patcher \
  --progressbars \
  --mono \
  --adjust-line-height \
  -c
```

### Expected Behavior

Same behavior as for other fonts, centered chars and glyphs.

### Example Symbols or Text

`-> import` but it happens for a lot of characters and nearly all ligatures.

### Font Used

Berkeley Mono v2

### Source of Font File

https://usgraphics.com/products/berkeley-mono

### Terminal Emulator (and the title of the terminal window)

ghostty, kitty

### Operating System and Version

NixOS Unstable

### Screenshots

Unpatched font:
![2024-12-30-At-10h39m30s](https://github.com/user-attachments/assets/7e96bce6-f8f5-4d62-8e70-3136566bf4bd)

Patched font:
![2024-12-30-At-10h37m13s](https://github.com/user-attachments/assets/dc94962f-4751-4039-9c56-445c81d539eb)

Finii:
Sorry to hear of your problems.

Just from your discription I'd say the problem is how the font constructs the ligatures. Probably not as CALT what is customary.
If that is the case there is nothing we can do for the parameters yoy give.

I guess the patched font is ok if you omit `--mono`? What is the reason for that flag, maybe we can solve that differently.

Also `--adjust-line-height`, do you have a specific reason to give that? I in fact wanted to remove that option several times already and I search for ppl to give me reason why to keep it.

Sent from Nine
________________________________
From: Clément POIRET ***@***.***>
[...]

PR #1773, opened 2024-12-31
Add new option to keep original glyphs as is when adding single-width icons

Finii:
#### Description

At the moment, if you want to have "monospaces" icons you always also force the font to be monospaced in Windows terms.
That has biten us in the past. Maybe It is better to leave that choice to the user and make this two options.

#### Requirements / Checklist

Remove `--use-single-width-glyphs` (to avoid confusion)
Add `--single-with-glyphs` which is analogous to `--variable-with-glyphs` and just affects our added icons.
The old `-s` and `--mono` force any font to be strictly monospaced and also sets the new `single-with-glyphs` option.

* #1772


Finii:
From the shellcheck CI

```
./bin/scripts/archive-fonts.sh:91:10: note: Double quote to prevent globbing and word splitting. [SC2086]
```

Finii:
Rebase on master, force push

Finii:
Also sneaking in the `--cell` expert option I always wanted to add but never came round to actually do it.
With that after-patching baseline-to-baseline tinkering is not needed anymore and the powerline glyphs match your personal choice.

Issue #1771, opened 2024-12-28
PyTorch logo has an artifact


aarmn:
### Requirements

- [X] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related or helpful
- [X] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [X] I have checked the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

### Experienced Behavior

pytorch logo look like sb by mistake put a dot in a random place of it!

### Expected Behavior

normal pytorch logo

### Example Symbols or Text

nf-dev-pytorch

### Font Used

Symbol Only

### Source of Font File

Same source as nerdfont font search website

### Screenshots

![image](https://github.com/user-attachments/assets/aee6060f-8ffc-4b63-818f-b4307df83da0)

Finii:
OH!

Thanks for reporting.

Dot needs to be near the opening.
That's already the 2nd logo in devicons with problems :-/

<img width="844" alt="Screenshot 2024-12-28 at 07 56 58" src="https://github.com/user-attachments/assets/49588655-38c6-4a6c-9bca-377bb1723845" />

Finii:
Note to self: Check also

overlapping outlines:

- bitbucket
- akka
- opera
- jira
- beats
- yii
- googlecloud
- graphql
- hibernate
- jetbrains
- kibana
- livewire
- materialui
- nimble
- norg
- openstack
- playwright
- plotly
- sequelize
- sqldeveloper
- streamlit
- tomcat
- vagrant
- vuestorefront
- woocommerce
- yunohost
- zig

small copyright circle or similar:
- archlinux
- arduino
- stata
- splunk
- spss
- unifiedmodellinglanguage

aarmn:
> OH!
> 
> Thanks for reporting.
> 
> Dot needs to be near the opening. That's already the 2nd logo in devicons with problems :-/
[...]

Issue #1760, opened 2024-12-07
Rust logo on Devicons.ttf is bad formed


atauzki:
### Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have checked the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

### Experienced Behavior

That's the icon in this repo's devicons.ttf
![image](https://github.com/user-attachments/assets/2257a4f9-fdf5-4ab9-9d4e-edb6d6dbc99c)

### Expected Behavior

The icon should be on the devicon repo.
![image](https://github.com/user-attachments/assets/890e7efa-639f-432b-a3b8-b20b120e5a0f)

Finii:
![image](https://github.com/user-attachments/assets/51d759ec-23be-4e0b-a4c5-faf46621efd3)

... investigating, thanks for reporting!

Finii:
You can use one of the other icons, until this is fixed.

![image](https://github.com/user-attachments/assets/f56cbbff-4b0b-4e54-a344-7b35fd0b00ae)

Finii:
In their SVG the missing holes are not closed paths, they start and end with a not-connected node on the same coordinate.
Normally fontforge should fix that on import, but here it seems to fail.

![Screenshot from 2024-12-07 19-52-39](https://github.com/user-attachments/assets/6c5f2de2-41d5-4238-acb3-8c19aaa9aff2)
_Fontforge_

![image](https://github.com/user-attachments/assets/4fed2cb1-b168-46e5-a8e6-2081f01794b8)
_Inkscape_

The arrows point to the smaller nodes that are end-nodes (2 of them on top of each other).
Strange fontforge manages to fix the left but not the right...

PR #1779, opened 2025-01-06
Bugfix some devicons icons

Finii:
#### Description

Some icons have problems, fix this manually in Inkscape.

Not all icons checked, just the ones that catched my eye on visual inspection.

#### Requirements / Checklist

* #1760 
* #1771


Finii:
These are the fixed icons, apart from `rust`. Notice the green lines which are paths that have been removed because they are inside existing path. While fonts _can_ work with this, and it is used with variable fonts always, in static fonts these can lead to rendering errors, especially on older renderers. And of course they are unneeded and/or crate selfintersecting paths, which are forbidden (but mostly tolerated by renderers).

The reason for these strange edges are a 'careless' conversion to monochrome (a.k.a. `plain`) upstream.
Automated tools are not always able to correctly remove these; but this feels here with these few icons it was not even attempted.

Also affected is upstream's `ttf` font, see image below.
The stylized letter `B` has this extra opening (?) with width almost? zero.

I feel I need to check all other icons systematically, these where just the visually conspicuous ones I found just by looking when fixing the PyTorch logo (see #1771). Some might follow later.

Ping to @Snailedlt from Devicons; I know you are no designer but maybe this is somehow interesting upstream.

image
_Fixes (dropped path parts) shown green_

image
_Bug? in upstream's font file_

Finii:
image
_Examples of more fixes, green = removed_

PR #978, opened 2022-10-20
Update seti to current version

Finii:
#### Description

They have added quite some more icons that people request/expect here also.

Unfortunately they seem to have no versioning anymore. So we update to this version

https://github.com/jesseweed/seti-ui/commit/fd20793e5a75b350eab8d489165fb9b420df3f

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/-/contributing.md)
- [x] Verified the license of any newly added font, glyph, or glyph set

#### What does this Pull Request (PR) do?

It adds a lot of new icons, existing/old ones are marked yellow in the following image.
The old codepoints are kept.
As they do not have stable codepoints that raises no additional issues for us (see [here](https://github.com/ryanoasis/nerd-fonts/issues/974#issuecomment-1283445156))

image

#### Any background context you can provide?

* #974

PR #1788, opened 2025-01-19
Improve gitlab icon

nataliakeniganti:
#### Description

Updates the GitLab icon as per https://about.gitlab.com/press/press-kit/.

#### Requirements / Checklist

- [x] Verified the license of any newly added font, glyph, or glyph set. License is: CC-BY-SA-4.0,3.0,2.5,2.0,1.0 license

### DIFF
--- a/src/svgs/gitlab.svg
+++ b/src/svgs/gitlab.svg
@@ -30,7 +30,7 @@
   <defs
      id="defs74">
     <style
-       id="style72">.cls-1{fill:#e24329;}.cls-2{fill:#fc6d26;}.cls-3{fill:#fca326;}</style>
+       id="style72">.cls-1{fill:#000;}.cls-2{fill:#000;}.cls-3{fill:#000;}</style>
   </defs>
   <g
      id="LOGO"
--- a/src/svgs/gitlab.svg
+++ b/src/svgs/gitlab.svg
@@ -1,18 +1,55 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   width="32"
-   height="32"
-   viewBox="0 0 125.785 125.785"
-   class="footer-logo"
-   version="1.1"
-   id="svg969"
+   xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
-   xmlns:svg="http://www.w3.org/2000/svg">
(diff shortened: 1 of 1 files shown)


Finii:
Thanks for the PR!

Technically the svg should be just one `path d`, while the svg here is colored. The font is monochrome, and the automatic conversion can and does fail (i.e. has suboptimal results) - thus the conversion to one path (without outline width, transforms, or clip paths).

Anyhow, I can fix that easily.

One problem I have with this is that you replace `nf-seti-gitlab`, which is an icon from the Seti set [1]. That would mean the original Seti and what we use differ. In fact the Seti set would have to be updated (first). Otoh I am a bit confused right now where the current icon comes from exactly. We updated Seti with
* #978 

but it seems there was a problem and later fixes and I need to untangle that first.

Probably the (your) icon needs to be added with a new codepoint as `nf-custom-github`.

Let me check that again when I'm on my real machine tomorrow or so.

[1] https://github.com/jesseweed/seti-ui

nataliakeniganti:
> Thanks for the PR!
> 
> Technically the svg should be just one `path d`, while the svg here is colored. The font is monochrome, and the automatic conversion can and does fail (i.e. has suboptimal results) - thus the conversion to one path (without outline width, transforms, or clip paths).
> 
> Anyhow, I can fix that easily.
> 
> One problem I have with this is that you replace `nf-seti-gitlab`, which is an icon from the Seti set [1]. That would mean the original Seti and what we use differ. In fact the Seti set would have to be updated (first). Otoh I am a bit confused right now where the current icon comes from exactly. We updated Seti with
> 
> * [Update seti to current version #978](https://github.com/ryanoasis/nerd-fonts/pull/978)
> 
> but it seems there was a problem and later fixes and I need to untangle that first.
> 
> Probably the (your) icon needs to be added with a new codepoint as `nf-custom-github`.
> 
> Let me check that again when I'm on my real machine tomorrow or so.
> 
> [1] https://github.com/jesseweed/seti-ui

Turn into monochrome

Finii:
Is there  reason you closed the PR? :thinking:

Sent from Nine
________________________________

Finii:
Screenshot 2025-01-19 at 21 46 13

Ok, whatever, the newer round form is available under other codepoints. (`nf-dev-gitlab` and `nf-fa-gitlab`)

PR #1079, opened 2023-01-26
No need --cask params anymore

Goooler:
#### Description

No need `--cask` param to install cask via brew.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/-/contributing.md)
- [ ] Verified the license of any newly added font, glyph, or glyph set

#### How should this be manually tested?

```sh
brew tap homebrew/cask-fonts
brew install font-hack-nerd-font
```

PR #1789, opened 2025-01-21
fix: update brew instructions

pablogrs:
Fix installation instructions using brew

#### Description
Current brew installation instructions do not work, this should fix it.

#### Requirements / Checklist

Fix the installation instructions

image

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -223,10 +223,10 @@ curl -OL https://github.com/ryanoasis/nerd-fonts/releases/latest/download/JetBra
 
 > Best option if on **macOS** and want to use **Homebrew**.
 
-All fonts are available via [Homebrew Cask](https://github.com/Homebrew/homebrew-cask) on macOS (OS X)
+All fonts are available via [Homebrew Cask](https://formulae.brew.sh/cask-font/) on macOS (OS X)
 
 ```sh
-brew install font-hack-nerd-font
+brew install --cask font-ubuntu-nerd-font
 ```
 
 ### `Option 3: Unofficial Chocolatey or Scoop Repositories`


Finii:
Thanks for the PR.

You mean
`brew install font-hack-nerd-font` failed (like shown above) and then you tried
`brew install --cask font-hack-nerd-font` and that succeeded?

Finii:
See the discussion at

* #1079

As for the link, not sure why you want to change it away from the Github repo, which _is_ the source of the casks.
* 96542aa9835f64f8279a5475de586273a514390d `doc: Correct moved Homebrew repo in readme.md`

pablogrs:
Thanks for the reply, I've tried adding the tap but without success, maybe I am adding a wrong one, but I didn't see it in the installation instructions either.
image

When installing with --cask it works (it works on my machine case?)
image

Finii:
Hmm, you use `brew tab homebrew/cask-fonts`, where did you get that?

That tap is deprecated, see https://github.com/Homebrew/homebrew-cask-fonts
and all the fonts are now in the 'default' tap.
96542aa9835f64f8279a5475de586273a514390d  `doc: Correct moved Homebrew repo in readme.md`

> maybe I am adding a wrong one, but I didn't see it in the installation instructions either

You do not need to add any tap.

> Distributor ID:	Ubuntu

Ah, you use Linux and not MacOS, I'll fire up the other machine, hold tight....

```
fini@Mac ~ % brew --version
Homebrew 4.4.17-15-g662a229
fini@Mac ~ % brew tap
fini@Mac ~ % brew search hack-nerd
==> Formulae
hackrf

==> Casks
font-hack-nerd-font                                                                                      hackmd
fini@Mac ~ %
fini@Mac ~ % brew info font-hack-nerd-font
==> font-hack-nerd-font: 3.3.0
https://github.com/ryanoasis/nerd-fonts
Not installed
From: https://github.com/Homebrew/homebrew-cask/blob/HEAD/Casks/font/font-h/font-hack-nerd-font.rb
==> Name
Hack Nerd Font (Hack)
==> Description
None
==> Artifacts
HackNerdFont-Bold.ttf (Font)
HackNerdFont-BoldItalic.ttf (Font)
HackNerdFont-Italic.ttf (Font)
HackNerdFont-Regular.ttf (Font)
HackNerdFontMono-Bold.ttf (Font)
HackNerdFontMono-BoldItalic.ttf (Font)
HackNerdFontMono-Italic.ttf (Font)
HackNerdFontMono-Regular.ttf (Font)
HackNerdFontPropo-Bold.ttf (Font)
HackNerdFontPropo-BoldItalic.ttf (Font)
HackNerdFontPropo-Italic.ttf (Font)
HackNerdFontPropo-Regular.ttf (Font)
==> Analytics
install: 6,025 (30 days), 18,952 (90 days), 48,590 (365 days)
fini@Mac ~ %
```

Finii:
### Note to self
Where I did the `brew info` above, where did the description go?
Obviously its not in the cask 🤔 

Screenshot 2025-01-22 at 07 40 15

pablogrs:
Thank you very much for following this up, it has to be an issue with my installation, I cannot see the same results as your commands.
image
I installed homebrew normally from [brew.sh](https://brew.sh/)

The cask name I got it from [here](https://gist.github.com/davidteren/898f2dcccd42d9f8680ec69a3a5d350e), but I should have read to the end, where it says there is no need for a cask anymore.

pablogrs:
maybe my brew installation is wrong

Finii:
> Thank you very much for following this up, it has to be an issue with my installation, I cannot see the same results as your commands.

I am confused. I said "Oh you use Ubuntu and not MacOS, let me try" and then I show ... my commands on a Macbook Air? :thinking: Maybe I intended to check it also on Ubuntu? And forgot or got sidetracked or someone walked into my office :thinking: 

image
_Running on Ubuntu 24.04 after fresh brew.sh install_

Ah I remember something like casks are not (normally) supported under Linux, but only formulae... Let me check...

Finii:
Casks are 'MacOS binary packages', and thus are not available on Linux. Which is unfortunate in the case of fonts :thinking: 

That's the reason you need `--cask` because casks are not considered on Linux.

I guess ;-) I can not find any explicit documentation. Doing a cask update on Linux is also not possible, I stumbled about that before: ... Cant find the Issue/Pull. Whatever.

image

----

So, at least until that changes we should probably hint that in the readme section. I will change it thus.
Thanks for the input and continued discussion :green_heart:

Finii:
Here the not-supported message for Casks on Linux, that has obviously changed (partially):

image

_https://github.com/ryanoasis/nerd-fonts/pull/1079_

See also
* https://github.com/orgs/Homebrew/discussions/3999 :grimacing: 
* https://github.com/Homebrew/homebrew-cask-fonts/issues/7040
* https://github.com/Homebrew/homebrew-linux-fonts

So according to the homebrew-linux-fonts, on Linux you tap _**it**_, and then get the fonts as Formulae (autotranslated).
Imho the fonts should have been Formulae to begin with, but :shrug: 
I think the solution for _us_ with `--cask` is better as no tapping is required and it lines up more closely the MacOS workflow.

PR #1844, opened 2025-04-16
Update Cascadia to 2407.24

Finii:
#### Description

Update Cascadia to 2407.24

#### Requirements / Checklist


Finii:
This also removes the subdir structure from Cascadia Code.

In preparation of the release it needs to clear out the patched fonts section too.

Issue #1848, opened 2025-04-22
Hurmit (Hermit) true type font converted from opentype font


duytrung:
### Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my request and found nothing related or helpful
- [x] I have checked the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have reviewed the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for relevant information

### Problem Description

Good day everyone,

I like the Hurmit aka Hermit, a nerd fonts that available from releases download section. However this is Opentype font. I used a web application (cloudconvert.com) to convert it from OTF to TTF. **_After conversion, the file sizes of each fonts reduced to a half_**.

I'm not sure whether this reduction affect the font display. Just use for some terminal app such as Neovim. I'm not sure why Windows Terminal sometimes didn't detect the OTF fonts instead of TTF fonts so I just try with TTF.

Is there any plan to release this font in TTF version along with OTF in the next release 3.4.0?

Great thanks

Finii:
Thanks for the interesting question/problem!

> After conversion, the file sizes of each fonts reduced to a half.

Well, I observed an even bigger factor, especially for the `Nerd Font Mono` variants. All `otf` fonts are rather big.

But there is a reason for that, and it can be 'fixed':

![Image](https://github.com/user-attachments/assets/f6ddaa49-d340-479f-9841-8a541ff2c779)
_Possible size reduction to 30%_

But to your question:

> Is there any plan to release this font in TTF version along with OTF in the next release 3.4.0?

No. For some fonts we had `otf` and `ttf` in the release set, but that has been dropped. 'Why' you might ask.
One problem is that in the inside the font formats are quite different, and so problems can turn up with one format but not with the other format. But when problems are reported people usually do not know which format they are using 😬 

Furthermore the hinting system is different, incompatible, and not convertible. Often users compare the patched and the unpatched font, and when we change the font format the hinting will be different and people WILL complain that the patched font looks different.
[...]

PR #1849, opened 2025-04-22
Use unrounded glyphs as source

Finii:
#### Description

We take svgs and produce intermediate font files; these font files will be used to patch in the glyphs.

When a `ttf` font is created the coordinates will be rounded to integers, and doing this twice (with scaling in between) adds more rounding errors than would be needed.

image
_Detail of `2665` in `octicons.ttf` (left) and `octicons.otf` (right)_

The imperfections of the left glyph will be exaggerated when it is scaled and again coordinate-rouneded, like for example:

image
_Patched Hermit with floating point node coordinates (left) and with integer rounding of coordinates (right)_

#### Requirements / Checklist

Minimize rounding errors introduced by the intermediate glyph fonts that we use to patch in lots and lots of svgs.

* #1848


Finii:
@allcontributors please add @duytrung for bug

PR #1851, opened 2025-04-22
font-patcher: Create smaller otf files

Finii:
[why]
The patched fonts in otf format are much bigger than the ttf counterparts. Even when the same font is used as source. The factor is about 2 to 3.5 in size.

[how]
Fonts in ttf format can not store nodes with sub-integer precision. The coordinates are rounded on export. But otf fonts can have real numbers (may it be floating point or fixed point notation, I am not sure right now and it does not matter) for coordinates, but the more precise numbers take much more space in the font file. If an otf is created where the coordinates are already integer this extra space is not needed.

So we do all the calculations with real numbers (including decimals) and only in the last step the coordinates of the added glyphs are rounded to integer (i.e. to the EM size grid).

Fixes: #1848

#### Requirements / Checklist

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -1586,17 +1586,11 @@ class font_patcher:
                     scale_ratio_x = scale_ratio_x * xy_ratio_max / xy_ratio
 
             if scale_ratio_x != 1.0 or scale_ratio_y != 1.0:
+                scale_ratio_x *= self.sourceFont.em / (self.sourceFont.em + 1) # scale a tiny bit too small to avoid rounding problems
                 self.sourceFont[currentSourceFontGlyph].transform(psMat.scale(scale_ratio_x, scale_ratio_y))
 
             # Drop nonintegral part of nodes' coordinates; ttf will do it anyhow, otf will be much smaller
             self.sourceFont[currentSourceFontGlyph].round()
-            if self.args.single:
-                (xmin, _, xmax, _) = self.sourceFont[currentSourceFontGlyph].boundingBox()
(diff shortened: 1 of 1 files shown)


Finii:
Commit Mono Nerd Font Mono

```
WARNING: Scaled glyph E645 wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph E8E7 wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph EED0 wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph EF96 wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph F13D wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph F02C8 wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph F0301 wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph F0C7A wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph F10FE wider than one monospace width (601 / 600 (overlap None))
WARNING: Scaled glyph F1314 wider than one monospace width (601 / 600 (overlap None))
```

Finii:
Maybe we should **always** scale a bit too small instead of 'exact', because exact will be bigger than the goal in 50% of the cases.
And a 1/600th to 1/2000th is not visible anyhow.
Aligning etc will be adjusted later, so right aligned should pose no problem even then.

Finii:
That commit (532cb9e) is a fix to the PR.

PR #1852, opened 2025-04-23
Fix alignment for some glyhs in propo

Finii:
#### Description

[why]
When the source glyph set is in a monospace group the proportional
scaling creates wrong advance widths and/or alignment.

Note that there are only very few monospace groups (i.e. 2).

[how]
Determine the 'cell' width, which can be one of three values depending
on mode. Use that width whereever a cell width is needed instead of the
fontwide-cellwidth (which might be wrong in proportional variants).

#### Requirements / Checklist

These are all affected in the symbols only font (usually too far right, outside 'cell')

image

### DIFF
--- a/font-patcher
+++ b/font-patcher
@@ -1070,9 +1070,8 @@ class font_patcher:
             range(0xf221, 0xf22d + 1), # gender or so
             range(0xf255, 0xf25b + 1), # hand symbols
         ]}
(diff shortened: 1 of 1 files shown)


Finii:
Here the new alignment (green) is wrong :-/

image

Finii:
The COD fixes are also real, here the `Nerd Font Propo` font before the PR, note that some glyphs are centered, while others are kind of left aligned, but the advance width is everywhere 1000, so ....

image

Issue #1795, opened 2025-02-05
Some Material Design Icon glyphs not in new v3.0 codepoints?


J-MR-T:
### Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and ~~found nothing related or helpful~~*included what I found*
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have checked the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

### Experienced Behavior

Hi :)

I'm in the process of replacing my version 2 nerd fonts across the board, so moving to the new code points for material design icons after #365 / #773. But in trying to migrate one of the glyphs: `nf-mdi-square_inc` (the old, conflicting codepoint was U+F9C9 (柳)), I couldn't find a replacement in the cheatsheet:

![Image](https://github.com/user-attachments/assets/9ad63552-87e2-47bc-94a2-2bb7181fdd1e)

(searching for just `square` doesn't bring up the icon either; nor do external tools like [nerdfix](https://github.com/loichyan/nerdfix))

Is this intended? Are there some material design icon glyphs that were simply removed and not allocated new codepoints in the move? From what I can tell, the symbol was not removed from MD itself, so I'm a bit confused ^^.

Finii:
According to this list: https://github.com/loichyan/nerdfix/issues/9#issuecomment-1576944348 I did a manual search for that `square_inc` icon and found no substitute (i.e. it has been dropped without replacement).

Looked manually through all MD icons, and the only half-way similar one is `square-circle` or `circle-box-outline`, but then even the border is different, and so I guess you would have to look yourself. I would use `fontforge` or similar for easier viewing.

> Is this intended? Are there some material design icon glyphs that were simply removed and not allocated new codepoints in the move?

Well, it was not only a move, it was also an update. So they are dropped upstream.

> the symbol was not removed from MD itself

Not?! Can you show me the symbol in MD?

J-MR-T:
Apologies, it seems like the page I found showing the icons was outdated itself (I was fooled by a "-dev" in the version name...), I now saw the official Google Fonts MDI browser, and can't find it there either, and the thread you link to also mentions it as being dropped without a replacement. Thanks for the clarification!

To make this clearer in the future, it might be worth amending this disclaimer on the cheatsheet page:

[...]

PR #1855, opened 2025-04-28
Fix cheatsheet footer last sentence tag

sliva0:
#### Description

Fixes random plus sign at the start of the line, which prevents correct application of `<h5>`  tag to the added sentence. Bug was introduced by https://github.com/ryanoasis/nerd-fonts/commit/a27373efd7dfc92426fe9f9a26004e27c9acd1f2, which copied changes from https://github.com/ryanoasis/nerd-fonts/commit/7a6b90f19a2fcc59162af363aeae7271b6b48b30, but it seems was a little sloppy and left out plus sign from manual git diff copying.

#### Requirements / Checklist

#### Relevant tickets
Original ticket is https://github.com/ryanoasis/nerd-fonts/issues/1795, fix to which introduced this bug.

#### Screenshots
image


Finii:
:sob:

> it seems was a little sloppy

Yes, was in a hurry.

Thanks for fixing!

Finii:
@allcontributors please add @sliva0 for doc

Issue #1864, opened 2025-05-05
DejaVuSansMNerdFont is not recognised as monospaced


pmadzik:
### Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [x] I have checked the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

### Experienced Behavior

DejaVuSansMNerdFont is not recognised as monospaced (xfce4-terminal, font-manager).

This is a consequence that 8 glyphs (uni2630 -- uni2637) have a width of 1652, I guess.

### Font Used

DejaVuSansMNerdFont-Regular.ttf

### Source of Font File

DejaVuSansMono.ttf

### Terminal Emulator (and the title of the terminal window)

xfce4-terminal

### Operating System and Version

Debian trixie/sid

Finii:
Thanks for reporting this.

Patching a font into a `Nerd Font` does usually not touch the existing glyphs in a font. They are only touched (forced to be monospaced) with a `Nerd Font Mono` font.

The trigraphs at `2630` - `2637` have that differing width already in the unpatched font:

![Image](https://github.com/user-attachments/assets/d4ef0450-6662-4ca9-8b00-1a96f9a5a60b)

The patcher does not touch any pre-existing trigraph.

And that font file is used already since ... years:

![Image](https://github.com/user-attachments/assets/130a39ea-a599-4f3c-ae52-4ced43b8bc19)

But wait, did WE add the extrawide trigraphs?

* https://github.com/ryanoasis/nerd-fonts/issues/100

Sigh, seems to just be copied in. I will do something.

Finii:
Also related:
* https://github.com/ryanoasis/nerd-fonts/issues/589

Finii:
Compare 'our' `DejaVuSansMono.ttf` versus the released one:

![Image](https://github.com/user-attachments/assets/33a78f8a-849c-4476-879c-44ea63b48bb0)

It must be this commit bc1f359497a8ff68685cbccf4ab32ab1da71496c

```
commit bc1f359497a8ff68685cbccf4ab32ab1da71496c
Author: Ryan L McIntyre <ryanoasis@gmail.com>
Date:   Sat Oct 22 12:57:41 2016 -0400

[...]

PR #1865, opened 2025-05-06
DejaVuSansMono: Fix monospaced

### LINKED ISSUE #100: Airline line indicator symbol is empty / not found
#### environment
- vim: 7.4
- OS: Debian Jessie
- terminal: XTerm
- $TERM variable: screen-256color-italic
- color configuration (:set t_Co?): 256
- font: DejaVu Sans Mono for Powerline Nerd Font Complete Mono.ttf

Hello,
A square character is appeared near line indicator, and I found g:airline_symbols.maxlinenr keeps that char and I am able to correct it from .vimrc, but I wonder why its default value is look like square.
That character is '☰'
screenshot from 2016-08-24 16 17 34

Finii:
**[why]**

The original / upstream version of DejaVuSansMono does not have the trigrams at codepoint U+2630 - U+2637. These have been added manually (together with some Powerline glyphs) with commit

bc1f35949  `Updates DejaVu fonts from version 2.33 to 2.37 & adds trigrams (2630 through 2627) (fixes #100)`

Unfortunately the newly added glyphs were just copied over from DejaVuSansCondensed and the width does not match the monospaced width of DejaVuSansMono: the resulting font is not monospaced anymore (but it should be, by its name part Mono).

**[how]**

Take the glyphs again from DejaVuSansCondensed but instead of directly transferring:
 - export glyphs as svg from Condensed font
 - in the Mono font set the width of the empty target glyphs to 1233
 - import the svgs
 - scale the glyphs X to 75% (100% in Y)
 - center the glyphs within the width

The exact same glyphs have been used for all 4 fonts.

**[note]**

Also remove the subdir structure as this is the new way to hold the font files. Remember to clean the patched-fonts/ directory up before release.

Fixes: #1864
Related: #100

Reported-by: @pmadzik

#### Requirements / Checklist


crochetedbytammysue:
Merge

PR #1866, opened 2025-05-09
Downgrade install.sh to bash 3

Finii:
#### Description
    [why]
    Vanilla MacOS has just some very old bash 3 installed by default. People
    might want to use the install.sh, but not install a newer bash (via
    Homebrew for example).

    [how]
    The `nerdfonts_dirs` array is replaced by a normal bash variable (string)
    where the directories are '/' separated. Because there are never
    subdirectories in the list but only direct children of 'patched-fonts/'
    it is possible to use slash as unique and out-of-band separator.

    The `fonts` array is replaced by a function that delivers the list of
    files in null-byte separated form. That means we can not manipulate the
    list and thus the duplicate checking is dropped. As we do not have any
    font in both formats (anymore) it will never happen anyhow.
    Dropping the duplicates check also removes the extension preference
    option functionality. We keep the now defunct option, though.

Well, the extension option is of course added back in another commit within this PR.

#### Requirements / Checklist

This works on Linux and needs a MacOS test.


Finii:
Checked on Sequoia 15.4.1 which has bash 3.2.57 :heavy_check_mark:

install.sh
```diff
@@ -1,21 +1,33 @@
 #!/usr/bin/env bash
 # Install Nerd Fonts
-__ScriptVersion="0.8"
+__ScriptVersion="1.0"
+
+# This script must run with bash 3
```
Finii, line 5: must be able to run with... sigh.

Issue #1877, opened 2025-06-06
Strange braking apart of name parts


Finii:
### Discussed in https://github.com/ryanoasis/nerd-fonts/discussions/1876

<div type='discussions-op-text'>

<sup>Originally posted by **mietzen** June  6, 2025</sup>

If I run with `--name 'MyFontForced'`, I get:

```Shell
Done with Patch Sets, generating font...
WARNING: Possible problem with the weight metadata detected, check with --debug
DEBUG: Weight approximations: OS2/PS/Name: 600/600/400 (from 600/'Demi'/'')
DEBUG: =====> Family (ID 1)      ok       (14 <=31): MyFont Ced For
DEBUG: =====> SubFamily (ID 2)   ok       ( 7 <=31): Regular
DEBUG: =====> Fullname (ID 4)    ok       (14 <=63): MyFont Ced For
DEBUG: =====> PSN (ID 6)         ok       (12 <=63): MyFontCedFor
DEBUG: =====> Filename '/out/MyFontCedFor-Regular.ttf'
```

All fonts are called `MyFontCedFor-Regular.ttf`, why? Also where does `Ced For` come from? Is this `Forced` splitted and swapped?</div>

Finii:
Can reproduce

![Image](https://github.com/user-attachments/assets/22fb2562-cf3d-474f-bc3b-eb33b0e104f4)

Finii:
The name `MyFontForced` is analyzed as this:

`[True, 'My', [], [], ['For'], 'Font Ced']`

(with `parse_ok` `basename` `weight_token` `style_token` `other_token` `rest`)

while `MyFontForced-Regular` is analyzed as this:

`[True, 'My Font Forced', [], ['Regular'], [], '']`

The correct case is parsed just normally, while the problematic one is parsed via `parse_simple_font_name()` with comment
```
"""Parse a fontname that does not follow the 'FontFamilyName-FontStyle' pattern"""
```

which tries on each CamelCase boundary (i.e. `My Font Forced`). That then triggers the special case for fonts with `SomeNameForPowerline` which is badly implemented and triggers on the `For` of `Forced` and ... well. Badly implemented ;)

Creating a fix, thank you @mietzen!

PR #1882, opened 2025-06-07
Bugfix removal of "For Powerline"

Finii:
#### Description

**[why]**
When the name to parse has 'for' in it - even when it is in the middle
of a word like 'formidable', 'unfortunate', or, 'forced' the "For" part
will be cut off and separated from the rest.

**[how]**
The reason is the code to remove 'For Powerline' from font names.

Instead of matching 'For' and 'Powerline' individually and then removing
the 'Powerline' and 'For' token later on (which fails if there is just a
'For' token like with 'Un' 'For' 'Tunate'), we now match only on a 'For'
if it is followed by a 'Powerline'.

To make the removal of the other-token easier later on we rename/unify
the token to be CamelCase.

Fixes: https://github.com/ryanoasis/nerd-fonts/issues/1877

Reported-by: @mietzen

#### Requirements / Checklist

Patch and see of the name is correct some fonts with 'For Powerline' in the name (there are several in `src/unpatched-fonts`) and some fonts with other 'for' substrings in the name.


Finii:
There is a similar problem with 'Book', but that is not so easy to solve. `UnBooked` is a hard case.

Will add the hint to the wiki that the name should be given 'full qualified' with a dash.

Finii:
@allcontributors please add @mietzen for bug

PR #1895, opened 2025-07-02
Add Server Mono

caidanw:
#### Description

Added the font Server Mono, followed the `contributing.md` guide.

#### Requirements / Checklist

- [x] Verified the license of any newly added font, glyph, or glyph set. License is: SIL Open Font License 1.1

Add the Server Mono font

Origin: https://github.com/internet-development/www-server-mono

I'm the maintainer of this font, please let me know if there are any questions/concerns and I'd be happy to help.

### DIFF
--- a/bin/scripts/lib/fonts.json
+++ b/bin/scripts/lib/fonts.json
@@ -891,7 +891,7 @@
       "unpatchedName": "Server Mono",
       "licenseId": "OFL-1.1-no-RFN",
       "RFN": false,
-      "version": "0.0.8",
+      "version": "1.200",
       "patchedName": "ServerMono",
       "folderName": "ServerMono",
       "imagePreviewFont": "ServerMono Nerd Font",
--- a/readme.md
+++ b/readme.md
@@ -176,7 +176,7 @@ A preview of all fonts can be found [here](https://www.nerdfonts.com/font-downlo
 | [RecMono Nerd Font][p-recursive]                  | [Recursive Mono][f-recursive]          | 1.085      | NO   |
 | [RobotoMono Nerd Font][p-roboto]                  | [Roboto Mono][f-roboto]                | 3.0        | NO   |
 | [SauceCodePro Nerd Font][p-source-code-pro]       | [Source Code Pro][f-source]            | 2.042      | YES  |
-| [ServerMono Nerd Font][p-server-mono]             | [Server Mono][f-server-mono]           | 0.0.8      | NO   |
+| [ServerMono Nerd Font][p-server-mono]             | [Server Mono][f-server-mono]           | 1.200      | NO   |
 | [ShureTechMono Nerd Font][p-share-tech-mono]      | [Share Tech Mono][f-share]             | 1.003      | YES  |
 | [SpaceMono Nerd Font][p-space-mono]               | [Space Mono][f-space]                  | 1.001      | NO   |
 | [Terminess Nerd Font][p-terminus]                 | [Terminus TTF][f-terminus]             | 4.49.3     | YES  |


caidanw:
@ryanoasis @Finii Are either of you able to review this please :)

src/unpatched-fonts/ServerMono/config.cfg
```diff
@@ -0,0 +1,4 @@
+# These config files are read by the font-patcher
+# They are INI style files, but some keys have JSON values
+[Config]
+commandline: --has-no-italic --boxdrawing --complete --adjust-line-height
```
Finii, line 4: ### `has-no-italic`

I fear that is not working, because it means in fact `has-oblique-instead-of-italic`, but here we have _Slanted_.

This needs a change in the name-parser (which is not a problem).

## `boxdrawing`

How strange is that, they have empty glyphs there? :sob: (Instead of no glyph).
I'd call that a bug in the source font.
:+1:

image

##  `complete`

That is the default

## `adjust-line-height`

Any reason for this?
I believe it is not changing anything and hunt for people who see a difference?

But here, the 'cell' height is already even, so it does not even do anything.

image

bin/scripts/lib/fonts.json
```diff
@@ -887,6 +887,21 @@
       "isMonospaced": true,
       "description": "Dashed zero, curved and straight character lines"
     },
+    {
+      "unpatchedName": "Server Mono",
+      "licenseId": "OFL-1.1-no-RFN",
+      "RFN": false,
+      "version": "0.0.8",
```
Finii, line 894: They put `1.0.0` resp. `1.000` into the font file, so that is an upstream bug :unamused:

Finii (commented):
You requested a review ;-)

The `fontfilenames` file has been forgotten (but I also hate that the information is scattered so much), probably we should autogenerate that file instead.

(No need for you to do anything, just noticing.)

https://github.com/ryanoasis/nerd-fonts/tree/master/src/unpatched-fonts#regenerating-the-fontfilenames-file

bin/scripts/lib/fonts.json
```diff
@@ -887,6 +887,21 @@
...
+      "patchedName": "ServerMono",
+      "folderName": "ServerMono",
+      "imagePreviewFont": "ServerMono Nerd Font",
+      "imagePreviewFontSource": "ServerMono/ServerMono-Regular.otf",
+      "linkPreviewFont": "server-mono",
```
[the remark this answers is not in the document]
Finii: That is the link addendum to https://programmingfonts.org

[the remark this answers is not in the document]
caidanw: Ah got it thanks! Will update!

src/unpatched-fonts/ServerMono/config.cfg
caidanw: Thank you for this, this incredibly helpful to understand the options.
I'll remove the default `complete` and useless `adjust-line-height`.

Yeah I am not sure what happened with the empty box drawing glyphs for the slanted style. I know our box drawing glyphs are slightly wrong in position by a point or two, so I opted to just replace them in their entirety with the option `boxdrawing`.

I'm fixing the font now to export under `Oblique` instead of `Slanted`, I'll update this PR with the new files.

caidanw:
Thanks for taking the time to review this and help guide me on the stuff I missed! I really appreciate the effort and knowledge! I agree that it'd be helpful for new folks like myself to have the documentation in one place (or at least pointing to the various places).

---

I noticed a bug when generating the `fontfilenames` file, I think the script may have been written with an older `jq` version because the `.fonts[$i]` gave an error. I slightly modified the script to make it work:
```diff
diff --git a/src/unpatched-fonts/README.md b/src/unpatched-fonts/README.md
index be50bdb4..4b597e19 100644
--- a/src/unpatched-fonts/README.md
+++ b/src/unpatched-fonts/README.md
@@ -13,8 +13,8 @@ all the fonts in `fontfilenames`. This can be used via
 
 Just call
 
-        jq -r '.fonts[$i].imagePreviewFontSource' ../../bin/scripts/lib/fonts.json \
+        jq -r '.fonts[].imagePreviewFontSource' ../../bin/scripts/lib/fonts.json \
             | grep -v '\.sfd$' \
             | sed 's/ /\\ /g' \
             | LC_ALL=C sort --ignore-case \
-            > fontfilenames
+            >| fontfilenames

```
I removed the problematic `$i` as it was an undeclared jq variable, and I changed the redirection to force overwrite even when folks like myself have `noclobber` enabled by default.

I can add this change to this PR, but if you'd prefer to keep it separate then I'll let you add it on your own accord.

Issue #1902, opened 2025-07-30
install script method deprecation complaint


laur89:
### Requirements

- [x] I have searched the [issues](https://github.com/ryanoasis/nerd-fonts/issues) for my issue and found nothing related or helpful
- [x] I have searched the [FAQ](https://github.com/ryanoasis/nerd-fonts/wiki/FAQ-and-Troubleshooting) for help
- [ ] I have checked the [Wiki](https://github.com/ryanoasis/nerd-fonts/wiki) for help

### Experienced Behavior

I'm registering my complaint regarding install script deprecation warning introduced in https://github.com/ryanoasis/nerd-fonts/commit/e8fa89bfcbc37ff0016ff1861386e655e0a9c1e9:
> Deprecated tool: Will not work to get newer fonts as they are not inside the repo anymore.

[Have always installed](https://github.com/laur89/dotfiles/blob/ceb0c902d6509184cbbd43b8192b8a0f294c1def/.bootstrap/install_system.sh#L5321-L5358) fonts using the [option #7](https://github.com/ryanoasis/nerd-fonts?tab=readme-ov-file#option-7-install-script) as it's done automatically by a shellscript.

What's the current go-to option for automated install, given option 7 is deprecated?

### Expected Behavior

Install script to work without any warnings.

Finii:
Thanks for the feedback.

How do you use the script... Install "all fonts" or some specific font(s)?
Can you give an example how it is usually called in your case?

laur89:
> Can you give an example how it is usually called in your case?

`Have always installed` in OP is linking to the shell logic in question. But essentially:
```sh
    fonts=(
        Hack
        SourceCodePro
        AnonymousPro
        Terminus
        ...
    )

    $ git shallow clone repo  (which is still ~8G lol)
    $ cd to repo dir

    for i in "${fonts[@]}"; do
        ./install.sh "$i"
    done
```

Finii:
> git shallow clone repo  (which is still ~8G lol)

Well, of course ;-}
That is why the `install.sh` is so bad.

While we have all the 'building blocks' to make a small script that just fetches the fonts you want, I did never come round to actually DO it - also because I thought noone is using it. But that does not seem very hard, so I will maybe modernize the `install.sh`, and also make it 'stand alone' (i.e. non-clone-repo, just download the script).

laur89:
[...]

PR #1913, opened 2025-08-15
[mod] fc_install: font config installer for the nerd-fonts

return42:
#### Description

Add script to install directly from the GH releases / no need to clone

#### Requirements / Checklist

As long the PR is not merged use this branch (`return42/nerd-fonts/bs-install`:

```bash
curl -s https://raw.githubusercontent.com/return42/nerd-fonts/bs-install/fc_install -o fc_install
chmod ugo+x fc_install
./fc_install --help
./fc_install install
```

- https://github.com/ryanoasis/nerd-fonts/issues/1902


Finii:
Thanks for the PR, it looks very good!

Well, Im only on my phone right now and can not really examine.
But two things come to my mind:

Iirc we have MIT license for all scripts, that simple "rule" would break with the license you choose.

Second I do not get the dependency on fc-cache. I didnt encounter a system that needs a fc-cache call for years.

And third ;-) apple has this bash 3 only, i would tweak the script to avoid the newer features.

But, whatever, great work!

return42:
> Iirc we have MIT license for all scripts, 

Sorry for my ignorance about .. changed lic to MIT --> [diff](https://github.com/ryanoasis/nerd-fonts/compare/cc005775f4d9558f2499680703ea51fcf4af68e6..22da4b4343e18d3ab6a4838d7b47514cd3eeec6f)

> Second I do not get the dependency on fc-cache.

fc-cache itself might not be needed, but the existence of this command ensures that a fontconfig [freedesktop.org](https://www.freedesktop.org/wiki/Software/fontconfig/) infrastructure (desktop vs. server distros) is installed  .. if you insist, I can remove the use of `fc-config`.

> apple has this bash 3 only

This script is written for Linux distros with a fontconfig [freedesktop.org](https://www.freedesktop.org/wiki/Software/fontconfig/) installation.

I sorted out my last Mac OS systems many years ago, but I suspect:

- Apple shipped the 2007 version of Bash v3 and cannot upgrade to Bash 4 due to GPL3 restrictions.
- Same with fontconfig / compared to what I know from Linux, Apple might have it own quirks (IDK but I would expect some differences)

Homebrew is the (missing/unofficially) package manager for Mac-OS and has a package for a updated version of `bash` .. however, Mac users are recommended to install the fonts from https://formulae.brew.sh/cask-font

return42:
To clarify: Installation scripts like the one in this PR do not replace a package manager; for example, this script does not offer automated updates (only manually or via a crontab entry).

Nerd Fonts should be installed (if possible) via the distribution's package manager. On voidlinux, for example, I use the [nerd-font](https://voidlinux.org/packages/?arch=x86_64&q=nerd-font) package. For macOS, Homebrew is (probably) recommended. TBH: I have no idea if there exists something similar for Windows .. [Scoop](https://scoop.sh/#/apps?q=nerd-fonts)? .. and [Scoop bucket for Nerd Fonts](https://github.com/matthewjberger/scoop-nerd-fonts)?

return42:
Last two force push fixed issues when TERM is unset and/or the script is used in a pipe (e.g. `VERBOSE=2 ./fc_install list | grep Nerd`).

Sorry for the *force push* mess / I hope I have now fixed last issues and PR is ready for review.

return42:
> Well, Im only on my phone right now and can not really examine.

ping @Finii .. if you have time, can you review / thanks ..

return42:
ping @Finii / FYI: last force push was just a rebase on current `main` .. can we progress with PR review?

mrsauravsahu:
Made a few changes, tested on Ubuntu 26.04 
https://github.com/mrsauravsahu/nerd-fonts/pull/1

Want to clean up the scripts a little bit more before merging

return42:
> Want to clean up the scripts a little bit more before merging

:shrug: if you are a maintainer, you can force push on this branch .. if not, open your alternative PR and wait (like me) years for a review .. I've given up hope of a review by now... but I'll leave the PR open for anyone interested in the patch.

Finii:
Sorry @return42, I see your frustration :unamused: 

It will probably not help, but I wanted to share _why_ this PR sits here seemingly unloved.

My main problem is lack of time. as you might have noticed there is also no new NF release despite the plans. There are other, more 'core' PRs also sitting and waiting. There are several rather personal reasons why the time is so scarce, but I can not help it.
The time that I find is mostly spent to 1st) improving the patching 2nd) updating glyph sets, 3rd) updating fonts, and 0th) answering questions and helping users. Adding yet another way to download the fonts is not really needed by many.

And then there are some complications with the PR in particular.

We already do have a script that fetches the release artifacts, which optionally lets you specify even the release version. This code duplication is not so nice in terms of maintenance.
There are other download-and-install helpers out there, they are references from our `readme.md`, but they are 3rd party tools in different repos. This splits the responsibility nicely. In fact all packaging and installing is done in other projects so that we can concentrate on the core work.

In this regard it is unclear for me, why you choose the `zip` instead of the `xz` archives, esp if you target Linux-like OSs only where nowadays `tar` supports both formats. Anyhow, that is small.

Then the script is very sophisticated with colored logging etc. Would it not make more sense to write this in a proper language instead of bash? And as I pointed out in the beginning, the script is so complex that you 'need' a recent bash, which excludes vanilla MacOS - and you do not care, 'they should use homebrew'. Why are you not using homebrew on your Linux? :grimacing:
Fontconfig is not related to the install in any case, but `fc-` is still in the name :thinking: 

The script does not - despite being Linux first - honor `XDG_*`?

But I appreciate the time you took to write the script, and in principle it could replace the obsolete `install.sh`.

After reading this again it might seem more negative than I intend it. In contrast, as I wrote above, I like the PR. But my private life is demanding right now and so it (the PR) falls off again and again :disappointed: It's not a casual glance and no-discussions-needed PR :woman_shrugging:

Sorry and thank you again!

Finii:
Ok, between 3.5.0 and 3.5.1 I had some time to look. Sorry again for the slow review rate 😒 

* I must admit this looks like - at least in part - AI generated code. There is a mix of simple and advanced techniques used, that I would not expect from a person coding this. We have no rules regarding AI usage, but I prefer when people are open / honest with tool usage. We have (at least) one PR where AI has been used, but the responsible author declared it as such. In principle there is no real difference, apart from me feeling more free to heavily (?) modify code that comes out of a LLM than touching code a person's heart went into.
* As I noted in the beginning I believe, I'm not sure if this needs to reference fontconfig in the name. A simple `install.sh` as name would imho be sufficient, and make this the natural replacement for the almost broken install script we have. And a nice one.
* Related to that, it is unclear why `fc-cache` is a hard runtime dependency. On usual modern systems that is not needed. Maybe it could call it if it exists, and omit the call if not. I believe we talked about that before, with no consensus.
* XGD support is missing
* Mac support is trivial to add but missing
* It uses `browser_download_url` instead of the asset ID, which would enable more uses of the install script (e.g. install unpublished releases) - getting there with a regex _is_ hard, and maybe not worth the work, but it could use `jq` (as soft dependency) and allow unpublished release installes at least when `jq` is installed. But then, the code would grow more complex. Hmm, all in all,, maybe good as is. [1]
* It does not utilise the `xz` artifacts 🤔 Reason unknown.
* There is no hint that `env` shows the variables one could influence. At least `VERBOSE` could be a command line option as we have with some other scripts
* API version 2022?

I would propose that I add a commit that:
* Change the name and make `fc-cache` optional
* Adds XGD and Mac
* Switches to `xz`
* Add more details to the help (that I foresee people will ask)
* And most importantly touch the code in 'many' places to make it (maybe) more maintainable for _me_

Once we have the changes that I would put on top, we can see what you say, and maybe it is then good for our both ideas 😇 

[1] I guess people who want an unpublished release are able to use `fetch-archives` instead. I use that all the time.

return42:
> I must admit this looks like - at least in part - AI generated code.

Nope, not a single line is from AI .. nor any AI was used to evaluate this solution.

*Since the beginning >40 years ago , I tried to implement in a structured way -- long before LLMs existed -- perhaps my way of thinking is a bit mechanical, which might give the impression that AI was involved*  ;-)

> As I noted in the beginning I believe, I'm not sure if this needs to reference fontconfig in the name. A simple `install.sh` as name would imho be sufficient, and make this the natural replacement for the almost broken install script we have. And a nice one.
...
> Mac support is trivial to add but missing
> Related to that, it is unclear why `fc-cache` is a hard runtime dependency ...  I believe we talked about that before, with no consensus.

- https://github.com/ryanoasis/nerd-fonts/pull/1913#issuecomment-3194163677


Initially, I wanted to modify the existing `install.sh` script, but I realized that doing so would require changes that would break backward compatibility. Since I only use Linux, I lack the knowledge to adapt the script for other operating systems, nor would I be able to test my modifications.


Ultimately, I needed an installer for `fontconfig`, which is how the name `fc_install` came about.

> XGD support is missing

:+1: 

> * It does not utilise the `xz` artifacts 🤔 Reason unknown.

`xz` and the `jq` you suggested .. are dependencies that might need to be installed first... `zip` is available on every system (*less dependencies*).  BTW I would offer an installation for users (admins); developers who need installations outside of the official releases will have their own solutions.

> esp if you target Linux-like OSs only where nowadays `tar` supports both formats. Anyhow, that is small.

If that’s the case, then `xz` really should be used :+1:  .. it didn't exist in the past, and in some cases it depended—or still depends—on the tar build .. *Some of my experiences date back quite a long time.* :-)

> * API version 2022?

When I posted the PR, that was the current release of the API; a new release didn't appear until 2026.

https://docs.github.com/de/rest/about-the-rest-api/breaking-changes?apiVersion=2026-03-10

> * There is no hint that `env` shows the variables one could influence.

:+1: 

> Add more details to the help (that I foresee people will ask)
> And most importantly touch the code in 'many' places to make it (maybe) more maintainable for *me*

.. whatever it means, *very vague requirements to me*  :shrug:

----

I’ve been using the script for two years for its intended purpose (Linux, fontconfig), and it works perfectly fine...

I’m tied up with other projects now and unfortunately don’t have the time to look after it anymore; I’m going to mark the PR as a draft. Maybe someone else will have the time to implement your additional requirements.

---

> In fact all packaging and installing is done in other projects so that we can concentrate on the core work.

I completely understand .. I'm a developer myself, and deployment isn’t exactly one of my strength ;-)

*Perhaps it really doesn’t belong here.*

Thanks for taking the time... and thanks also for nerd-fonts in general.

Finii:
These you might want to change in your version also (i.e. bugs)

* 2b211fb
* d2ac674
* (2fa9b7f)
* 38868a0
* 40c6a2b
* b5ea9ed

Thanks again for the script and your energy!!

I will check this on Mac later, lets see what we can do.

return42:
I’ve taken a quick look at your changes; everything looks quite good to me .. and it is definitely a significant improvement.

Unfortunately, I don't have the time right now to look into it more deeply, so please feel free to handle the branch however you like .. squash, force-push, merge, whatever and whenever you want. It’s all fine by me, and I don't mind whether my name is associated with it or not. 

> Thanks again for the script and your energy!!

If it helps the project, then my work has served its purpose, that’s all I care about. :-)

Again: thanks a lot :+1:

Finii:
Works on MacOS now, with both stock `bash v3` and Homebrew `bash v5`

Screenshot 2026-09-04 at 00 09 57

Finii:
The _use-color-yes-or-no_ had also a bug, maybe backport...

* e0bd379

Finii:
@allcontributors please add @return42 for code

Finii:
Thanks again. That was a long standing issue that I never found the energy to even start to tackle.
Very happy with the result :heart: :tada:

PR #1925, opened 2025-09-07
Scale non-mono icons to be at least as large as mono

danielwe:
#### Description

For icons with the `pa` stretch rule, `font_patcher` currently scales icons up to fit the cell in monospace patching, but only scales down for non-monospace. This makes some icons significantly _smaller_ with non-monospaced patching than with monospaced, which is rather unexpected. Case in point: the `nd-md-alpha` alphabetic icons are tiny in non-monospaced fonts, but substantial in monospaced.

With this PR, non-monospaced patching never makes smaller icons than monospaced. For `pa`, if monospaced patching would scale up, non-monospaced scales up by the same amount, but no more.

#### Requirements / Checklist

* Factors the logic for determining `iconheight` out into a separate function that takes an optional `single_width` parameter
* Uses this function to correctly calculate the scale factors that would be obtained in `single_width` when needed
* For `pa` icons that don't need downscaling, sets the scale factor to the maximum of 1 and the single width scale factor.

* Patch a font in both mono and non-mono
* Look at `󰬇 (nd-md-alpha_z)` in both variants at the same font size
* With the previous patcher, you'll see something like this, where the larger icon is with the Mono variant
  Screenshot 2025-09-07 at 01 46 35Screenshot 2025-09-07 at 01 46 54
* With this PR, the icon will have the larger size in both variants

See explanation.

NA.

See above.


danielwe:
I borked a force push from a shallow clone, so this PR died. Resubmitted as #1926.

PR #1980, opened 2026-01-22
Add Braille support (with generator)

### LINKED ISSUE #482: Support for Braille Characters?

Finii:
#### Description

Add Braille Glyphs that are generated on the fly to exactly match the 'cell' size of the patched font.
Just scaling is not really possible for circular dots. When they are expected to be circular in the patched font.

Created this PR in @GoldPigg's stead.

I believe it is easier to discuss details in this/a separate PR.

Thanks for preparing the code extension!

#### Requirements / Checklist

See discussion in 'parent' PR
* #837 

Closes #482

### DIFF
--- a/bin/scripts/braille/Braille.py
+++ b/bin/scripts/braille/Braille.py
@@ -51,21 +51,16 @@ def draw_braille_glyph(glyph, idx, width, ymax, ymin, r):
     pen = None
 
 
-def set_scent(font, ymax, ymin, ascent, descent):
+def set_scent(font, ymax, ymin):
     """" Set ascent and descent """
     # This may not impact the final patch result,
     # but to better check middle result, we set them.
-    # Note: Sometimes `ymax` may exceed `ascent`. 
-    #       This is because some fonts have a system ascent (e.g., `hhea_ascent`) that bigger than 
-    #       their own `ascent` value. Setting `font.ascent` directly to `ymax` may cause the height to 
-    #       exceed the em, triggering the em adjustment mechanism and resulting in errors.
-    font.ascent = ascent
-    font.descent = descent
-
     # We set them simply, because this is enough
     # to reveal the true situation when we check the middle result.
     font.hhea_ascent = ymax
     font.hhea_descent = ymin
+    font.hhea_ascent_add = 0
+    font.hhea_descent_add = 0
     font.hhea_linegap = 0
 
     font.os2_winascent = ymax
@@ -75,11 +70,13 @@ def set_scent(font, ymax, ymin, ascent, descent):
 
     font.os2_typoascent = ymax
     font.os2_typodescent = ymin
+    font.os2_typoascent_add = 0
+    font.os2_typodescent_add = 0
     font.os2_typolinegap = 0
 
 
 
-def get_braille_font(em, width, ymax, ymin, descent, ascent, r_ratio):
+def get_braille_font(em, width, ymax, ymin, r_ratio):
     """ Get braille font where the ratio of the radius to the shortest distance from the center is `r_ratio` """
     braille_font = fontforge.font()
 
@@ -92,15 +89,16 @@ def get_braille_font(em, width, ymax, ymin, descent, ascent, r_ratio):
     braille_font.version = "1.0.0"
 
     braille_font.em = em
-    set_scent(braille_font, ymax, ymin, descent, ascent)
+    set_scent(braille_font, ymax, ymin)
 
     r = min(width / 4, (ymax - ymin) / 8) * r_ratio
 
     START = 0x2800
     END = 0x28FF
     for i in range(START, END + 1):
-        glyph = braille_font.createChar(i)
+        idx = i - START
+        glyph = braille_font.createChar(i, f'braille-{idx}')
         glyph.width = width
-        draw_braille_glyph(glyph, i - START, width, ymax, ymin, r)
+        draw_braille_glyph(glyph, idx, width, ymax, ymin, r)
 
     return braille_font
--- a/bin/scripts/braille/Braille.py
+++ b/bin/scripts/braille/Braille.py
@@ -1,0 +1,106 @@
+import fontforge
+import math
+
+
+def get_circle_center(dot, width, ymax, ymin):
+    """ Get the center of the circle centered at the point with ID `dot` """
+    # Note dots' position(a little strange):
+    # 0 3 
+    # 1 4
+    # 2 5
+    # 6 7
+    if dot < 6:
+        row = dot % 3
+        col = dot // 3
+    else:
+        row = 3
+        col = dot - 6
+
+    height = ymax - ymin
+    x0 = width / 4
+    y0 = -height / 8 + ymax
+    return (x0 + col * width / 2, y0 - row * height / 4)
+
+
+def draw_circle(pen, center, r):
+    """ Control the pen to draw a circle """
+    K = 4 / 3 * (math.sqrt(2) - 1)  # a ratio for determining control points
+    offset = r * K  # offset distance of control points
+
+    cx, cy = center
+    tx, ty = cx, cy + r  # top
+    rx, ry = cx + r, cy  # right
+    bx, by = cx, cy - r  # bottom
+    lx, ly = cx - r, cy  # left
+
+    pen.moveTo((tx, ty))
+    pen.curveTo((tx + offset, ty), (rx, ry + offset), (rx, ry))  # top -> right
+    pen.curveTo((rx, ry - offset), (bx + offset, by), (bx, by))  # right -> bottom
+    pen.curveTo((bx - offset, by), (lx, ly - offset), (lx, ly))  # bottom -> left
+    pen.curveTo((lx, ly + offset), (tx - offset, ty), (tx, ty))  # left -> top
+    pen.closePath()
+
+
+def draw_braille_glyph(glyph, idx, width, ymax, ymin, r):
+    """ Draw the braille glyph with the corresponding number """
+    pen = glyph.glyphPen()
+    for i in range(8):
+        if (1 << i) & idx > 0:
+            center = get_circle_center(i, width, ymax, ymin)
+            draw_circle(pen, center, r)
+    pen = None
+
+
+def set_scent(font, ymax, ymin, ascent, descent):
+    """" Set ascent and descent """
+    # This may not impact the final patch result,
+    # but to better check middle result, we set them.
+    # Note: Sometimes `ymax` may exceed `ascent`. 
+    #       This is because some fonts have a system ascent (e.g., `hhea_ascent`) that bigger than 
+    #       their own `ascent` value. Setting `font.ascent` directly to `ymax` may cause the height to 
+    #       exceed the em, triggering the em adjustment mechanism and resulting in errors.
+    font.ascent = ascent
+    font.descent = descent
+
+    # We set them simply, because this is enough
+    # to reveal the true situation when we check the middle result.
+    font.hhea_ascent = ymax
+    font.hhea_descent = ymin
+    font.hhea_linegap = 0
+
+    font.os2_winascent = ymax
+    font.os2_windescent = -ymin
+    font.os2_winascent_add = 0
+    font.os2_windescent_add = 0
+
+    font.os2_typoascent = ymax
+    font.os2_typodescent = ymin
+    font.os2_typolinegap = 0
+
+
+
+def get_braille_font(em, width, ymax, ymin, descent, ascent, r_ratio):
+    """ Get braille font where the ratio of the radius to the shortest distance from the center is `r_ratio` """
+    braille_font = fontforge.font()
+
+    braille_font.fontname = "BrailleFont-Regular"
+    braille_font.familyname = "Braille Font"
+    braille_font.fullname = "Braille Font Regular"
+    braille_font.encoding = "UnicodeFull"
+
+    braille_font.copyright = ""
+    braille_font.version = "1.0.0"
+
+    braille_font.em = em
+    set_scent(braille_font, ymax, ymin, descent, ascent)
+
+    r = min(width / 4, (ymax - ymin) / 8) * r_ratio
+
+    START = 0x2800
+    END = 0x28FF
+    for i in range(START, END + 1):
+        glyph = braille_font.createChar(i)
+        glyph.width = width
+        draw_braille_glyph(glyph, i - START, width, ymax, ymin, r)
+
+    return braille_font
--- a/bin/scripts/braille/Braille.py
+++ b/bin/scripts/braille/Braille.py
@@ -22,6 +22,20 @@ def get_circle_center(dot, width, ymax, ymin):
     return (x0 + col * width / 2, y0 - row * height / 4)
 
 
+def draw_rectangle(pen, center, rx, ry):
+    """ Control the pen to draw a rectangle """
+
+    cx, cy = center
+
+    pen.moveTo((cx - rx, cy + ry))
+    pen.lineTo((cx + rx, cy + ry))
+    pen.lineTo((cx + rx, cy - ry))
+    pen.lineTo((cx - rx, cy - ry))
+    pen.lineTo((cx - rx, cy + ry))
+
+    pen.closePath()
+
+
 def draw_circle(pen, center, r):
     """ Control the pen to draw a circle """
     K = 4 / 3 * (math.sqrt(2) - 1)  # a ratio for determining control points
@@ -41,13 +55,16 @@ def draw_circle(pen, center, r):
     pen.closePath()
 
 
-def draw_braille_glyph(glyph, idx, width, ymax, ymin, r):
+def draw_braille_glyph(glyph, idx, width, ymax, ymin, style, rx, ry):
     """ Draw the braille glyph with the corresponding number """
     pen = glyph.glyphPen()
     for i in range(8):
         if (1 << i) & idx > 0:
             center = get_circle_center(i, width, ymax, ymin)
-            draw_circle(pen, center, r)
+            if style == 'rectangle':
+                draw_rectangle(pen, center, rx, ry)
+            elif style == 'circle':
+                draw_circle(pen, center, min(rx, ry))
     pen = None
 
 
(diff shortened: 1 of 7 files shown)


GoldPigg:
I'm glad I could help you.

### [NOTE]
**I seem to have forgotten to update the version of `font-patcher`. You might need to update it.**

GoldPigg:
> You create the classic round 'dots', what do you think about the octants style proposed above?

I'm not entirely sure about this, because the reason I'm doing this work is that I myself am a **victim of non-equidistant Braille fonts**. Therefore, classic round ‘dots’ represent my **original requirement**.

And I suspect most users feel the same way. Perhaps we could introduce **command-line arguments** for customization, or allow both styles to **coexist**.

However, if adding octants style is desired, I don't think it would be difficult. Based on the existing code, we could reuse most of it—basically just adjust the `draw` function.

If you feel it's truly necessary, I could give it a try. But I might not have much free time recently, so it might not get done quickly.

What do you think?🤔

Finii:
* https://github.com/ryanoasis/nerd-fonts/discussions/1931

Finii:
> > what do you think about

> What do you think?🤔

:-D

I'm not sure I understand this:

> the reason I'm doing this work is that I myself am a victim of non-equidistant Braille fonts. Therefore, classic round ‘dots’ represent my original requirement.

If the 'dots' are circular or rectangles does not change the equidistance? Of course they can not / should not be square.

And yes, your code design is nice where the circles can easily be exchanged for rectangles.

In the other thread "the people" seemed to like octants more, so I wanted to know your preference / reasoning.
Maybe you want to try rectangular dots with your "image"?

> And I suspect most users feel the same way.

I had the impression most would prefer rectangular 'dots'. Equidistant of course. Or even gapless (but I know from experience that gapless is always hard to achieve over different clients that render the glyphs differently).

> command-line arguments for customization

The problem is that only ONE style could be the default; and most of the time people use the prepatched fonts (which use the defaults of course).

_Edit: Add last parenthesised sentence to better explain what I mean_

bin/scripts/braille/Braille.py
```diff
@@ -0,0 +1,106 @@
...
+    set_scent(braille_font, ymax, ymin, descent, ascent)
+
+    r = min(width / 4, (ymax - ymin) / 8) * r_ratio
+
+    START = 0x2800
+    END = 0x28FF
+    for i in range(START, END + 1):
+        glyph = braille_font.createChar(i)
+        glyph.width = width
+        draw_braille_glyph(glyph, i - START, width, ymax, ymin, r)
```
Finii: Setting the `glyph.name` might be good; I'm not sure fontforge sets them if unspecified to the unicode values, and even if, I believe that can be turned off when fontforge is compiled. Hmm, unsure.

But it would certainly be good to have meaningful names on the glyphs.

Finii (commented):
Again: Nice work

This is not a full review, just listing the things that crossed my mind when reading your code.

GoldPigg:
> I'm not sure I understand this

I'm sorry for my confusing expression.

I've now re-evaluated the two options. Although I've generated rectangles as squares, it doesn't significantly impact the outcome. We can make a simple adjustment later.

**circular**:
circular


**rectangles**:
rectangles

Now my opinion seems to have changed—rectangles do appear to work better (may more rich and full).

GoldPigg:
I fixed the Rectangle font .it's no longer square ,although it doesn't seem particularly noticeable in this font. 

And I generated a nearly gapless font. Here's the result:

- **rectangle**：
   rectangle
   
- **gapless**:
   gapless

Below are the various fonts generated. You may use them for local testing.

[BrailleFonts.zip](https://github.com/user-attachments/files/24837102/BrailleFonts.zip)

I hope this helps you.

And if you need, I can also upload the code to branchs in my repository for your reference.

Finii:
> And if you need, I can also upload the code to branchs in my repository for your reference.

That would be nice!

GoldPigg:
> That would be nice!

Ok, it's now on [**support-braille-style**](https://github.com/GoldPigg/nerd-fonts/compare/braille-support...GoldPigg:nerd-fonts:support-braille-style)

I set **rectangle as the default** because I conducted a quick survey among my friends, and the results showed that **10 people chose rectangle**, **7 chose gapless**, and **3 chose circle**.

Finii:
Great!

Gapless is always a problem depending on the client the user uses. To circumvent them we usually overlap the cell with its neighbor by some percent; and still we get colored lines sometimes. Search for colored vertical lines if interested. So that would be a specialty anyhow.

I tried to clean up the PR (remove the two first-attempt commits not-by-you, and rebase on current `master`. But it seems I can't force push to your branch (you removed the maintainers may modify PR checkmark, I assume).
Additionally I wanted to add a commit that adds the braille/ file to the two font-patcher releases and increase the font-patcher script version.

There might come up other small things that I would "just fix" normally. If you are comfortable with me messing inside your repo/PR branch maybe you can allow that (there's a tick mark on the PR page I believe, on the right hand sidebar).

Thank you so much for the contribution!

Finii:
Here a commit that can be added via `git am ...`
This is after a `git rebase origin/master`.
And later pushed via `git push --force-with-lease`.

```diff
From 5584bdece502c84dcd4be4220e846882afcc111f Mon Sep 17 00:00:00 2001
From: Fini Jastrow <ulf.fini.jastrow@desy.de>
Date: Mon, 26 Jan 2026 15:24:56 +0100
Subject: [PATCH] Include Braille generator in release variants

Releases of the font-patcher script:
* Docker release
* Zip release

Signed-off-by: Fini Jastrow <ulf.fini.jastrow@desy.de>
---
 .dockerignore                        | 1 +
 .github/workflows/docker-release.yml | 1 +
 bin/scripts/archive-font-patcher.sh  | 3 ++-
 font-patcher                         | 2 +-
 4 files changed, 5 insertions(+), 2 deletions(-)

diff --git a/.dockerignore b/.dockerignore
index 0283a0b24..37f3c20e8 100644
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,5 +4,6 @@
 !font-patcher
 !glyphnames.json
 !bin/scripts/name_parser/Fontname*.py
+!bin/scripts/braille/Braille.py
 !bin/scripts/docker-entrypoint.sh
 !.codeclimate.yml
diff --git a/.github/workflows/docker-release.yml b/.github/workflows/docker-release.yml
index 9d6765db9..46ef5889c 100644
--- a/.github/workflows/docker-release.yml
+++ b/.github/workflows/docker-release.yml
@@ -8,6 +8,7 @@ on:
       # Keep this in line with .dockerignore
       - bin/scripts/docker-entrypoint.sh
       - bin/scripts/name_parser/Fontname*.py
+      - bin/scripts/braille/Braille.py
       - src/glyphs/**
       - .dockerignore
       - Dockerfile
diff --git a/bin/scripts/archive-font-patcher.sh b/bin/scripts/archive-font-patcher.sh
index e650c4986..5d224687d 100755
--- a/bin/scripts/archive-font-patcher.sh
+++ b/bin/scripts/archive-font-patcher.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # Nerd Fonts Version: 3.4.0
-# Script Version: 1.1.0
+# Script Version: 1.2.0
 # Archives the font patcher script and the required source files
 # If some (any) argument is given this is though of as intermediate version
 # used for debugging
@@ -28,6 +28,7 @@ find "${outputdir:?}" -name "FontPatcher.zip" -type f -delete
 
 cd -- "$scripts_root_dir/../../" || exit 1
 find "src/glyphs" | zip -9 "$outputdir/FontPatcher" -@
+find "bin/scripts/braille" -name "Braille.py" | zip -9 "$outputdir/FontPatcher" -@
 find "bin/scripts/name_parser" -name "Fontname*.py" | zip -9 "$outputdir/FontPatcher" -@
 find "glyphnames.json" | zip -9 "$outputdir/FontPatcher" -@
 find "font-patcher" | zip -9 "$outputdir/FontPatcher" -@
diff --git a/font-patcher b/font-patcher
index 8fb3550a9..a37d03692 100755
--- a/font-patcher
+++ b/font-patcher
@@ -6,7 +6,7 @@
 from __future__ import absolute_import, print_function, unicode_literals
 
 # Change the script version when you edit this script:
-script_version = "4.21.0"
+script_version = "4.22.0"
 
 version = "3.4.0"
 projectName = "Nerd Fonts"
-- 
2.48.1

```

GoldPigg:
> there's a tick mark on the PR page I believe, on the right hand sidebar

The option doesn't seem to be available on the right side. I suspect this might be because you initiated this PR, so I added you to the Collaborators list. This may solve the poblem.
screenshot

Finii:
Removed the first two commits (that were from the other PR / first try) unrelated to this PR.

That needed of course a force push, make sure to - for example - `git fetch && git reset --hard origin/master`.

image

Finii:
I broke it :grimacing:  ... fixing

image

Finii:
Sorry for the frequent force-pushes, I was not concentrated enough :-(
Should not work on this while at-work :grimacing: but wanted to show some progress.
Will only add and not force-push from now on :angel: 

Thanks again!

Finii (commented):
Fast review, still not in depth ;)

Finii (commented):
I did not find any other items to discuss.
This looks excellent! Thank you so much!

From my side we could pull this now; please tell me if you like all my changes or if we need to tweak something more.

Ah the only thing is if we want

```
font-patcher --braille --braile-style circle xxx.ttf
font-patcher --braille circle xxx.ttf
```

I'm not certain. The more options there are, the more complicated it looks, better maybe to have one a bit more complicated option instead of two, where one is just also complicated?

On the other hand, all glyph sets options so fare are booleans. That would be the first which needs a parameter.

But then, people usually use `--complete` anyhow.

I think all in all I would drop `--braille-style` and merge into `--braille` but I did not want to do that without first consulting with you, as that is a "bigger" change of your ideas.

GoldPigg:
> please tell me if you like all my changes or if we need to tweak something more

I think everything is **nice** now.🎉

---

> I think all in all I would drop `--braille-style` and merge into `--braille` but I did not want to do that without first consulting with you, as that is a "bigger" change of your ideas.

I think this is a **good idea**. In fact, that's exactly what I had in mind from the start, but just as you said:

> On the other hand, all glyph sets options so fare are booleans. That would be the first which needs a parameter.

Therefore, I did't do it, as I'm afraid it might cause some unexpected issues since I'm not familiar enough with the complete code. However, I believe **you can be able to handle it well.👍**

And Here's one potential point I thought of that might require attention(may this could help you):
```python
args.braille = True
``` 
This line is set when `--complete` is enabled. When modifying it, we **cannot directly assign a value**. We must **first check** whether `--braille` has been set. If not, then assign the value. Otherwise, `--braille` will become inactive when `--complete` is used.

However, if we continue to store style using `args.brailleStyle` but assign it via `--braille`, this issue seems unlikely to occur, and it may require **minimal code changes**.

GoldPigg:
I attempted the second approach. Below is the commit (on the `merge-two-args` branch):

[Merge `--braille-style` into `--braille`](https://github.com/ryanoasis/nerd-fonts/commit/26748a7949ed41513bf36bab781645dd43c50c08)

I added the Braille suffix. It seems to have been accidentally deleted while organizing the commit history. Below is the commit:

[font-patcher: Add Braille Suffix](https://github.com/ryanoasis/nerd-fonts/commit/84cf7066fd56b6a6bbe596a1e0276a6188bbcf20)

They are on the same branch. You can merge them directly or make modifications before merging.

Finii:
Ah, thank you!

> Braille suffix. It seems to have been accidentally deleted

In fact I dropped it willingly, as I doubt the usefullness of the suffix. Further up the commnet says:

```python
        if not self.args.complete:
            # NOTE not all symbol fonts have appended their suffix here
            if self.args.fontawesome:
                additionalFontNameSuffix += " A"
                verboseAdditionalFontNameSuffix += " Plus Font Awesome"
```

Lets see if the comment is correct

```
    sym_font_group.add_argument('--braille',
    sym_font_group.add_argument('--complete',
    sym_font_group.add_argument('--codicons',
    sym_font_group.add_argument('--fontawesome',
    sym_font_group.add_argument('--fontawesomeext',
    sym_font_group.add_argument('--fontlogos',
    sym_font_group.add_argument('--material',
    sym_font_group.add_argument('--octicons',
    sym_font_group.add_argument('--pomicons',
    sym_font_group.add_argument('--powerline',
    sym_font_group.add_argument('--powerlineextra',
    sym_font_group.add_argument('--powersymbols',  
    sym_font_group.add_argument('--weather',
```
```
        if not self.args.complete:
            # NOTE not all symbol fonts have appended their suffix here
            if self.args.braille:    
            if self.args.codicons:
            if self.args.fontawesome:
            if self.args.fontawesomeextension:
            if self.args.fontlogos:
            if self.args.material:
            if self.args.octicons: 
            if self.args.pomicons:
            if self.args.powersymbols:
            if self.args.weather:
```

Ah, the Powerline and PowerlineExtension have no suffix.

I do not really care, the resulting name will be too long anyhow and most circumstances. I consider the not `--complete` fonts to be special purpose and 'unsupported' ;-)

So whatever you like, we can have the suffix.

> Merge `--braille-style` into `--braille`

Here I have two comments.

First I would just get rid of `brailleStyle` and just put the style falvor into the `braille` argument, which is also un-surprising as the option will be stored there, and only by your code move to another variable later. I added a commit with a proposition of mine, please have a look.

Two, you use the `nargs='?'` to allow a stand alone `--braille` for the default and `--braille circle` for specifics.
I dread that ambiguity a bit, the parser is also not very smart. Take for example

```console
$ fontforge font-patcher --braille Arimo-Regular.ttf
font-patcher: error: argument --braille: invalid choice: 'Arimo-Regular.ttf' (choose from rectangle, circle, gapless)

works:
$ fontforge font-patcher Arimo-Regular.ttf --braille
$ fontforge font-patcher --braille --weather Arimo-Regular.ttf
```

Hmm, not very intuitive. In fact usually you NEED to add a specific option anyhow (or reorder to have the font name in the beginning, but I guess a lot of people go the ordinary way: first options, than filename).

Also congruent with how the usage is reported... first options, last is fontfile:

<details>

```
Nerd Fonts Patcher v3.4.0-86 (4.22.0) (ff 20230101)
usage: font-patcher [--careful] [--debug [{0,1,2,3}]] [--extension EXTENSION] [--help] [--makegroups [{-1,0,1,2,3,4,5,6}]] [--mono] [--outputdir OUTPUTDIR]
                    [--quiet] [--single-width-glyphs] [--variable-width-glyphs] [--version] [--braille [{rectangle,circle,gapless}]] [--complete]
                    [--codicons] [--fontawesome] [--fontawesomeext] [--fontlogos] [--material] [--octicons] [--pomicons] [--powerline] [--powerlineextra]
                    [--powersymbols] [--weather] [--adjust-line-height] [--boxdrawing] [--cell CELLOPT] [--configfile CONFIGFILE] [--custom CUSTOM] [--dry]
                    [--glyphdir GLYPHDIR] [--has-no-italic] [--metrics {HHEA,TYPO,WIN}] [--name FORCE_NAME] [--postprocess POSTPROCESS] [--removeligs]
                    [--xavgcharwidth [XAVGWIDTH]] [--progressbars | --no-progressbars]
                    font
```

</details>


I know, it is already there for `--debug` and `--makegroups` (and `--xavg_whatever`), but that means the position of the option is important now, which for me is unexpected, and I would rather remove the default values. I believe I first introduced it for `--makegroups` to keep compatible with previous versions (that had no number argument), and then modeled `--debug` and `--xavg` in the same way, unaware of the possible problems. But then we had issues of people ordering the options in arbitrary ways, and ... I believe it is a bad idea. It just does not work well together with a positional argument (i.e. the font name). Or maybe there is a way to make the parser more smart; I didn't find any solution.

(With `--xavg` there is no way around this `nargs='?'`, at least not trivially; but that option is a very special expert option anyhow)

So I would probably drop `--braille` (with default). Hmmm.

_Edit: I mean "drop `--braille`" but "keep `--braille circle`, `--braille rectangle`", making patching without `-c` but with Braille more complicated :thinking:_

What do you think :-D

_Edit: Well, add some details and examples_

GoldPigg:
> First I would just get rid of `brailleStyle` and just put the style falvor into the `braille` argument

Ok, I think this is no problem.

> So I would probably drop `--braille` (with default).

I think this is a right choice. 
After all, it doesn't seem like many people need to use `--braille` only, so making it a bit more complex (without a default) is acceptable.

Finii:
Hmm, we have
* `--braille`
* `--xavgcharwidth`
* `--debug`
* `--makegroups`

which have on optional argument. Maybe just keep `--braille` with it, as you programmed it :thinking: 

I guess this can be merged now from my side. Anything you want to add/change before?

GoldPigg:
> which have on optional argument. Maybe just keep `--braille` with it, as you programmed it 🤔

Hmm, perhaps when faced with complexity and potential mistakes, I'd choose complexity. So I think things are just fine as they are now.

> I guess this can be merged now from my side. Anything you want to add/change before?

I don't think anything needs to be added or changed now. This can be merged now from my side, too.

panda667:
This looks awesome. I've read both chains and everything looks great!

Finii:
Thanks a lot @GoldPigg for the work, the discussions, and the patience.

Very nice feature, indeed!

_Edit: Typo_

Finii:
@allcontributors please add @GoldPigg for code

Finii:
Added 3 commits with the forgotten things, as always ;-)

d7c5c11e9 `doc: Update font-patcher params`
1fe4fc2a3 `font-patcher: Do not automatically add Braille in some cases`
a66c673fb `font-patcher: Always force all Braille glyphs in`

:woman_shrugging:

Finii:
Hmm, I see the `gapless` ist not really gapless, there is a bit of empty space in the top?

image

Also visible in the screenshots above...

image

Finii:
Ah srry that was wrong, the FULL BLOCK shown above is the 'original' one, not the one we create with `--box` to fit the cell size. Everything ok, except maybe overlap?

image

Finii:
```
SYM_ATTR_BRAILLE = {
    # Don't scale because we generate font in the suitable size directly
    'default': {'align': '', 'valign': '', 'stretch': '', 'params': {'careful': False}}
}
SYM_ATTR_BOX = {
    'default': {'align': 'c', 'valign': 'c', 'stretch': '^xy', 'params': {'overlap': 0.02, 'dont_copy': box_keep}},
```

Ah yes. And probably no overlap is 100% ok.

clort81:
Reasons for gapless:
dot-braille_vs_block_braille1
dot-braille_vs_block_braille2
dot-braille_vs_block_braille3

Finii:
> Reasons for gapless

Well, your image rendering software should instead use the Block-Octant glyphs at codepoint `1CD00` to `1CDE5`, which is exactly what you prefer.

https://www.unicode.org/charts/PDF/U1CC00.pdf

Screenshot 2026-08-10 at 09 27 04

PR #1926, opened 2025-09-07
Scale non-mono icons to be at least as large as mono

danielwe:
#### Description

For icons with the `pa` stretch rule, `font_patcher` currently scales icons up to fit the cell in monospace patching, but only scales down for non-monospace. This makes some icons significantly _smaller_ with non-monospaced patching than with monospaced, which is rather unexpected. Case in point: the `nd-md-alpha` alphabetic icons are tiny in non-monospaced fonts, but substantial in monospaced.

With this PR, non-monospaced patching never makes smaller icons than monospaced. For `pa`, if monospaced patching would scale up, non-monospaced scales up by the same amount, but no more.

#### Requirements / Checklist

- [x] Read the [Contributing Guidelines](https://github.com/ryanoasis/nerd-fonts/blob/-/contributing.md)
- [ ] I've discussed this with core contributors already. If not checked, I'm ready to accept this work might be rejected in favor of a different grand plan.
      Issue number where discussion took place: #xxx
- [ ] If this contains a font/glyph add its origin as background info below (e.g. URL)
- [ ] Verified the license of any newly added font, glyph, or glyph set. License is: xxx

#### What does this Pull Request (PR) do?

* Factors the logic for determining `iconheight` out into a separate function that takes an optional `single_width` parameter
* Uses this function to correctly calculate the scale factors that would be obtained in `single_width` when needed
* For `pa` icons that don't need downscaling, sets the scale factor to the maximum of 1 and the single width scale factor.

#### How should this be manually tested?

* Patch a font in both mono and non-mono
* Look at `󰬇 (nd-md-alpha_z)` in both variants at the same font size
* With the previous patcher, you'll see something like this, where the larger icon is with the Mono variant
  Screenshot 2025-09-07 at 01 46 35Screenshot 2025-09-07 at 01 46 54
* With this PR, the icon will have the larger size in both variants

#### Any background context you can provide?

See explanation.

#### What are the relevant tickets (if any)?

NA.

#### Screenshots (if appropriate or helpful)

See above.

PR #2023, opened 2026-07-08
Fix Material Design icon names

Finii:
[why]
They have the same glyphs on two different codepoints with different meanings, but of course only one name. This will break the cheat sheet.

[how]
Add some hand-picked names for the problematic glyphs/codepoints.

Rerun to update the library file:

        ./generate-glyph-info-from-set.py \
          --font ../../src/glyphs/materialdesign/MaterialDesignIconsDesktop.ttf \
          --start f0001 --end f1af0 --offset 0 --prefix md > lib/i_md.sh

On release that will update the glyphnames and cheat sheet databases.

[note]
Noticed when working on PR #1926.

#### Requirements / Checklist

### DIFF
--- a/bin/scripts/generate-glyph-info-from-set.py
+++ b/bin/scripts/generate-glyph-info-from-set.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 # coding=utf8
 # Nerd Fonts Version: 3.4.0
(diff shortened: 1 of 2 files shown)


Finii:
Example ...

### before

image

### after

image

PR #2048, opened 2026-07-28
Several font updates for NF 3.5

Finii:
#### Description

Update fonts from upstreams

#### Requirements / Checklist


Finii:
Not updating Inconsolata
* Original repo is sill at 3.000
* Google Fonts has a variable and static 3.100
* But no repo given, how I do hate google fonts
* Very cheapo static conversion with overlapping paths, that are not really 'allowed' in statics:
image

_Left: 3.000, right; 3.100; both static_

PR #2077, opened 2026-09-04
Update IBM Plex Mono to 2.005

vinitsangwan:
I have updated this


Finii:
How does this differ from
* #2072

Even the titel is identical 🤔

Finii:
No reply

